PR #1, opened 2021-07-29
Scaffold Demo Project and Initial Modules

sshropshire:
### Summary of changes

 - Pilot project structure
 - Create Demo Application
 - Create Core Module with a `PaymentsClient`
 - Create Card Module with `PaymentsClient` extension to add card functionality

 ### Checklist

### Notes

This PR is an attempt to develop an initial project structure. It should facilitate discussion about best practices for our upcoming SDK work.

### Authors

- @kate-paypal
- @sshropshire


.gitignore
```diff
@@ -83,3 +83,7 @@ lint/generated/
 lint/outputs/
 lint/tmp/
 # lint/reports/
+
+# macOS
+.DS_Store
+
```
sarahkoop, line 89: Looking at `braintree_android` gitignore we should probably include 
```
.gradle
/local.properties
/.idea
```
in the gitignore and remove the .idea files already included. There are a few other things in that .gitignore that I'm not sure if we need here or not yet.

sshropshire: Right yeah this was pretty much *autogen*. I didn't modify much since I figure leaning in to Android studio might be for the best lol.

Demo/src/main/java/com/paypal/android/DemoActivity.kt
```diff
@@ -0,0 +1,18 @@
+package com.paypal.android
+
+import androidx.appcompat.app.AppCompatActivity
+import android.os.Bundle
+import com.paypal.android.card.payWithCard
+import com.paypal.android.core.PaymentsClient
+
+class DemoActivity : AppCompatActivity() {
```
tdchow, line 8: What are everyone's thoughts on creating a kotlin specific demo and a java specific demo? If we have the bandwidth, it'd be nice to show a java integration along with the ability to test java syntax for new sdk surface area.

Card/build.gradle
```diff
@@ -0,0 +1,45 @@
...
+    }
+}
+
+dependencies {
+    api project(':Core')
+
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    implementation 'androidx.core:core-ktx:1.6.0'
+    implementation 'androidx.appcompat:appcompat:1.3.1'
+    implementation 'com.google.android.material:material:1.4.0'
```
tdchow, line 41: I know this is all generated by Android Studio, but it'd be cool if we could consolidate library versions and even dependency imports in the root build.gradle file that all modules can reference. No need to make that change in this PR, but I'll add a story to the backlog to update this!

Card/build.gradle
```diff
@@ -0,0 +1,45 @@
...
+    id 'com.android.library'
+    id 'kotlin-android'
+}
+
+android {
+    compileSdkVersion 30
+    buildToolsVersion "30.0.2"
+
+    defaultConfig {
+        minSdkVersion 21
```
tdchow, line 11: We had an issue with the PayPal Auth SDK with minSDKversions of less than 23. We can always update this before our 1.0 release, but probably worth chatting as a team about it.

Demo/src/main/java/com/paypal/android/DemoActivity.kt
sshropshire: I think that's a good idea. It'll be a good playground to write sample code as well.

Card/build.gradle
sshropshire: Yeah for sure. We've kept them as strings in the past to get Android studio notices about upgrades. I know there's another `buildSrc` pattern that may allow us to cut down on some repetition.

Card/build.gradle
sshropshire: Ok sounds good. We currently support API 21 in the Braintree SDK and I think in previous versions we went all the way back to API 16. I did also hear of some merchants with heavy traffic on some of the older versions of both iOS and Android. We should definitely support what we can, but I agree this will be a trade-off to allow us to ship the best SDK possible.

gradle.properties
```diff
@@ -0,0 +1,21 @@
...
+# When configured, Gradle will run in incubating parallel mode.
+# This option should only be used with decoupled projects. More details, visit
+# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
+# org.gradle.parallel=true
+# AndroidX package structure to make it clearer which packages are bundled with the
+# Android operating system, and which are packaged with your app"s APK
+# https://developer.android.com/topic/libraries/support-library/androidx-rn
+android.useAndroidX=true
+# Automatically convert third-party libraries to use AndroidX
+android.enableJetifier=true
```
sarahkoop: Maybe we could remove this for now to make sure we are aware if we introduce any dependencies that aren't on AndroidX

sshropshire: Yeah that's a good call. There are supposed to be some build performance benefits from this as well.

Card/build.gradle
tdchow: Sounds good. We can include a few solutions that we can weigh out once we have the story created.

Card/build.gradle
tdchow: Actually, now that I'm thinking about it, we'll probably want to keep support for API level 21. The functions that invoke the PayPal paysheet require API 23 and up, so If we end up pulling in the PayPal Checkout SDK for the PayPal module, the merchant would have to have some fallback for API 21 and 22 if their app supported it.

sarahkoop (approved):
I think we've got some good discussions here to have in the future, but approving this as a starting point 🌟

tdchow (approved):
I agree with @sarahkoop, approving as well!

PR #6, opened 2021-08-10
[DTPPSDK-227] - Consolidate Gradle imports and versions

tdchow:

### Summary of changes

 - Consolidate Gradle imports and versions in the root directory's build.gradle file
 - Update all module build.gradle files to pull from the root directory's build.gradle file
 - Add detekt version to gradle.properties file
 - Remove jcenter

 ### Checklist

- @tdchow

### DIFF
--- a/Card/build.gradle
+++ b/Card/build.gradle
@@ -44,7 +44,6 @@ dependencies {
     implementation deps.kotlinStdLib
     implementation deps.androidxCoreKtx
     implementation deps.androidxAppcompat
-    implementation deps.material
 
     testImplementation deps.junit
     androidTestImplementation deps.androidxJUnit
--- a/Card/build.gradle
+++ b/Card/build.gradle
@@ -41,12 +41,12 @@ android {
 dependencies {
     api project(':Core')
 
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-    implementation 'androidx.core:core-ktx:1.6.0'
-    implementation 'androidx.appcompat:appcompat:1.3.1'
-    implementation 'com.google.android.material:material:1.4.0'
-    testImplementation 'junit:junit:4.13.2'
-    androidTestImplementation 'androidx.test.ext:junit:1.1.3'
-    androidTestImplementation 'androidx.test.espresso:espresso-core:3.4.0'
+    implementation deps.kotlinStdLib
+    implementation deps.androidxCoreKtx
+    implementation deps.androidxAppcompat
+    implementation deps.material
 
+    testImplementation deps.junit
+    androidTestImplementation deps.androidxJUnit
+    androidTestImplementation deps.androidxEspressoCore
 }
--- a/build.gradle
+++ b/build.gradle
@@ -1,30 +1,30 @@
 buildscript {
     ext.versions = [
-            'kotlin': '1.5.21'
+            "kotlin": "1.5.21"
     ]
     ext.deps = [
             // Kotlin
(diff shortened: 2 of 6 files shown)


gradle.properties
```diff
@@ -17,4 +17,6 @@ org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
 android.useAndroidX=true
 
 # Kotlin code style for this project: "official" or "obsolete":
-kotlin.code.style=official
\ No newline at end of file
+kotlin.code.style=official
+
+detektVersion=1.18.0-RC2
```
tdchow, line 22: The detekt version needed to be added in gradle.properties due to this gradle plugin limitation: https://github.com/gradle/gradle/issues/1697#issuecomment-506910915

Card/build.gradle
```diff
@@ -41,12 +41,12 @@ android {
...
-    implementation 'androidx.core:core-ktx:1.6.0'
-    implementation 'androidx.appcompat:appcompat:1.3.1'
-    implementation 'com.google.android.material:material:1.4.0'
-    testImplementation 'junit:junit:4.13.2'
-    androidTestImplementation 'androidx.test.ext:junit:1.1.3'
-    androidTestImplementation 'androidx.test.espresso:espresso-core:3.4.0'
+    implementation deps.kotlinStdLib
+    implementation deps.androidxCoreKtx
+    implementation deps.androidxAppcompat
+    implementation deps.material
```
sarahkoop: Do we need material outside of the demo app?

build.gradle
```diff
@@ -1,29 +1,41 @@
-// Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
-    ext.kotlin_version = "1.4.10"
+    ext.versions = [
+            'kotlin': '1.5.21'
+    ]
+    ext.deps = [
+            // Kotlin
+            'kotlinStdLib'        : "org.jetbrains.kotlin:kotlin-stdlib:${versions.kotlin}",
```
sarahkoop: Nitpick but do we want to standardize using double or single quotes?

build.gradle
```diff
@@ -1,29 +1,41 @@
...
+
+            // Android
+            'androidxCoreKtx'     : 'androidx.core:core-ktx:1.6.0',
+            'androidxAppcompat'   : 'androidx.appcompat:appcompat:1.3.1',
```
sarahkoop:
```suggestion
            'androidxAppCompat'   : 'androidx.appcompat:appcompat:1.3.1',
```
Not sure if this should be camel cased like that but it looks right to me

Card/build.gradle
tdchow: Good catch. I don't think we do at the moment. I'll remove material from the card and core modules.

build.gradle
tdchow: Unfortunately groovy requires the use of double quotes when using string interpolation 😞  I prefer using single quotes, but cool with updating all to double quotes to keep things uniform.

build.gradle
tdchow: I thought about this one a little. `Appcompat` looked a little strange to me as well, but the library's artifact ID is `appcompat` which is why I ended up going with that.

build.gradle
sarahkoop: I don't feel that strongly - we can make the convention double for string interpolation and single for everything else or go all double.

tdchow: I'm cool going all double-quotes. If we need to use string interpolation later for certain libraries, we won't need to update the type of quotes used.

sarahkoop (approved):
This is so much cleaner! LGTM 🌟

sshropshire (approved):
🌊  Clean!

PR #7, opened 2021-08-11
[DTPPSDK-235] Add Diffuse AAR Analysis to GitHub Actions

sarahkoop:
### Summary of changes

 - Set up [Diffuse](https://github.com/JakeWharton/diffuse) aar size analysis in GitHub actions
 - Running locally requires Diffuse to be installed via homebrew - this doesn't seem necessary (CI should be fine for now), so did not include steps to run locally in the README. 

 ### Checklist

- @sarahkoop

### DIFF
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -27,7 +27,7 @@ jobs:
           distribution: 'zulu'
       - name: Lint
         run: ./gradlew lint
-  build_main:
+  build_main_and_current:
     name: Build Main Branch
     runs-on: ubuntu-latest
     steps:
@@ -45,7 +45,24 @@ jobs:
       - name: Upload Artifacts
         uses: actions/upload-artifact@v2
         with:
-          name: Card AAR
+          name: Main Card AAR
+          path: Card/build/outputs/aar/Card-release.aar
+    name: Build Current Branch
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Current Branch
+        uses: actions/checkout@v2
+      - name: Set up Java 8
+        uses: actions/setup-java@v2
+        with:
+          java-version: '8'
+          distribution: 'zulu'
+      - name: Assemble Release AAR
+        run: ./gradlew clean assembleRelease
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: Current Card AAR
           path: Card/build/outputs/aar/Card-release.aar
   diffuse:
     name: Diffuse AAR Analysis
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -76,6 +76,6 @@ jobs:
           distribution: 'zulu'
           java-version: '11'
       - name: Run Diffuse Analysis - Card
-        run: diffuse diff --aar main-card-aar/Card-release.aar current-card-aar/Card-release.aar --stdout --text output.txt
+        run: diffuse diff --aar main-card-aar/Card-release.aar current-card-aar/Card-release.aar --text output.txt
       - name: Run Diffuse Analysis - Core
-        run: diffuse diff --aar main-core-aar/Core-release.aar current-core-aar/Core-release.aar --stdout --html output.html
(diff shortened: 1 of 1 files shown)


.github/workflows/static_analysis.yml
```diff
@@ -27,3 +28,59 @@ jobs:
           distribution: 'zulu'
       - name: Lint
         run: ./gradlew lint
+
+  diffuse:
+    name: Diffuse AAR Analysis
+    runs-on: macOS-11
```
sarahkoop, line 34: Running this on macOS-11 because it uses homebrew to install diffuse

.github/workflows/static_analysis.yml
```diff
@@ -27,3 +28,59 @@ jobs:
...
+          path: |
+            Core/build/outputs/aar/Core-release.aar
+            Card/build/outputs/aar/Card-release.aar
+
+      # Set up Diffuse
+      - name: Install Diffuse
+        run: brew install JakeWharton/repo/diffuse
+      - name: Download Workflow Artifacts 
+        uses: actions/download-artifact@v2
+      - name: Set up Java 11
```
sarahkoop, line 76: This is a diffuse bug - currently aar comparisons can only be done using Java 11

.github/workflows/static_analysis.yml
```diff
@@ -27,3 +28,59 @@ jobs:
...
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'zulu'
+          java-version: '11'
+
+      # Run Diffuse analysis
```
sarahkoop, line 82: I did some trial and error with the output of this. HTML reports aren't currently supported, but text reports are. I think looking at the GitHub actions standard output is easier than downloading a zip with text file outputs - but open to other ideas on the best way to view the comparison

.github/workflows/static_analysis.yml
```diff
@@ -27,3 +28,59 @@ jobs:
...
+      - name: Run Diffuse Analysis - Card
+        run: diffuse diff --aar main-aar/Card/build/outputs/aar/Card-release.aar current-aar/Card/build/outputs/aar/Card-release.aar
+      - name: Run Diffuse Analysis - Core
```
sarahkoop, line 85: The run steps and the upload / download artifact steps would have to be modified for each new module added. With this approach they each get their own std output report printed in the action. Do we think this is an OK approach? 

Alternatively - we could compare demo app APK size  (if the demo app includes all modules any changes in module size would be reflected in the demo app APK size change right?). This would allow us to only run diffuse once, but would provide less granular insight into actual SDK size.

sarahkoop:
Providing an example screenshot of how to view the diffuse output with the current setup. In the GH Actions checks on the PR select `details` on the `Static Analysis / Diffuse AAR Analysis (pull_request)` and expand the `Run Diffuse Analysis - Module` steps to see the output
Screen Shot 2021-08-11 at 2 03 57 PM

.github/workflows/static_analysis.yml
sshropshire: Yeah granular is probably best honestly, although if we use the Demo project as an example, we know the upper limit of our SDK's size which is probably a good enough metric for anyone who asks.

sshropshire (approved):
🌊  The diff report is awesome!

.github/workflows/static_analysis.yml
scannillo: I think the GH output is fine!

scannillo (approved):
This looks awesome!

For future - I wonder if it's valuable to have it run on every PR, or would be more something we would rather run as one-off via the workflow_dispatch.

.github/workflows/static_analysis.yml
tdchow: The console output sounds good to me!

.github/workflows/static_analysis.yml
tdchow: I agree with @sshropshire. Having granular reports will help pinpoint which module needs to be investigated if we see a large increase in SDK size.

tdchow (approved):
The reports look great! 🎉

PR #11, opened 2021-08-17
Publish to Maven

jcnoriega:
Added job to publish to Maven. 
Also added complete release workflow.

Screen Shot 2021-08-20 at 13 06 58


Card/build.gradle
```diff
@@ -12,14 +12,15 @@ android {
         minSdkVersion 21
         targetSdkVersion 30
         versionCode 1
-        versionName "1.0"
+        versionName "0.0.1-testCard"
```
sarahkoop: Let's pull all these into the top level gradle (I think `versionName` can be the same for all modules)

build.gradle
```diff
@@ -22,7 +22,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:4.2.2"
+        classpath 'com.android.tools.build:gradle:4.1.3'
```
sarahkoop: Did we need to downgrade the build tools version for publishing?

gradle.properties
```diff
@@ -19,4 +19,23 @@ android.useAndroidX=true
...
+
+
+POM_NAME=PayPal Payments - Android
+POM_PACKAGING=aar
+POM_DESCRIPTION=Paypal payment methods for Android
+POM_URL=not_defined
+
+POM_SCM_URL=https://github.com/paypal/Android-SDK
+POM_SCM_CONNECTION=scm:git:git://paypal/Android-SDK.git
+POM_SCM_DEV_CONNECTION=scm:git:ssh://paypal/Android-SDK.git
```
sarahkoop: Should these be using the `git@github` urls instead?

gradle.properties
```diff
@@ -19,4 +19,23 @@ android.useAndroidX=true
...
 kotlin.code.style=official
 
-detektVersion=1.18.0-RC2
\ No newline at end of file
+detektVersion=1.18.0-RC2
+
+
+POM_NAME=PayPal Payments - Android
+POM_PACKAGING=aar
+POM_DESCRIPTION=Paypal payment methods for Android
```
sarahkoop: I think these need to be specific to each module not project level

gradle/gradle-publish.gradle
```diff
@@ -0,0 +1,81 @@
...
+                    license {
+                        name = POM_LICENCE_NAME
+                        url = POM_LICENCE_URL
+                    }
+                }
+                developers {
+                    developer {
+                        id = POM_DEVELOPER_ID
+                        name = POM_DEVELOPER_NAME
+                        email = POM_DEVELOPER_EMAIL
```
sarahkoop, line 36: Is email required? If we aren't defining it and it's not required lets remove

gradle/gradle-publish.gradle
```diff
@@ -0,0 +1,81 @@
...
+ext["SONATYPE_NEXUS_USERNAME"] = System.getenv('SONATYPE_NEXUS_USERNAME') ?: ''
+ext["SONATYPE_NEXUS_PASSWORD"] = System.getenv('SONATYPE_NEXUS_PASSWORD') ?: ''
+
+
+afterEvaluate {
+    android.libraryVariants.each { variant ->
+        def variantName = variant.name
+        def resolvedVersion = project.version
+        publishing.publications.create variantName, MavenPublication, {
+            println "$variantName resolvedVersion: ${resolvedVersion}  ${group} ${project.name}"
```
sarahkoop: Do we want this `println`? Is it printing to the GH actions console?

gradle/gradle-publish.gradle
```diff
@@ -0,0 +1,81 @@
...
+ext["signing.password"] = System.getenv('SIGNING_KEY_PASSWORD') ?: ''
+ext["signing.secretKeyRingFile"] = System.getenv('SIGNING_KEY_FILE') ?: ''
+ext["SONATYPE_NEXUS_USERNAME"] = System.getenv('SONATYPE_NEXUS_USERNAME') ?: ''
+ext["SONATYPE_NEXUS_PASSWORD"] = System.getenv('SONATYPE_NEXUS_PASSWORD') ?: ''
+
+
+afterEvaluate {
+    android.libraryVariants.each { variant ->
+        def variantName = variant.name
+        def resolvedVersion = project.version
```
sarahkoop: Do these variables need to be defined separately or can we just access `variant.name` and `project.version` directly?

gradle/gradle-publish.gradle
```diff
@@ -0,0 +1,81 @@
...
+        publishing.publications.create variantName, MavenPublication, {
+            println "$variantName resolvedVersion: ${resolvedVersion}  ${group} ${project.name}"
+            groupId group
+            version resolvedVersion
+            artifactId project.name
```
sarahkoop: Is `project.name` the module name here? So card and core can have their own artifact IDs?

gradle/gradle-publish.gradle
jcnoriega: I added this to debug, i'll remove it

Card/build.gradle
jcnoriega: @sarahkoop do you mean that every module is going to have the same `versionName`?

gradle/gradle-publish.gradle
jcnoriega: yes, `project.name` is the module name. So we upload  `com.paypal.android.core` and `com.paypal.android.card` for example

Card/build.gradle
sarahkoop: Yes in `braintree_android` this field is set at [top level](https://github.com/braintree/braintree_android/blob/master/build.gradle#L35) and corresponds to the version to be released like `1.0.0-SNAPSHOT` or `1.1.0` or 1.0.0-beta`. All modules should be on the same version I believe.

sshropshire: @jcnoriega Yeah historically it's been a bit of a compromise. Separate version numbers complicate git tagging and releases in general. We went with a monolith versioning scheme. We mainly use semantic versioning to communicate new features / breaking changes to merchants.

.github/workflows/build.yml
```diff
@@ -12,5 +14,14 @@ jobs:
...
           distribution: 'zulu'
+            #After decoding the secret key, place the file in ~ /. Gradle/ secring.gpg
+      - name: Decode secret key
+        run: |
+            echo "${{secrets.SIGNING_KEY_FILE}}" > ~/secretKey.gpg.b64
+            base64 -d ~/secretKey.gpg.b64 > ${{ env.SIGNING_KEY_FILE_PATH }}
       - name: Assemble
         run: ./gradlew --stacktrace assemble
+        env:
+          SIGNING_KEY_ID: ${{ secrets.SIGNING_KEY_ID }}
```
sshropshire, line 26: random q: where do we go to manage secrets in GH actions?

jcnoriega: you need to have admin access to the repo, then go to Settings > Secrets.

gradle.properties
jcnoriega: i copied this from our checkout sdk, but then after reading a bit, it turns out that that is the way maven requires it: https://maven.apache.org/scm/git.html

.github/workflows/build.yml
sshropshire: 👍  TIL

Card/build.gradle
```diff
@@ -12,14 +12,14 @@ android {
         minSdkVersion 21
         targetSdkVersion 30
         versionCode 1
```
sarahkoop: Can we extract these numbers in Card and Core into the shared gradle?

build.gradle
```diff
@@ -22,7 +27,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:4.2.2"
+        classpath 'com.android.tools.build:gradle:4.2.2'
```
sarahkoop: Nitpick but I think we are going with all double quotes for standardization

gradle.properties
sarahkoop: Does Sonatype require this too? In BT Android we use git@github. Not sure which is better here

gradle.properties
```diff
@@ -19,4 +19,23 @@ android.useAndroidX=true
...
+POM_URL=not_defined
+
+POM_SCM_URL=https://github.com/paypal/Android-SDK
+POM_SCM_CONNECTION=scm:git:git://paypal/Android-SDK.git
+POM_SCM_DEV_CONNECTION=scm:git:ssh://paypal/Android-SDK.git
+
+POM_LICENCE_NAME=not_defined
+POM_LICENCE_URL=not_defined
```
sarahkoop, line 35: Are these going to be addressed in a separate PR? Let's define them now if we can

gradle/gradle-publish.gradle
```diff
@@ -0,0 +1,77 @@
...
+ext["SONATYPE_NEXUS_USERNAME"] = System.getenv('SONATYPE_NEXUS_USERNAME') ?: ''
+ext["SONATYPE_NEXUS_PASSWORD"] = System.getenv('SONATYPE_NEXUS_PASSWORD') ?: ''
+
+
+afterEvaluate {
+    android.libraryVariants.each { variant ->
+        publishing.publications.create variant.name, MavenPublication, {
+            groupId group
+            version project.version
+            artifactId project.name.toLowerCase()
```
sarahkoop: We might want some additional logic here for the artifact naming. Do we want dashes rather than all lower case for modules with two words (i.e. `core-payments` rather than `corepayments`)?

gradle/gradle-publish.gradle
```diff
@@ -0,0 +1,77 @@
...
+            from components."${variant.name}"
+            artifact androidSourcesJar
```
sarahkoop: Not sure exactly the kotlin equivalent of javadocs but we are missing publishing those here - Should we add a TODO somewhere so we don't forget that before release?

.github/workflows/build.yml
```diff
@@ -4,6 +4,8 @@ jobs:
   build_aar:
     name: Build
     runs-on: ubuntu-latest
+    env:
```
tdchow, line 7: Do we need to have these signing steps in our build workflow?

.github/workflows/publish_to_maven.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Publish Card Module
+        run: ./gradlew --stacktrace clean :Card:publishReleasePublicationToMavenRepository
+        env:
+          SONATYPE_NEXUS_USERNAME: ${{ secrets.SONATYPE_NEXUS_USERNAME }}
+          SONATYPE_NEXUS_PASSWORD: ${{ secrets.SONATYPE_NEXUS_PASSWORD }}
+          SIGNING_KEY_ID: ${{ secrets.SIGNING_KEY_ID }}
+          SIGNING_KEY_PASSWORD: ${{ secrets.SIGNING_KEY_PASSWORD }}
+          SIGNING_KEY_FILE: ${{ env.SIGNING_KEY_FILE_PATH }}
+
+      - name: Publish Core Module
```
tdchow: Currently, these publish steps will run in sequence. Did we want to create separate jobs for each that will run in parallel?

Core/build.gradle
```diff
@@ -11,15 +11,16 @@ android {
     defaultConfig {
         minSdkVersion 21
         targetSdkVersion 30
-        versionCode 1
-        versionName "1.0"
-
+        versionCode 2
```
tdchow: Is there a reason why we bumped versionCode to 2?

gradle/gradle-publish.gradle
tdchow: I vote for using dashes in the artifactIDs.

Core/build.gradle
jcnoriega: Oops, forgot about this, was doing some testing

.github/workflows/build.yml
jcnoriega: yes, because the assemble task tries to sign the artifacts, so i need to include the signing keys in the build

gradle/gradle-publish.gradle
jcnoriega: I added `toLowerCase` to transform `Card` and `Code` into `card`, `code`. If a module has two words, i would add the dash on the module name.

gradle/gradle-publish.gradle
jcnoriega: im good with adding a TODO, im not really sure how to generate those. I dont want to just paste something from `stackoverflow`. Maybe we can also add a ticket and sync later on it. Thoughts?

gradle.properties
jcnoriega: This is the one we have in our checkout repo: https://developer.paypal.com/docs/limited-release/native-checkout/license/
Im not sure if it applies though. Perhaps we can add TODO and create a ticket for it until we are sure?

gradle/gradle-publish.gradle
sarahkoop: I'd opt to pull the artifact names into their own gradle files along with the POM name and descriptions (since those should be module specific). Dashes in module names doesn't feel like Kotlin/Java convention to me. Thoughts?

gradle/gradle-publish.gradle
sarahkoop: Yeah let's just create a JIRA card for that

gradle/gradle-publish.gradle
jcnoriega: @sarahkoop agreed

gradle.properties
jcnoriega: Me neither, i dont mind putting git@github

.github/workflows/publish_to_maven.yml
```diff
@@ -0,0 +1,55 @@
+name: Publish to Maven
+on: [pull_request, workflow_dispatch]
```
sarahkoop:
```suggestion
on: [workflow_dispatch]
```
Don't want to release on every pull request

.github/workflows/release.yml
```diff
@@ -0,0 +1,125 @@
+name: Publish Release
+on: [pull_request, workflow_dispatch]
```
sarahkoop:
```suggestion
on: [workflow_dispatch]
```

.github/workflows/publish_to_maven.yml
jcnoriega: @sarahkoop I had to add `pull_request` because of a bug in github actions. If i only add `workflow_dispatch` it wont appear, unless that file is already on `main` branch. I was going to delete them once this PR is approved

sarahkoop: Maybe I'm confused - it looks like the release github workflow is executing successfully, what is preventing these artifacts from being published automatically currently?

jcnoriega: its like this: github actions has a bug. If your `.yml` is not on your default branch (in our case `main` branch) and your workflow ONLY has `workflow_dispatch`, you wont be able to see it in the github UI. So, in order for me to test it, i had to add `pull_request`, and only then the workflow appeared in the github UI. Before merging it to main, i'm going to delete `pull_request` tag, but until i get all the approvals, i need to leave it there so that i can test the workflow. Does that make sense?

sshropshire (approved):

sarahkoop (approved):
Looks good to me - pending removing `pull_request` as the GH workflow trigger, and adding a step to close and release the repository in the future

PR #12, opened 2021-08-26
Add Networking Layer

sshropshire:
### Summary of changes

This PR scaffolds our `Http` class, which can be used to make network requests

 - Add `Http` class to make a simple `GET` request
 - We will use `HttpURLConnection` for networking to minimize dependency on 3rd-party libraries
 - The remainder of this class can be TDD'd, e.g. once we have a feature that requires an HTTP `POST`, we can add this feature to `Http` as needed 

 ### Checklist

- @sshropshire


Core/src/main/java/com/paypal/android/core/Http.kt
```diff
@@ -0,0 +1,24 @@
...
+
+class Http {
+
+    suspend fun send(request: HttpRequest) = send(request, Dispatchers.IO)
+
+    @VisibleForTesting
+    suspend fun send(request: HttpRequest, dispatcher: CoroutineDispatcher) =
+        withContext(dispatcher) {
+            val url = request.url
+            val connection = url.openConnection() as HttpURLConnection
```
jcnoriega, line 13: Just to clarify,  are we not we going to use `OkHttp` or `Retrofit`?

sshropshire: I think so yeah. One other thing I didn't think about was Retrofit requires us to know the URLs at [compile time](https://square.github.io/retrofit/). I think we may need that runtime ability to specify our HTTP request paths.

sshropshire: Also using the built-in `HttpURLConnection` will keep us from having dependency conflicts with merchants consuming the library.

Core/build.gradle
```diff
@@ -30,23 +30,33 @@ android {
...
+        exclude 'META-INF/AL2.0'
+        exclude 'META-INF/LGPL2.1'
+    }
 }
 
 dependencies {
     implementation deps.kotlinStdLib
     implementation deps.androidxCoreKtx
     implementation deps.androidxAppcompat
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.0"
```
tdchow: Can we define these imports in the root build.gradle's `deps`?

Core/src/androidTest/java/com/paypal/android/core/HttpIT.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.After
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class HttpIT {
+
+    @ExperimentalCoroutinesApi
```
tdchow: We should be able to annotate the entire class with `@ExperimentalCoroutinesApi` which should clean up the annotations on each function/field.

Core/src/main/java/com/paypal/android/core/Http.kt
```diff
@@ -0,0 +1,24 @@
+package com.paypal.android.core
+
+import androidx.annotation.VisibleForTesting
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.net.HttpURLConnection
+
+class Http {
```
tdchow: Injecting the `CoroutineContext` in the constructor would remove the need to overload the send function or having `Dispatchers.IO` as a default value on the function.

Core/src/test/java/com/paypal/android/core/HttpUnitTest.kt
```diff
@@ -0,0 +1,54 @@
...
+
+class HttpUnitTest {
+
+    @ExperimentalCoroutinesApi
+    val testCoroutineDispatcher = TestCoroutineDispatcher()
+
+    @ExperimentalCoroutinesApi
+    @Test
+    fun `send sets request method on url connection`() = runBlockingTest {
+        val httpRequest = spyk(HttpRequest("https://www.example.com"))
```
tdchow: It looks like the setup for these tests has a lot of similarities. Could we move these properties to be defined at the class level and utilize `@Before` on a function that could set up the expected behavior of these mocks?

```kotlin
    private val httpRequest = spyk(HttpRequest("https://www.example.com"))
    private val url = mockk<URL>()
    private val urlConnection = mockk<HttpURLConnection>(relaxed = true)
    private lateinit var sut: Http

    @Before
    fun setUp() {
        sut = Http()
        every { httpRequest.url } returns url
        every { url.openConnection() } returns urlConnection
    }

    @Test
    fun `send sets request method on url connection`() = runBlockingTest {
        sut.send(httpRequest, testCoroutineDispatcher)

        verify { urlConnection.requestMethod = "GET" }
    }

    @Test
    fun `send returns an http result`() = runBlockingTest {
        every { urlConnection.responseCode } returns 123

        val result = sut.send(httpRequest, testCoroutineDispatcher)

        assertEquals(123, result.responseCode)
    }
```

tdchow (commented):
The initial networking code is looking good! Another question: do we want to include the sample server in this repo?

sshropshire:
Seems like suspend functions don't get included in coverage reports: https://stackoverflow.com/questions/62402633/coroutines-body-does-not-cover-in-unit-test-code-coverage-android-studio.

Core/src/androidTest/java/com/paypal/android/core/HttpIT.kt
```diff
@@ -0,0 +1,37 @@
...
+import kotlinx.coroutines.test.runBlockingTest
+import org.junit.After
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@ExperimentalCoroutinesApi
+@RunWith(AndroidJUnit4::class)
+class HttpIT {
```
sarahkoop: Is `IT` the naming scheme we want to go with for integration test files? Seems like a good time to establish conventions on this. I'd vote for `HttpIntegrationTest`

Core/src/androidTest/java/com/paypal/android/core/HttpIT.kt
```diff
@@ -0,0 +1,37 @@
...
+        testDispatcher = TestCoroutineDispatcher()
+    }
+
+    @After
+    fun afterEach() {
+        testDispatcher.cleanupTestCoroutines()
+    }
+
+    @Test
+    fun itMakesAnHttpRequest() = runBlockingTest {
```
sarahkoop: Same comment here - do we want to establish a test naming pattern?

Demo/build.gradle
```diff
@@ -45,6 +46,7 @@ dependencies {
     implementation deps.androidxAppcompat
     implementation deps.material
     implementation deps.constraintLayout
+    implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.3.1'
```
sarahkoop: Can this also be added to deps?

Core/src/androidTest/java/com/paypal/android/core/HttpIT.kt
sshropshire: Yeah that makes sense. I think leading with the method name makes sense. Went with `send_makesAnHttpRequest` here.

Demo/build.gradle
sshropshire: Def yeah I missed this one 🙏 .

server/index.js
```diff
@@ -0,0 +1,37 @@
+
```
sarahkoop: Is this sample server still being used in this PR? It looks like we are just hitting example sites - do we still need this?

sshropshire: Oh right nah I'll delete.

sshropshire:
> The initial networking code is looking good! Another question: do we want to include the sample server in this repo?

I think we would need to update the sample server. At the moment I think it only vends BT client tokens.

sarahkoop (approved):

Core/src/main/java/com/paypal/android/core/Http.kt
```diff
@@ -0,0 +1,20 @@
+package com.paypal.android.core
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.net.HttpURLConnection
+
+class Http {
```
tdchow:
```suggestion
class Http(private val dispatcher: CoroutineDispatcher = Dispatchers.IO) {
```

Core/src/main/java/com/paypal/android/core/Http.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    suspend fun send(request: HttpRequest, dispatcher: CoroutineDispatcher = Dispatchers.IO) =
```
tdchow:
```suggestion
    suspend fun send(request: HttpRequest) =
```

tdchow (commented):
Had one more suggestion for moving the dispatcher to the constructor of the `Http` class.

scannillo:
> The initial networking code is looking good! Another question: do we want to include the sample server in this repo?

> I think we would need to update the sample server. At the moment I think it only vends BT client tokens.

We can use this sample merchant, which we had for PPCP (https://github.com/braintree/paypal-commerce-platform-sample-merchant). It vends order IDs, low scoped tokens, and full scoped tokens. Once we have account credentials for whatever prod/sandbox merchant we want to use, we can easily update that resources file in the repo. It also captures/authorizes as well.

tdchow (approved):
Looks great! Thanks for making the updates.

PR #13, opened 2021-08-26
Run test coverage reports on main branch

tdchow:

### Summary of changes

 - Moved the code coverage job to a separate workflow
 - Added a workflow trigger to run code coverage reports on the main branch.

We currently don't have a baseline report generated on the main branch. This change should fix future code coverage reports.

 ### Checklist

- @tdchow

### DIFF
--- a/.github/workflows/test_coverage.yml
+++ b/.github/workflows/test_coverage.yml
@@ -1,0 +1,31 @@
+name: Test Coverage
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+jobs:
+  test_coverage_job:
+    name: Test Code Coverage
+    runs-on: macos-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/test_coverage.yml
```diff
@@ -0,0 +1,31 @@
+name: Test Coverage
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
```
sarahkoop, line 5: Do we need to run it on push?

tdchow: I _think_ this is the correct trigger. The idea is that once a PR is merged into `main` the code coverage workflow will run. Is there a better workflow trigger to achieve this?

sarahkoop: Ah no you are right I forgot we don't push to main unless we are merging a PR (was thinking it would run on every branch push)

sarahkoop (approved):

sshropshire (approved):

PR #14, opened 2021-09-15
Create Core Config

sshropshire:
### Summary of changes

 - Create `PaymentsConfiguration` data class
 - Create `Environment` enum

 ### Checklist

### Authors

- @jcnoriega
- @sshropshire


Core/src/test/java/com/paypal/android/core/EnvironmentUnitTest.kt
```diff
@@ -0,0 +1,17 @@
+package com.paypal.android.core
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class EnvironmentUnitTest {
+
+    @Test
+    fun `it should return the correct url for the LIVE environment`() {
+        assertEquals("https://api.paypal.com", Environment.LIVE.url)
```
tdchow, line 10: Discussion topic for the team: Do we want to introduce an assertion library for our test suite? 

Here's a (somewhat outdated) [article](https://blog.frankel.ch/comparison-assertion-libraries/) comparing a few options. But it gives a good idea of some of the benefits of using an assertion library.

AssertJ - https://joel-costigliola.github.io/assertj/
Strikt - https://strikt.io/

tdchow (approved):
Changes are looking good! Only a question for the team on if we want to use an assertion library.

Core/src/test/java/com/paypal/android/core/EnvironmentUnitTest.kt
sarahkoop: In my opinion, I don't think the benefits of an assertion library outweigh the cost of additional outside dependencies.

Having just worked on Android 12 support for our other SDKs, I'm thinking of a future scenario where we want to add support for a new Android version and we could be blocked by some of these assertion libraries not offering support for it yet. 

I'm open to the potential of adding it in the future if we feel there is an assertion we want to make but aren't able to with junit, but I'd hold off on adding it until then.

sarahkoop (approved):

Core/src/test/java/com/paypal/android/core/EnvironmentUnitTest.kt
tdchow: I don't think we'll run into any compatibility issues here since we'd only pull in the dependency for unit tests and not in the SDKs source code.

But I'm cool holding off until we have a harder need for it.

PR #16, opened 2021-09-22
Demo App Home Screen

sarahkoop:
### Summary of changes

 - Add a home screen to the Demo App displaying `PaymentMethodsFragment` with buttons to launch `CardFragment` or `PayPalFragment`
 - Add `SettingsFragment` where environment (sandbox or live) can be configured. 
 - Use JetPack Compose for UI elements, JetPack NavigationComponent for navigation between fragments, and JetPack Preference for demo app settings

**Note**: If you are able to, please pull down this branch and test out the app - we updated the Java version for the project to 11 to use some JetPack libraries and we want to make sure it's configured / runs correctly for all developers.

 ### Checklist

- @andrespelaezp 
- @tdchow 
- @sarahkoop


Demo/build.gradle
```diff
@@ -62,6 +63,11 @@ dependencies {
     implementation deps.composeRuntimeLivedata
     implementation deps.composeFoundation
     implementation deps.composeMaterial
+    implementation 'androidx.legacy:legacy-support-v4:1.0.0'
```
tdchow: Is the legacy-support library needed? Not sure if this was added for the Navigation Component.

Demo/build.gradle
```diff
@@ -62,6 +63,11 @@ dependencies {
...
+    implementation deps.navigationFragment
+    implementation deps.navigationUI
+    implementation deps.preference
+    implementation 'androidx.preference:preference:1.1.1'
```
tdchow: Can we extract this dependency to the `deps` in the root build.gradle file?

Demo/build.gradle
sshropshire: I've seen Android studio do this a few times I can never figure out why.

.idea/misc.xml
```diff
@@ -1,5 +1,18 @@
...
+        <entry key="../../../../../layout/compose-model-1631723355678.xml" value="0.67" />
+        <entry key="../../../../../layout/compose-model-1631731131942.xml" value="0.5638888888888889" />
+        <entry key="../../../../../layout/compose-model-1631886742866.xml" value="0.1" />
+        <entry key="../../../../../layout/compose-model-1631908005804.xml" value="1.0824074074074075" />
+        <entry key="Demo/src/main/res/layout/activity_demo.xml" value="0.325" />
+        <entry key="Demo/src/main/res/layout/fragment_pay_pal.xml" value="0.3671875" />
+        <entry key="Demo/src/main/res/xml/root_preferences.xml" value="0.32382246376811596" />
+      </map>
+    </option>
+  </component>
```
sshropshire, line 16: q: what does this file do? I'm never sure what from the `.idea` folder should be kept in the repo.

sshropshire (approved):
🌊  +1 on the imports.

.idea/misc.xml
tdchow: @sarahkoop and I were asking the same question. Jetbrains recommends keeping most files in the .idea directory in source control. https://intellij-support.jetbrains.com/hc/en-us/articles/206544839

Although I'm not exactly sure what the misc.xml file is used for. It seems that additional tags are added when resources are added to the project along with new composables.

sshropshire: Yeah it does kind of remind me of Xcode storyboards a bit. It jumped out during the review I'm fine with leaving as well.

Demo/build.gradle
sarahkoop: Yeah not sure how this snuck in there - removed it.

Demo/build.gradle
sarahkoop: I think this is a duplicate dependency - removed it 👍

tdchow (approved):
🚢 🚢 🚢

PR #17, opened 2021-09-23
CardClient Confirm Payment Source

sshropshire:
### Summary of changes

 - This PR adds `CardClient#confirmPaymentSource()` method that calls
   `/v2/checkout/orders/<ORDER_ID>/confirm-payment-source` endpoint

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @jcnoriega
- @sshropshire


Card/src/main/java/com/paypal/android/card/Card.kt
```diff
@@ -0,0 +1,3 @@
+package com.paypal.android.card
+
+data class Card(val number: String, val expiry: String)
```
tdchow: Do we need an additional value for security code?

Card/src/main/java/com/paypal/android/card/network/CardAPIClient.kt
```diff
@@ -0,0 +1,42 @@
...
+class CardAPIClient(private val api: APIClient) {
+
+    suspend fun confirmPaymentSource(orderId: String, card: Card): CardResult {
+        val path = "v2/checkout/orders/$orderId/confirm-payment-source"
+        val body = """
+            {
+                "payment_source": {
+                    "card": {
+                        "number": "${card.number}",
+                        "expiry": "${card.expiry}"
```
tdchow: @jcnoriega brought this up in our channel, but I think having a way to serialize request and response objects will be easier to maintain than raw strings of JSON.

Core/src/main/java/com/paypal/android/core/HttpMethod.kt
```diff
@@ -0,0 +1,6 @@
+package com.paypal.android.core
+
+enum class HttpMethod(val asString: String) {
```
tdchow: We could remove the `asString` field and use the `name` value built into Enums.

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-enum/name.html

Core/src/main/java/com/paypal/android/core/PaymentsConfiguration.kt
```diff
@@ -2,5 +2,6 @@ package com.paypal.android.core
 
 data class PaymentsConfiguration(
     val clientId: String,
-    val environment: Environment = Environment.SANDBOX
+    val environment: Environment = Environment.SANDBOX,
+    val authToken: String? = null
```
tdchow: Is `authToken` needed here? I think we can rely on the `clientId` for authorization.

Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt
```diff
@@ -4,9 +4,22 @@ import android.util.Log
...
 
 class CardViewModel : ViewModel() {
 
+    companion object {
+        private val TAG = CardViewModel::class.qualifiedName
+        private const val ORDER_ID = "45413380SJ116311E"
+        private const val CLIENT_ID =
+            "ASUApeBhpz9-IhrBRpHbBfVBklK4XOr1lvZdgu1UlSK0OvoJut6R-zPUP7iufxso55Yvyl6IZYV3yr0g"
+        private const val AUTH_TOKEN =
+            "A21AAKm8H7XlpB0LWC6hHDbEKOzn5bZzSbt74-QCOYl63WnnHWjKwNttWZyCPiY1tfRsB5PFWV6iBmmulC01Y5BleYXNlRGLw"
```
tdchow: I'm not sure if we should be checking in ClientIds or auth tokens. Can we remove these and add a placeholder?

Core/src/main/java/com/paypal/android/core/APIClient.kt
```diff
@@ -0,0 +1,18 @@
+package com.paypal.android.core
+
+class APIClient(private val configuration: PaymentsConfiguration) {
+
+    private val http = Http()
+    private val httpRequestFactory = HttpRequestFactory()
```
tdchow: Can we inject `http` and `httpRequestFactory` in the constructor? It should make passing instances much easier in our unit tests.

Core/src/main/java/com/paypal/android/core/PaymentsConfiguration.kt
jcnoriega: We realized that we can also use `authtoken` to make the request. This is just here for testing, because the request accepts encoded: `<client_id>:<secret>` or the `authtoken`. I think we need to some global variables to the sample app, that the merchant can input his `secret` and `clientId`.

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -1,6 +1,20 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 
-@Suppress("UnusedPrivateMember")
-class CardClient(paymentsClient: PaymentsClient)
+class CardClient(val configuration: PaymentsConfiguration) {
+
+    private val cardAPI: CardAPIClient = CardAPIClient(APIClient(configuration))
+
+    fun approveOrder(orderId: String, card: Card, completion: (CardResult) -> Unit) {
```
jonathajones: on iOS I believe we're using caps when referencing `..ID` objects - id `orderID` instead of `orderId`. not a big deal, just wanted to bring it up since it was a discussion on iOS

Card/src/main/java/com/paypal/android/card/network/CardAPIClient.kt
jcnoriega: i think we need to define the idea of just copying Braintree's SDK networking layer (that has Json Parsing) and put it in our SDK.
But in this case in particular, with @sshropshire we thought that if the queries are small, we could consider just putting the raw json like this. I know its not pretty code wise, BUT it almost looks like documentation, anybody who reads this, knows exactly the body of this API. Of course this is not super scalable for huge requests, but it might we worth perhaps sacrificing some code styling for better "readability". Does that make sense?

Card/src/main/java/com/paypal/android/card/Card.kt
jcnoriega: the `confirm_payment_source` query accepts `number` and `expiry` and thats it. Im not sure what the endpoint does though, is it just validating that you can use the card, or is it actually "paying" the order?

Core/src/main/java/com/paypal/android/core/HttpRequestFactory.kt
```diff
@@ -0,0 +1,26 @@
...
+        val url = URL("$baseUrl/$path")
+        val method = apiRequest.method
+        val body = apiRequest.body
+
+        val httpRequest = HttpRequest(url, method, body)
+        if (method == HttpMethod.POST) {
+            httpRequest.contentType = HttpContentType.JSON
+        }
+
+        httpRequest.headers["Authorization"] = "Bearer ${configuration.authToken}"
```
jonathajones: Aren't we going to use `"Basic <clientId>:"` for all of our network requests?

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -1,6 +1,20 @@
...
+import com.paypal.android.card.network.CardAPIClient
+import com.paypal.android.core.APIClient
+import com.paypal.android.core.PaymentsConfiguration
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 
-@Suppress("UnusedPrivateMember")
-class CardClient(paymentsClient: PaymentsClient)
+class CardClient(val configuration: PaymentsConfiguration) {
```
sarahkoop: Question - do we want to add doc strings as we go for this? I don't want to cause a bunch of rework since these classes are still in flux, but could be good to do it now to make it clear which classes / methods / fields will be merchant facing and which stuff is internal

Card/src/main/java/com/paypal/android/card/network/CardAPIClient.kt
sshropshire: Yeah this made the most sense to me because of the API knowledge gaps. We did this initially to get it to work, but I never thought about using heredoc template strings; they're super-expressive of intent here. There's a strong correlation with Postman as well.

sshropshire: Adding additional endpoints to the sdk in the short term would be super simple, and we can provide more abstraction once the requirements become clear.

Core/src/main/java/com/paypal/android/core/PaymentsConfiguration.kt
sshropshire: Yeah and IMO moving away from client secret as soon as possible is best. We'll decrease the risk of committing one of our test merchant's secret into GitHub. We implemented a script to fetch access-tokens in this PR for convenience.

Card/src/main/java/com/paypal/android/card/network/CardAPIClient.kt
```diff
@@ -0,0 +1,42 @@
...
+                    }
+                }
+            }
+        """.trimIndent()
+
+        val httpResponse = api.post(path, body)
+        return if (httpResponse.status == HTTP_OK) {
+            val bodyJson = httpResponse.body?.let { JSONObject(it) } ?: JSONObject()
+            val status = bodyJson.getString("status")
```
sarahkoop: Should these all be `optString` instead? I think we'd get errors if `bodyJson` is instantiated as an empty JSONObject object above

Core/src/main/java/com/paypal/android/core/PaymentsConfiguration.kt
sshropshire: @tdchow we also tried `Basic ${base64(clientId)}` auth and got `4xx` responses.

Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt
sshropshire: Yeah this was a hack but the auth token at least has an expiry. I agree we could inject `CLIENT_ID` and `AUTH_TOKEN` via a project-level, .gitignore'd `gradle.properties`.

Card/src/main/java/com/paypal/android/card/CardClient.kt
sshropshire: Yeah I know the caps `ID` is a bit of a convention on iOS. I'm down to achieve parity wherever possible.

Core/src/main/java/com/paypal/android/core/HttpRequestFactory.kt
sshropshire: Oh we never tried with a trailing `:` does this currently work for the `confirm-payment-source` route?

Card/src/main/java/com/paypal/android/card/CardClient.kt
sshropshire: That's a good callout. We could also start declaring classes `internal` where appropriate. I think `CardClient` and whatever is in its public interface will be merchant facing though.

Card/src/main/java/com/paypal/android/card/network/CardAPIClient.kt
sshropshire: Yeah that's true as well. Kotlin with the unchecked exceptions makes it tricky to spot those errors. We talked about the possibility of separating this out into a parser too. I guess we would need to return an error when the response is malformed.

Core/src/main/java/com/paypal/android/core/HttpRequestFactory.kt
jonathajones: `confirm-payment-source` currently requires an upgraded access token - on iOS we added a 'secret' field to be populated while activelytesting: https://github.com/paypal/iOS-SDK/pull/13/files#diff-c0a4c09ae6d8bbf7cdf018e06a195f26c3fd22604cc61e385cd2d1ab1400d04fR19-R20

Card/src/main/java/com/paypal/android/card/network/CardAPIClient.kt
```diff
@@ -0,0 +1,42 @@
...
+
+        val httpResponse = api.post(path, body)
+        return if (httpResponse.status == HTTP_OK) {
+            val json = PayPalJSON(httpResponse.body)
+
+            val status = json.optString("status")
+            val id = json.optString("id")
+            val lastDigits = json.optString("payment_source.card.last_digits")
+            val brand = json.optString("payment_source.card.brand")
+            val type = json.optString("payment_source.card.type")
```
sshropshire: We could do something like this 👆 . We're mostly pulling strings and numbers from the output I'd imagine. We do something similar in BT iOS. A few convenience wrappers allow us to keep our merchants' number of transitive dependencies small.

Core/src/main/java/com/paypal/android/core/HttpRequestFactory.kt
sshropshire: Ah yeah I remember now. We could go for something similar to keep consistent with iOS. Is the plan to only support client ID auth in our first release, or would the merchant be using access tokens?

jonathajones: Plan is support client ID basic auth on release, pending the API team lifting the requirement for the `confirm-payment-source` endpoint having an upgraded token.

Card/src/main/java/com/paypal/android/card/Card.kt
tdchow: The `confirm-payment-source` endpoint will attach a card (or other payment source) to an orderId. 

If you look at the ppaas documentation for this call, there are additional optional params that can be passed into the Card object (`security_code`, `card_type` etc.). I think we'll need to at least pass security code to the API for the first iteration of the SDK.

Card/src/main/java/com/paypal/android/card/network/CardAPIClient.kt
tdchow: That sounds good. I think for more complex request bodies we'll need to find a more scalable solution. I'm good with merging this in as is though since this call is pretty basic. 

I was thinking for optional params sent in the request body, things might get a little messier than just using string interpolation.

Core/src/main/java/com/paypal/android/core/PaymentsConfiguration.kt
tdchow: Got it. Currently the API will not accept the `Basic ${base64(clientId)}` header. Hopefully, the API team wraps up the changes to properly accept this auth header. Can we add a comment to call out that `authToken` is currently used for testing and that it'll be removed after the API changes are made?

Card/src/main/java/com/paypal/android/card/CardClient.kt
tdchow: I think we agreed to use `ID` (all caps) for naming in Northstar.

Card/src/main/java/com/paypal/android/card/Card.kt
sshropshire: Rebased onto main so the Card object should now be up to date.

Card/src/main/java/com/paypal/android/card/network/CardAPIClient.kt
sshropshire: Yeah I agree Jose voiced the same concern here. I also thought about the case where we may have to iterate over a collection of items to produce a `POST` body. The main thing here is postponing the additional abstraction layer until we need it. There are a few knowledge gaps on the API side that'll become clear as we add more features.

Core/src/main/java/com/paypal/android/core/PaymentsConfiguration.kt
sshropshire: Went ahead and added basic auth with `base64(CLIENT_ID:CLIENT_SECRET)`. We won't need to mention auth token going forward.

Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt
sshropshire: `paypal.properties` file can be used to load CLIENT_ID and CLIENT_SECRET without having to check them into the repo.

Card/src/main/java/com/paypal/android/card/CardClient.kt
sshropshire: Yup it's `ID` now.

Core/src/main/java/com/paypal/android/core/HttpRequestFactory.kt
sshropshire: Cool yeah we migrated to Basic Auth with `base64(CLIENT_ID:CLIENT_SECRET)`.

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -1,6 +1,30 @@
 package com.paypal.android.card
 
-import com.paypal.android.core.PaymentsClient
+import com.paypal.android.card.network.CardAPIClient
+import com.paypal.android.core.APIClient
+import com.paypal.android.core.PaymentsConfiguration
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
```
sarahkoop, line 8: I'm getting a missing dependency error on these imports, is my Android Studio wrong or did we lose a dependency somewhere?

Core/src/main/java/com/paypal/android/core/PayPalJSON.kt
```diff
@@ -0,0 +1,19 @@
+package com.paypal.android.core
+
+import org.json.JSONObject
+
+class PayPalJSON(private val json: JSONObject) {
```
sarahkoop: This looks good 👍  Maybe a more generic name (PaymentsJSON, CustomJSON, JSONParser?) would be helpful though? I feel like we should reserve the name `PayPal` for classes related to the PayPal payment method, or things will get confusing.

Demo/src/main/java/com/paypal/android/ui/card/CardFields.kt
```diff
@@ -14,18 +15,31 @@ import androidx.compose.ui.res.stringResource
...
 import androidx.compose.ui.text.input.TextFieldValue
+import androidx.compose.ui.unit.dp
 import com.paypal.android.R
 
 @Composable
 fun CardFields(
     cardViewModel: CardViewModel,
     modifier: Modifier = Modifier
 ) {
+    val orderID: String by cardViewModel.orderID.observeAsState("")
```
sarahkoop: Is this just added until we get the Demo app networking PR up?  If we want this demo app to be similar to a merchant app, having this as an input field doesn't seem right to me - can we have somewhere in the code where we can update this hard-coded value until the networking PR is complete?

Card/src/main/java/com/paypal/android/card/CardAPIRequestBuilder.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.paypal.android.core.HttpMethod
+
+internal class CardAPIRequestBuilder {
+
+    fun buildConfirmPaymentSourceRequest(orderID: String, card: Card): APIRequest {
+        val path = "v2/checkout/orders/$orderID/confirm-payment-source"
+
+        val cardNumber = card.number.replace("\\s".toRegex(), "")
+        val expirationDate = card.run {
+            val parsedSecurityCode = expirationDate.split("/")
```
tdchow: Should this val be named `parsedExpriationDate`?

Card/src/main/java/com/paypal/android/card/Card.kt
```diff
@@ -1,7 +1,7 @@
 package com.paypal.android.card
 
 data class Card(
-    val cardNumber: String,
-    val expirationDate: String,
-    val securityCode: String
+    var number: String = "",
+    var expirationDate: String = "",
+    var securityCode: String = ""
```
tdchow, line 26: Can we keep these as vals? do the variables need to change after creating a `Card` instance?

Demo/src/main/java/com/paypal/android/ui/card/CardFields.kt
tdchow: I agree with @sarahkoop. I think we should hardcode the order ID, until we're able to make calls to the merchant demo server to grab the generated order ID.

Card/src/main/java/com/paypal/android/card/Card.kt
```diff
@@ -1,7 +1,7 @@
 package com.paypal.android.card
 
 data class Card(
-    val cardNumber: String,
-    val expirationDate: String,
-    val securityCode: String
+    var number: String = "",
+    var expirationDate: String = "",
```
minhthenguyen: It looks to me we're expecting merchants to pass in a string of format `MM/YY` correct? 

In iOS we're having a `CardExpiry` class that takes in a `month: Int` and `year: Int` => then internally `CardExpiry` will be responsible for formatting the date to format `YYYY-MM` that our server expects. What do you think of doing that here? :smile:

minhthenguyen: And what do you think on naming `expirationDate` vs `expiry`?
we're doing `expiry` like in the request body but either works for me

Card/src/main/java/com/paypal/android/card/ConfirmedPaymentSource.kt
```diff
@@ -0,0 +1,9 @@
+package com.paypal.android.card
+
+data class ConfirmedPaymentSource(
```
minhthenguyen: I have an idea on the response type. Lemme know your thought and we can discuss more as a team.

minhthenguyen: Looking at all of the responses for all endpoints that we're gonna use, they look like the same object to me (just with or without some extra params)

For example, create order:
```
{
    "id": "1GP91429WA145883B",
    "status": "CREATED",
}
```

Get order's info:
```
{
    "id": "1GP91429WA145883B",
    "intent": "CAPTURE",
    "status": "APPROVED",
    "payment_source": {
        "card": {
            "last_digits": "7321",
            "brand": "VISA",
            "type": "CREDIT"
        }
    },
    "purchase_units": [
        {
            "reference_id": "default",
            "amount": {
                "currency_code": "USD",
                "value": "56.00"
            },
            "payee": {
                "email_address": "sb-jpbcl6949968@business.example.com",
                "merchant_id": "5DAA242G4YA4W"
            },
            "soft_descriptor": "TEST STORE"
        }
    ]
}
```

Confirm:
```
{
    "id": "1GP91429WA145883B",
    "status": "APPROVED",
    "payment_source": {
        "card": {
            "last_digits": "7321",
            "brand": "VISA",
            "type": "CREDIT"
        }
    }
}
```

Capture:
```
{
    "id": "1GP91429WA145883B",
    "status": "COMPLETED",
    "payment_source": {
        "card": {
            "last_digits": "7321",
            "brand": "VISA",
            "type": "CREDIT"
        }
    }
}
```

minhthenguyen: what do you think of having something like this:
```
data class OrderDataResponse(
  val orderID: String
  val status: OrderStatus     // I know this is String but we can do enum so it's friendlier to integration merchants
  val intent: Intent?    // Same with String to enum
  val payementSource: PaymentSourceResponse?     // ==> In this case can be CardResponse having last_digits, brands, etc. but can extends to other payment methods in the future
  val purchaseUnits: [PurchaseUnit]?
)
```

That way we can use the same response type for everything, so:
- it's clearer to both us and merchants
- no need to implement multiple response types that look kinda similar to each other.

But also downside:
- Can we assume this? maybe good to confirm with API team whether these responses are the same object from the server side? 
- what if our API changes these format in the future?

minhthenguyen: sorry for the long comment, but please lemme know your thoughts :smile:

minhthenguyen:
This looks great!!!! 🚀 
No change requests from me, just some questions to make our interface matched and discussion.

Card/src/main/java/com/paypal/android/card/ConfirmedPaymentSource.kt
sarahkoop: I think this would be good to discuss more as a team - a single response object seems like it would work well now, but feels like it will cause problems later if we have to force all payment method responses to conform to one object type. Also depending on API changes, it could lead to this class/module needing to know details about specific payment methods that exist outside this module(ex: if the response for a Venmo order contains venmo-specific fields, we would have to update this), which we want to avoid. 

Another idea is a having a `BaseOrderDataResponse` class that lives in the core module that could follow the current expected pattern, and a `CardOrderDataResponse` that extends it and contains any card-specific fields

Core/src/main/java/com/paypal/android/core/PayPalJSON.kt
sshropshire: Yeah this is a good call i'll rename. I like`PaymentsJSON` to mirror the `BTJSON` concept.

Demo/src/main/java/com/paypal/android/ui/card/CardFields.kt
sshropshire: Ok cool I can revert.

Card/src/main/java/com/paypal/android/card/Card.kt
sshropshire: Yeah we did something similar with `ExpiryDate`. I think `expirationDate` will probably read better for merchants, but technically either could work.

Card/src/main/java/com/paypal/android/card/Card.kt
sshropshire: We can I remember David saying `Card` has a ton of optional fields. I know on the BT side we sometimes skip security codes. It might be good to re-visit in that next ticket for adding card fields.

Card/src/main/java/com/paypal/android/card/CardClient.kt
sshropshire: I think this was intermittent, hopefully should be good now.

Card/src/main/java/com/paypal/android/card/ConfirmedPaymentSource.kt
sshropshire: Yeah I like the idea of having more simplified types to make things easier to comprehend for the merchant. We have done similar things in BT Android with base types. But overall I agree, keeping things similar across different payment methods will probably lead to increased adoption.

Card/src/main/java/com/paypal/android/card/Card.kt
minhthenguyen: yep I think `expirationDate` is clearer to merchants too. I can change `CardExpiry` -> `ExpirationDate` in iOS.

Question: Are we able to use the new `CardExpiry` here instead of `String`? 
And also what do you think of `CardExpiry(val month: Int, val year: Int)` vs `CardExpiry(val month: String, val year: String)` ?

sshropshire: Yeah I'm down with storing month and year as a string. I guess I envisioned this class as being internal though. The best developer experience would probably be for the merchant to provide a string IMO.

sshropshire: But there are some concerns. We would probably need to surface an error if the merchant were to provide something in an incorrect format.

minhthenguyen: I see you just changed it to Int [here](https://github.com/paypal/Android-SDK/commit/0b885993afc4e37f79e95f7c8cce625b667abd71#diff-9376f8fdbbaec31b90530d944ffcf2ee3211b4e44753261bd81b1023149532e4R3). I think it looks great, we're having the same `ExpirationDate(month: Int, year: Int)` in iOS :smile:

minhthenguyen: just wondering should the line `var expirationDate: String` above be changed to `var expirationDate: ExpirationDate` instead?

Card/src/main/java/com/paypal/android/card/Card.kt
tdchow: I think we can still keep these fields as values instead of variables and keep them optional in the constructor. As long as there's a default value, a merchant is able to omit that field from the constructor.

```kotlin
data class Card(
    val number: String = "",
    val expirationDate: String = "",
    val securityCode: String = ""
)

val card = Card("1234512345", "02/2022")
```

This would just prevent the modification of each value after the card instance is created.

Card/src/main/java/com/paypal/android/card/ConfirmPaymentSourceResult.kt
```diff
@@ -0,0 +1,9 @@
+package com.paypal.android.card
+
+import com.paypal.android.core.OrderData
+import com.paypal.android.core.OrderError
+
+data class ConfirmPaymentSourceResult(
+    val response: OrderData? = null,
+    val error: OrderError? = null
+)
```
scannillo, line 9: We should align across iOS & Android how we name these results & how they're structured. I was thinking of the Core module having something like

`PaymentResult` (has the standard order & error) and then subclasses in each module like `CardResult`, `VenmoResult` etc that can extend each and provide additional params.

I know @minhthenguyen had another idea. We should definitely keep this naming and structure in parity across iOS and Android.

sshropshire: Yeah I'm down for unifying the structure. The response objects do seem pretty uniform I think this will help to improve the developer experience.

minhthenguyen (approved):
This looks great to me!!!! :rocket:

demerino (approved):
Solid foundation. At some point we will want to align on naming and overall alignment, but we should wait to do that. 👍

andrespelaezp (approved):

Card/src/main/java/com/paypal/android/card/Card.kt
tdchow: Following up here, Can we convert these vars to vals? Also from looking at the documentation for the confirm call, `number` and `expiration` are required fields. Should we remove the default values here?

tdchow (approved):
Looks good! Just a question about converting some vars to vals in a data class, but it shouldn't hold up the PR!

sshropshire:
> Following up here, Can we convert these vars to vals? Also from looking at the documentation for the confirm call, `number` and `expiration` are required fields. Should we remove the default values here?

@tdchow done!

PHEON1XENT (commented):
CloudtNine

PR #19, opened 2021-09-24
[DTNOR-151] Demo App - Auto fill card fields with test data

tdchow:

device-2021-09-24-155159

### Summary of changes

 - Added the ability to automatically fill card fields with test card data
 - Added string formatting to the card and expiration date fields

 ### Checklist

- @sarahkoop 
- @andrespelaezp 
- @tdchow


Card/src/main/java/com/paypal/android/card/Card.kt
```diff
@@ -0,0 +1,7 @@
+package com.paypal.android.card
+
+data class Card(
```
tdchow, line 3: We may have some conflicts with the current Card PR. But they should be simple to resolve.

Demo/src/main/java/com/paypal/android/ui/card/validation/CardType.kt
```diff
@@ -0,0 +1,30 @@
+package com.paypal.android.ui.card.validation
+
+@Suppress("MagicNumber")
+enum class CardType(
```
tdchow, line 4: We should be able to reuse this enum once we look at supporting Card Fields in the SDK, but for now, I kept this as simple as possible to only handle formatting card numbers and max card lengths.

Demo/src/main/java/com/paypal/android/ui/card/validation/CardType.kt
```diff
@@ -0,0 +1,30 @@
...
+    ),
+    UNKNOWN_CARD(
+        cardNumberIndices = listOf(4, 9, 14),
+        maxCardLength = 16
+    )
+}
+
+fun getCardType(cardNumber: String): CardType {
+    return when {
+        cardNumber.startsWith("34") || cardNumber.startsWith("37") -> {
```
sarahkoop, line 24: This might get a little unruly since there are so many different number patterns for some card types. I think it's ok to leave it like this while it's just for the demo app but something to think about if we use this for determining card types in the SDK.

sarahkoop (approved):
This looks great 🌟

Demo/src/main/java/com/paypal/android/ui/card/validation/CardType.kt
tdchow: Agreed! I was thinking of moving these IIN ranges into the `CardType` enum and having a function live within each card type, but thought that this was sufficient for the demo app.

Card/src/main/java/com/paypal/android/card/Card.kt
sshropshire: Yeah we only have `cardNumber` and `expirationDate` at the moment.

Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt
```diff
@@ -36,8 +44,12 @@ class CardViewModel : ViewModel() {
         // Invoke Card SDK
     }
 
-    private fun formatExpirationDate(expirationDate: String): String {
-        return if (expirationDate.length == 2) "$expirationDate/" else expirationDate
+    fun onPrefillCardSelected(cardName: String) {
+        autoFillCards.find { it.first == cardName }?.second?.apply {
```
sshropshire, line 48: qq: does this pick the second card?

sshropshire (approved):
🌊

Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt
tdchow: `autoFillCards` is a list of pairs, so on this line `.second` would be referencing the `Card` object.

PR #20, opened 2021-09-28
DTNOR-150 - Demo App Networking Layer

andrespelaezp:

### Summary of changes

 - Networking layer structure for the Demo App
 - Used Hilt for Dependency injection strategy and Retrofit for network calls

 ### Checklist

 - [ ] Added a changelog entry

-


Demo/src/main/java/com/paypal/android/di/NetworkModule.kt
```diff
@@ -0,0 +1,39 @@
...
+    val okHttpBuilder = OkHttpClient.Builder()
+    // Timeouts
+    okHttpBuilder.connectTimeout(CONNECT_TIMEOUT_IN_SEC, TimeUnit.SECONDS)
+        .readTimeout(READ_TIMEOUT_IN_SEC, TimeUnit.SECONDS)
+        .writeTimeout(WRITE_TIMEOUT_IN_SEC, TimeUnit.SECONDS)
+
+    return okHttpBuilder.build()
+}
+
+private fun prepareTokenHeaderValue(token: String): String {
```
tdchow: It doesn't look like this function is used. Is it still needed?

build.gradle
```diff
@@ -40,6 +47,13 @@ buildscript {
             // Jetpack
             "preference"                  : "androidx.preference:preference-ktx:1.1.1",
 
+            // Networking
+            "retrofit"                    : "com.squareup.retrofit2:retrofit:${versions.retrofit}",
+            "gson"                        : "com.squareup.retrofit2:converter-gson:${versions.gson}",
+
+            // Dependency Injection
+            "koin"                        : "io.insert-koin:koin-android:${versions.koin}",
```
tdchow: We should be safe declaring the version numbers here within the `ext.deps` array since these version numbers are only referenced once. I think we have the pattern of adding values to `ext.versions` when that specific library version is referenced more than once.

```suggestion
// Networking
            "retrofit"                    : "com.squareup.retrofit2:retrofit:2.9.0",
            "gson"                        : "com.squareup.retrofit2:converter-gson:2.3.0",

            // Dependency Injection
            "koin"                        : "io.insert-koin:koin-android:3.1.2",
```

tdchow (approved):
Changes are looking good! Just had a question and suggestion for the build.gradle file. We can also ask the rest of the team which dependency injection tool they want to go with.

Demo/src/main/java/com/paypal/android/di/NetworkModule.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.google.gson.GsonBuilder
+import okhttp3.OkHttpClient
+import org.koin.dsl.module
+import java.util.concurrent.TimeUnit
+
+private const val AUTH_HEADER_VALUE_START = "Token "
+
+private const val CONNECT_TIMEOUT_IN_SEC = 20L
+private const val READ_TIMEOUT_IN_SEC = 30L
+private const val WRITE_TIMEOUT_IN_SEC = 30L
```
sarahkoop, line 15: Just curious - how did we decide on these timeout values? Are they standard?

Demo/build.gradle
```diff
@@ -67,6 +67,13 @@ dependencies {
     implementation deps.navigationUI
     implementation deps.preference
 
+    // Networking
+    implementation deps.retrofit
+    implementation deps.gson
+
+    // Dependency Injection
+    implementation deps.koin
```
sarahkoop: Nitpick - not sure we need comments here (since we don't have them for any other dependencies and we comment what they are for in the other file)

Demo/src/main/java/com/paypal/android/di/NetworkModule.kt
andrespelaezp: nice catch!

Demo/src/main/java/com/paypal/android/di/NetworkModule.kt
andrespelaezp: These are just values I have previously used, is there any PayPal standard times so I can update them?

Demo/src/main/java/com/paypal/android/di/ApiModule.kt
```diff
@@ -0,0 +1,18 @@
+package com.paypal.android.di
```
sarahkoop: Potentially dumb question - what does di mean here?

andrespelaezp: Not dumb at all, is short for Dependency Injection, would you like to see it in a different package or with a more explicit name?

sarahkoop: Hmmm.... open to other opinions here (maybe di is obvious) but I think dependency-injection might be more helpful as a package name

Demo/src/main/java/com/paypal/android/di/NetworkModule.kt
sarahkoop: Not sure of standard times - in Braintree currently we use 30 seconds for all, but 20 is good with me to

tdchow: We have 60-second timeouts in the PayPal Checkout SDK, but I don't think there's a company-wide standard. I think what you have in the demo module is good.

Demo/src/main/java/com/paypal/android/di/ApiModule.kt
tdchow: My vote is for keeping the package name `di`. I've seen this pattern the most in projects with dependency injection. But maybe this is something we can poll the team on.

Demo/src/main/java/com/paypal/android/DemoApplication.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.paypal.android.di.networkModule
+import org.koin.android.ext.koin.androidContext
+import org.koin.core.context.startKoin
+
+class DemoApplication : Application() {
+
+    override fun onCreate() {
+        super.onCreate()
+
+        startKoin {
```
sshropshire: randomq: is there a reason we chose Koin over Hilt?

andrespelaezp: same question came up when syncing yesterday, it was more like a quick implementation decision, we thought about discussing this with the SDK team to decide whether we would move to Hilt or leave it as it is, what are your thoughts? I could move it to Hilt if you guys think it is the best approach

tdchow: My vote is for Hilt! I've run into runtime issues with Koin which would have been caught with Dagger/Hilt.

sshropshire: I personally prefer Jetpack libraries since there's consistency within the Jetpack ecosystem. Interop with other Jetpack libraries for example. I'm not sure how others feel, but I'd vote for Hilt for `di`.

Demo/src/main/java/com/paypal/android/DemoApplication.kt
```diff
@@ -0,0 +1,12 @@
+package com.paypal.android
+
+import android.app.Application
+import dagger.hilt.android.HiltAndroidApp
+
+@HiltAndroidApp
+class DemoApplication : Application() {
+
+    override fun onCreate() {
```
tdchow: It doesn't look were overriding anything in this `onCreate`. Maybe we can clean it up and add it back when we need?

Demo/src/main/java/com/paypal/android/di/NetworkModule.kt
```diff
@@ -1,33 +1,91 @@
...
+        val sslSocketFactory: SSLSocketFactory = sslContext.socketFactory
+        val builder = OkHttpClient.Builder()
+        builder.sslSocketFactory(
+            sslSocketFactory,
+            trustAllCerts[0] as X509TrustManager
+        )
+        builder.hostnameVerifier { _, _ -> true }
+        return builder
+    }
+}
```
andrespelaezp: Just a small clarification about this method: since our backend repo creates a SSL server with Heroku, seems like it is not trusted or well configured, so this would be like a workaround to make it work

tdchow (approved):
The Hilt integration is looking great! I just had a small clean-up item, but no need to hold the PR.

sshropshire (approved):

PR #21, opened 2021-10-07
Card Add Additional Properties

sshropshire:
### Summary of changes

 - Add Properties to `Card`:
   * `cardholderName`
   * `billingAddress`
- Make `securityCode` property optional on `Card`

 ### Checklist

 - [x] Added a changelog entry

- @andrespelaezp
- @sshropshire


Card/src/main/java/com/paypal/android/card/Card.kt
```diff
@@ -23,5 +25,20 @@ data class Card(
     /**
      * The card's security code (CVV, CVC, CVN, CVE, or CID)
      */
-    var securityCode: String = ""
+    var securityCode: String? = null,
+
+    /**
+     * The card holder's name as it appears on the card
+     */
+    var name: String? = null,
```
scannillo:
```suggestion
    var cardholderName: String? = null,
```

Name might be confused with something optional like a "nickname"

Card/src/main/java/com/paypal/android/card/Card.kt
```diff
@@ -1,5 +1,7 @@
 package com.paypal.android.card
 
+import com.paypal.android.core.Address
+
```
scannillo, line 4: Note: We should be marking the `optional` fields appropriately in these docstrings. Example:

```
    /**
     * Optional. The card holder's name as it appears on the card
     */
```

Card/src/main/java/com/paypal/android/card/Card.kt
```diff
@@ -23,5 +25,20 @@ data class Card(
...
+    /**
+     * The portable international postal address
+     */
+    var billingAddress: Address? = null,
+
+    /**
+     * Additional attributes associated with the use of this card
+     */
+    var attributes: CardAttributes? = null
```
scannillo: It feels a little arbitrary which params are in `CardAttributes` vs in `Card`

scannillo: We could have a `CardRequest` be the param to the `confirmPayment` function. And then `CardRequest` has a required Card and then optional additional params, such as billingAddress, email, etc.?

Another idea is everything just top level on one request obj.

scannillo:
@minhthenguyen started a branch for the equivalent work on iOS here (https://github.com/paypal/iOS-SDK/compare/card-optional-params). We should align on integration pattern and naming!

Card/src/main/java/com/paypal/android/card/Card.kt
minhthenguyen: another note is that `CardAttributes` is marked `LIMITED_RELEASED`
If it's like how `validate-payment-method` being limited vs `confirm-payment-source` being publicly exposed, I wonder maybe `CardAttributes` was also created only for some internal use case, and maybe we don't need to expose this to merchants? Maybe we should confirm that and figure out what it's really used for :smile:

sshropshire: I agree the properties on `CardAttributes` do feel like they're separate from an actual `Card` object. I also agree it may be a good rule of thumb going forward to support Orders `v2` features that aren't marked as `LIMITED_RELEASE`. We can always add those back on request.

Card/src/main/java/com/paypal/android/card/CardAPIRequestFactory.kt
```diff
@@ -2,26 +2,37 @@ package com.paypal.android.card
...
-            }
-        """.trimIndent()
+        val cardJSON = JSONObject()
+            .put("number", cardNumber)
+            .put("expiry", cardExpiry)
+
+        card.cardholderName?.let { cardJSON.put("name", it) }
+        card.securityCode?.let { cardJSON.put("security_code", it) }
+
+        card.billingAddress?.let { billingAddress ->
```
tdchow: Instead of using the `let` scope function here, `apply` could be used. This way we'd have `this` as the object reference in the lambda, allowing us to clean up `billingAddress` from each of the calls to `.put()`.

tdchow (approved):
The changes look good! Just had a minor suggestion that shouldn't hold the PR.

Card/src/main/java/com/paypal/android/card/CardAPIRequestFactory.kt
sshropshire: Definitely yeah that's a good call. It does clean it up nicely.

scannillo (approved):
Changes look good! Think we should revisit the convo of dedicated `CardRequest`, `PayPalRequest` top level objects per payment method, for the future though instead of putting all on `Card`.

PR #23, opened 2021-10-20
DTNOR-203 - Align result types for Card

tdchow:

### Summary of changes

 - Created `CardResult` sealed class that contain `Success` and `Error` types
 - Passed back correlation ID for the error case

 ### Checklist

- @sshropshire 
- @tdchow


Card/src/main/java/com/paypal/android/card/CardOrder.kt
```diff
@@ -0,0 +1,9 @@
+package com.paypal.android.card
+
+import com.paypal.android.core.OrderData
+import com.paypal.android.core.OrderStatus
+
+data class CardOrder(
```
scannillo: I think this file can be removed

Card/src/main/java/com/paypal/android/card/CardResult.kt
```diff
@@ -0,0 +1,17 @@
...
+sealed class CardResult {
+
+    class Success(
+        val orderID: String,
+        val status: OrderStatus
+    ) : CardResult()
+
+    class Error(
+        val orderError: OrderError,
+        val correlationID: String?
```
scannillo, line 15: Did we want to align on the details returned in the error across [iOS](https://github.com/paypal/iOS-SDK/blob/main/Sources/PaymentsCore/PayPalSDKError.swift) and Android in this PR as well?

tdchow: @andrespelaezp and @ndmunoz10 are currently working on the error story on Android, so I think those changes would be in a separate PR. But I think we can chat and make sure we're aligned with what you all have on iOS!

Card/src/main/java/com/paypal/android/card/CardAPI.kt
```diff
@@ -12,17 +11,28 @@ internal class CardAPI(
...
+                CardResult.Success(
+                    orderID = id,
+                    status = OrderStatus.valueOf(status)
+                )
             }.recover {
-                ConfirmPaymentSourceResult(error = OrderError("PARSING_ERROR", "Error parsing json response."))
+                CardResult.Error(
+                    orderError = OrderError(
+                        "PARSING_ERROR",
+                        "Error parsing json response."
```
andrespelaezp, line 32: Just out of curiosity, do we follow a specific standard for arguments naming? Since there are three expected and only two are provided, should the namings be included?

Card/src/main/java/com/paypal/android/card/CardResult.kt
andrespelaezp: @tdchow gotcha!

Card/src/main/java/com/paypal/android/card/CardAPI.kt
tdchow: I don't think we have a standard for named parameters. The compiler is still happy in this case since there's a default value on the last `details` param.

But I'm all for using named params. While we get some help in Android Studio when param names are omitted, having them specifically defined helps in PR review where we don't have the IDE's help.

andrespelaezp (approved):

ndmunoz10 (approved):

PR #24, opened 2021-10-26
DTNOR-192 - Card error handling

andrespelaezp:

### Summary of changes

 - Created APIClientError with PayPalSDKError types to handle confirmPaymentSource response

 ### Checklist

 - [ ] Added a changelog entry

- ndmunoz10
- andrespelaezp


Card/src/main/java/com/paypal/android/card/APIClientError.kt
```diff
@@ -0,0 +1,68 @@
...
+        errorDescription = "An error occurred constructing an HTTP request. Contact developer.paypal.com/support."
+    )
+
+    // 5. The server's response body returned an error message.
+    val serverResponseError = PayPalSDKError(
+        code = Code.SERVER_RESPONSE_ERROR.ordinal,
+        domain = domain,
+        errorDescription = "A server occurred. Contact developer.paypal.com/support."
+    )
+
```
scannillo:
```suggestion
```

Card/src/main/java/com/paypal/android/card/PayPalSDKError.kt
```diff
@@ -0,0 +1,7 @@
+package com.paypal.android.card
+
+data class PayPalSDKError(
+    val code: Int?,
+    val domain: String?,
+    val errorDescription: String?
```
scannillo: Are error code, domain, & description common pattern for Android errors? We used that on iOS as it follows Apple's `NSError` pattern and is comfortable/common to iOS devs.

andrespelaezp: The domain is not something that is common, Nico and I discussed about this and left it as it was implemented in iOS since we thought it was more a business logic. But I guess we can remove it then

ndmunoz10: Yeah we'll remove it then

Card/src/main/java/com/paypal/android/card/CardAPI.kt
```diff
@@ -12,33 +14,80 @@ internal class CardAPI(
...
+        if (URLUtil.isValidUrl(apiRequest.path).not()) {
+            CardResult.Error(
+                payPalSDKError = APIClientError.invalidUrlRequest,
+                correlationID = correlationId
+            )
+        }
+        val bodyResponse = httpResponse.body
+        if (bodyResponse.isNullOrBlank()) {
+            return CardResult.Error(
+                payPalSDKError = APIClientError.noResponseData,
```
sshropshire: qq: should this be considered an error? Thinking of [204 No Content](https://httpstatusdogs.com/204-no-content) here.

andrespelaezp: Do you know what is the respective error Class for this on Java? This is so we can map it on "Http" data class. 
Ps. nice pic from status dogs haha

sshropshire: Ah nice. Yeah haha that's a good site. I see what you're saying though I forgot this was `CardAPI`. It makes perfect sense to return an error in that case.

sshropshire (approved):
🌊

Card/src/main/java/com/paypal/android/card/APIClientError.kt
```diff
@@ -0,0 +1,57 @@
...
+    // 5. The server's response body returned an error message.
+    val serverResponseError = PayPalSDKError(
+        code = Code.SERVER_RESPONSE_ERROR.ordinal,
+        errorDescription = "A server occurred. Contact developer.paypal.com/support."
+    )
+
+    // Error returned from HttpURLConnection while making request.
+    val httpURLConnectionError: (code: Int, description: String) -> PayPalSDKError = { code, description ->
```
tdchow: Would a function that takes in `code` and `description` as params and returns a `PayPalSDKError` object be more readable here?

Card/src/main/java/com/paypal/android/card/CardAPI.kt
```diff
@@ -15,43 +18,54 @@ internal class CardAPI(
         val apiRequest = requestFactory.createConfirmPaymentSourceRequest(orderID, card)
         val httpResponse = api.send(apiRequest)
         val correlationId = httpResponse.headers["Paypal-Debug-Id"]
-        return if (httpResponse.status == HTTP_OK) {
-            runCatching {
-                val json = PaymentsJSON(httpResponse.body)
-                val status = json.getString("status")
-                val id = json.getString("id")
+        if (URLUtil.isValidUrl(apiRequest.path).not()) {
```
tdchow: Would using `!` instead of `.not()` be clearer here?

Card/src/main/java/com/paypal/android/card/APIClientError.kt
```diff
@@ -0,0 +1,57 @@
+package com.paypal.android.card
```
tdchow: Are these errors something that could be used across all SDK modules? Would it make more sense for them to live in the core module?

Core/src/main/java/com/paypal/android/core/Http.kt
```diff
@@ -47,6 +48,7 @@ internal class Http(
             }.recover {
                 val status = when (it) {
                     is UnknownHostException -> HttpResponse.STATUS_UNKNOWN_HOST
+                    is IllegalStateException -> HttpResponse.SERVER_ERROR
```
tdchow, line 51: Could we add a unit test for this additional case?

Card/src/main/java/com/paypal/android/card/APIClientError.kt
andrespelaezp: We made it this way to be consistent with the rest of the errors, but I think it is a good suggestion, what do you think @ndmunoz10?

Card/src/main/java/com/paypal/android/card/CardAPI.kt
andrespelaezp: Sounds good!

Card/src/main/java/com/paypal/android/card/APIClientError.kt
ndmunoz10: I'd prefer if possible to maintain a val property here since it's cheaper and It keeps consistency along with the other declared errors

tdchow: Makes sense and sounds good!

Core/src/main/java/com/paypal/android/core/Http.kt
andrespelaezp: I realized this is covered on _`it returns serverError when the order api call returns an error body`_, it would be the status number "-3".

tdchow (approved):
Thanks for making the updates. The PR looks good!

PR #28, opened 2021-11-12
Use Java Style Callback for CardClient

sshropshire:
### Summary of changes

 - Use Java Style Callback in CardClient
 - This will eliminate the need for merchants to return `null` when implementing a callback in Java

 ### Checklist

### DIFF
--- a/Card/src/main/java/com/paypal/android/card/CardResultCallback.java
+++ b/Card/src/main/java/com/paypal/android/card/CardResultCallback.java
@@ -1,0 +1,7 @@
+package com.paypal.android.card;
+
+import androidx.annotation.NonNull;
+
+public interface CardResultCallback {
+    void onCardResult(@NonNull CardResult result);
+}
--- a/Card/src/main/java/com/paypal/android/card/CardClient.kt
+++ b/Card/src/main/java/com/paypal/android/card/CardClient.kt
@@ -10,10 +10,7 @@ import kotlinx.coroutines.launch
 /**
  * Use this client to approve an order with a [Card].
  */
-class CardClient internal constructor(
-    private val cardAPI: CardAPI,
-    private val dispatcher: CoroutineDispatcher = Dispatchers.Main
-) {
+class CardClient internal constructor(private val cardAPI: CardAPI) {
 
     constructor(configuration: CoreConfig) :
             this(CardAPI(API(configuration)))
@@ -25,14 +22,6 @@ class CardClient internal constructor(
      * @param card The card to use for approval
      * @param completion A completion handler for receiving a [CardResult]
      */
-    fun approveOrder(
-        orderID: String,
-        card: Card,
-        completion: (CardResult) -> Unit
-    ) {
-        CoroutineScope(dispatcher).launch {
-            val cardResult = cardAPI.confirmPaymentSource(orderID, card)
-            completion(cardResult)
-        }
-    }
+    suspend fun approveOrder(orderID: String, card: Card) =
+        cardAPI.confirmPaymentSource(orderID, card)
 }
--- a/Card/src/main/java/com/paypal/android/card/CardClient.kt
+++ b/Card/src/main/java/com/paypal/android/card/CardClient.kt
@@ -2,10 +2,6 @@ package com.paypal.android.card
 
 import com.paypal.android.core.API
 import com.paypal.android.core.CoreConfig
-import kotlinx.coroutines.CoroutineDispatcher
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.launch
 
 /**
  * Use this client to approve an order with a [Card].
@@ -20,7 +16,7 @@ class CardClient internal constructor(private val cardAPI: CardAPI) {
      *
      * @param orderID The id of the order
      * @param card The card to use for approval
-     * @param completion A completion handler for receiving a [CardResult]
+     * @return [CardResult]
      */
     suspend fun approveOrder(orderID: String, card: Card) =
         cardAPI.confirmPaymentSource(orderID, card)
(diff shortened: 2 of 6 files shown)


tdchow:
I think we'll still need to keep in the callback pattern here. Only supporting coroutines will set a hard requirement for merchants. Merchants who are not using coroutines for background work wouldn't be able to use the SDK. One example of this is a merchant who is using RxJava.

sshropshire:
> I think we'll still need to keep in the callback pattern here.

@tdchow I agree that we should offer support for callbacks. I'm concerned about how Java onboarding looks at the moment. The current implementation requires merchants to return `null` when implementing a callback:

```java
CoreConfig config = new CoreConfig(BuildConfig.CLIENT_ID, BuildConfig.CLIENT_SECRET, Environment.SANDBOX);
CardClient cardClient = new CardClient(config);

Card card = new Card("4111111111111111", "01", "22", null, null, null);
cardClient.approveOrder("123", card, result -> {
  if (result instanceof CardResult.Success) {
    CardResult.Success success = (CardResult.Success) result;
    Log.d(TAG, "SUCCESS");
    Log.d(TAG, String.format("Status: %s", success.getStatus().name()));
  }
  return null;
});
```

Coroutines are [recommended by Google](https://developer.android.com/kotlin/coroutines#features), and we already use them internally. By making coroutines first-class in the SDK, we can focus directly on integrating with the Orders API. Then, we can come up with a sound strategy for providing explicit callback / Java support before going to GA.

sshropshire:
@tdchow there may be a middle ground. By using a Java callback interface, we eliminate the need for the merchant to return `null` when implementing a callback.

tdchow:
This sounds good to me! This is similar to what we have in the Native Checkout SDK.

Card/src/main/java/com/paypal/android/card/CardResultCallback.java
```diff
@@ -0,0 +1,7 @@
+package com.paypal.android.card;
+
+import androidx.annotation.NonNull;
+
+public interface CardResultCallback {
```
tdchow, line 5: Can we convert this interface to Kotlin? I believe the Kotlin equivalent will still work for Java integrations.

```kotlin
interface CardResultCallback {
    fun onCardResult(result: CardResult)
}
```

sshropshire: @tdchow @andrespelaezp I agree that a pure Kotlin codebase is appealing, but if we do that, we lose the function-like interface in Kotlin:

```
cardClient.approveOrder(orderId, card, object : CardResultCallback {
  override fun onCardResult(result: CardResult) {
    when (result) {
      is CardResult.Success -> {
        // handle success
      }
      is CardResult.Error -> {
        // handle error
      }
    }
  }
})
```

sshropshire: Java looks great with a Kotlin interface though, but Kotlin would take a hit if we use non-Java callbacks. I thinks that's the reverse of what we want. Java-style callbacks give us the best of both worlds.

scannillo: Maybe we can put together what all 4 of the code samples would look like. So 2 snippets (Java & Kotlin) for how this looks for a merchant if the source code is written in Java, and 2 snippets (Java & Kotlin) for how this looks for a merchant if the source code is written in Kotlin.

This would help, me at least, visualize our options.

scannillo: Correct me if I'm wrong, but it sounds like what Steven is advocating for is to **not** require our Java merchants to do a cast like `if (result instanceof CardResult.Success)`, and not force them to return `null`. Currently, on one hand our Kotlin integration looks super slick - but our Java integration is clunky. This solution proposes a middle ground. Again - I think the explicit snippets next to each other would help visualize.

tdchow: > Java looks great with a Kotlin interface though, but Kotlin would take a hit if we use non-Java callbacks. I thinks that's the reverse of what we want. Java-style callbacks give us the best of both worlds.

Ah right, I remember running into this while we were working on the 3P NXO SDK. I think the only caveat with going with a Java defined interface is that the Kotlin call site requires the interface type to be defined.

```kotlin
CardClient().approveOrder { data ->

}

// vs

CardClient().approveOrder(JavaInterface { data ->

})
```

Here's what we did in the NXO SDK to get around the issue:

```kotlin
class CardClient {

    fun approveOrder(testInterface: TestInterface) {
        testInterface.testFunction("")
    }

    fun approveOrder(callback: (String) -> Unit) {
        approveOrder(object : TestInterface {
            override fun testFunction(field: String) {
                callback.invoke(field)
            }
        })
    }
}
```
We have multiple functions on the object to satisfy both languages. One of the issues with this pattern is that a merchant could call the wrong function for their specific language. 

I came across this section in the Kotlin-Java interop guide in the developer docs that calls out this exact case: https://developer.android.com/kotlin/interop#lambda_arguments To Sammy's point, I think it would help to have integration examples for each of the solutions.

sshropshire: It actually compiles as-is right now. The `CardViewModel` code is unchanged [here](https://github.com/paypal/Android-SDK/blob/3d73756b6000ee1298f4929e39c22be197e30d67/Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt#L81). I have seen that style before, but in this case, it looks like Kotlin perceives a Function style callback.

tdchow: Ah, you're right! Looks like this was added in Kotlin 1.4: https://kotlinlang.org/docs/whatsnew14.html#sam-conversions-for-kotlin-interfaces 

I was originally testing with an older version of Kotlin.

tdchow (approved):

sarahkoop (approved):

tdchow (approved):
PayPal updates are looking good!

PR #30, opened 2021-12-02
Card Request Parameter Object

sshropshire:
### Summary of changes

 - Add a `CardRequest` [parameter object](https://wiki.c2.com/?ParameterObject) to `CardClient.approveOrder` method

 ### Checklist

 - [x] Added a changelog entry

### DIFF
--- a/Card/src/main/java/com/paypal/android/card/CardRequest.kt
+++ b/Card/src/main/java/com/paypal/android/card/CardRequest.kt
@@ -1,0 +1,3 @@
+package com.paypal.android.card
+
+data class CardRequest(val orderID: String, val card: Card)
--- a/Card/src/main/java/com/paypal/android/card/CardClient.kt
+++ b/Card/src/main/java/com/paypal/android/card/CardClient.kt
@@ -21,17 +21,15 @@ class CardClient internal constructor(
     /**
      * Confirm [Card] payment source for an order.
      *
-     * @param orderID The id of the order
-     * @param card The card to use for approval
-     * @param completion A completion handler for receiving a [CardResult]
+     * @param request [CardRequest] for requesting an order approval
+     * @param callback [CardResultCallback] for receiving a [CardResult]
      */
     fun approveOrder(
-        orderID: String,
-        card: Card,
+        request: CardRequest,
         callback: CardResultCallback
     ) {
         CoroutineScope(dispatcher).launch {
-            val cardResult = cardAPI.confirmPaymentSource(orderID, card)
+            val cardResult = cardAPI.confirmPaymentSource(request.orderID, request.card)
             callback.onCardResult(cardResult)
         }
     }
--- a/Card/src/test/java/com/paypal/android/card/CardClientUnitTest.kt
+++ b/Card/src/test/java/com/paypal/android/card/CardClientUnitTest.kt
@@ -43,8 +43,10 @@ class CardClientUnitTest {
 
     @Test
     fun `approve order confirms payment source using card api`() = runBlockingTest {
+        val request = CardRequest(orderID, card)
+
         lateinit var capturedResult: CardResult
-        sut.approveOrder(orderID, card) { result ->
+        sut.approveOrder(request) { result ->
             capturedResult = result
         }
         assertSame(cardResult, capturedResult)
--- a/Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt
+++ b/Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt
@@ -78,7 +78,7 @@ class CardViewModel @Inject constructor(
 
         viewModelScope.launch {
             val order = fetchOrder()
-            cardClient.approveOrder(order.id!!, card) { result ->
+            cardClient.approveOrder() { result ->
                 when (result) {
                     is CardResult.Success -> {
                         Log.d(TAG, "SUCCESS")
--- a/Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt
+++ b/Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt
@@ -14,6 +14,7 @@ import com.paypal.android.api.model.PurchaseUnit
 import com.paypal.android.api.services.PayPalDemoApi
 import com.paypal.android.card.Card
 import com.paypal.android.card.CardClient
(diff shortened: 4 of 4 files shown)


Card/src/main/java/com/paypal/android/card/CardRequest.kt
```diff
@@ -0,0 +1,3 @@
+package com.paypal.android.card
+
+data class CardRequest(val orderID: String, val card: Card)
```
sarahkoop, line 3: Do we anticipate more fields being added to this request object? Should these parameters be in the constructor or set via `request.orderID = some-order-id` etc?

sshropshire: Potentially yeah. I think since these are technically required in the current implementation, they are set via constructor to enforce `@NonNull`.

I'm also not opposed to requiring empty constructors for request objects. In Kotlin, the `apply` method works well when building a request. A bit more maintenance checking for `null` params internally, but it could be worth it for API consistency.

tdchow: I think there's some safety in utilizing `val`s here. It'll ensure that the field is only set once in a merchant's app. As more fields are added to the request, we can utilize default parameters to not break current integrations.

tdchow (approved):

Card/src/main/java/com/paypal/android/card/CardRequest.kt
sarahkoop: Does using default params create a poor experience for Java merchants? Something like `CardRequest card = new CardRequest(orderID, card, null, null);` where the default params are null for Java? If so I think we should reconsider an empty constructor and build with setters/apply.

sshropshire: Yeah that's a good callout. I do like this pattern in Kotlin as well:

```kotlin
val cardRequest = CardRequest().apply {
  orderId = "sample-order-id"
  card = Card("4111 1111 1111 1111", "01", "23", "123")
}
```

Also to Tim's point, we do gain some immutability using `val` in the constructor, but we do need to be mindful of the Kotlin / Java interop. There may be a way for us to explicitly define the constructor for `Card` in Kotlin to prevent having to pass `null` for the default parameters.

tdchow: We can utilize the `@JvmOverloads` annotation here to generate the overloading functions for Java.

```kotlin
data class CardRequest @JvmOverloads constructor(
    val orderId: String,
    val optionalField: String? = null
)
```
Java Example
```java
CardRequest cardRequest = new CardRequest("orderId");
```

One thing that we'd need to think about is the ordering of the properties. New optional fields will need to be added to the end of the constructor to prevent breaking existing integrations.

I'm also good with converting them to `var`s and moving them out of the constructor if we feel that the setter syntax is the way we want to go.

sarahkoop: Hmmm.. Ok I see the argument for immutability, and if the null parameters aren't required in Java integrations that could be ok. One other thing to consider is the docs story for this. I know Android Studio provides helper labels if you have a bunch of parameters in an object so that you can see what they refer to, but a developer doc for Java integration with multiple different String parameters wouldn't be as explicit:
`CardRequest cardRequest = new CardRequest("orderId", "123456", "some-other-id-type");`

Also, which approach would be preferred if we had 30 optional fields (ex: the Braintree Android [ThreeDSecureAdditionalInformation](https://github.com/braintree/braintree_android/blob/master/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureAdditionalInformation.java) class)? I think whatever pattern we use for the request objects should probably be used across all merchant-facing objects.

tdchow: > a developer doc for Java integration with multiple different String parameters wouldn't be as explicit

This is a great callout. It still can be an issue in Kotlin when named parameters are not used. Along with developer docs, it can be really confusing when it comes to PR reviews. 

Throwing out an idea, maybe we could use the builder pattern alongside Kotlin's data class to get the best of both worlds.

```kotlin
data class CardRequest @JvmOverloads constructor(
    val orderId: String,
    val optionalField: String? = null
) {
    class Builder {
        private var orderId: String = ""
        private var optionalField: String? = null

        fun orderId(orderId: String) = apply { this.orderId = orderId }
        fun optionalField(optionalField: String) = apply { this.optionalField = optionalField }

        fun build() = CardRequest(orderId, optionalField)
    }
}
```

```java
CardRequest cardRequest = new CardRequest.Builder()
        .orderId("orderId")
        .optionalField("optionalField")
        .build();
```

A downside to this approach is that it's possible for Kotlin integrations to use the builder, or for Java integrations to use the data class's constructor. It might be less confusing to have a single integration point. 🤔

sshropshire: > A downside to this approach is that it's possible for Kotlin integrations to use the builder, or for Java integrations to use the data class's constructor. It might be less confusing to have a single integration point.

Yeah I agree. I do like going through the exercise of evaluating each option, but it does feel like a single integration point would be best for merchants.

I'd vote for required params in the constructor and we can continue to re-evaluate as we build more of the SDK to see if this pattern scales well for other payment methods.

sarahkoop: Yeah ok since both params in this PR are required I think it can go as-is and we can re-evaluate for optionals in the future

sarahkoop (approved):
Noticed the matching iOS PR has `CardOrder` instead of `CardRequest`. I prefer request but don't feel strongly. Either way, this PR looks good assuming we get those names aligned.

scannillo:
Agreed with Sarah we should align on naming across iOS and Android. I prefer the `CardRequest` and `CardResult` naming convention.

Card/src/main/java/com/paypal/android/card/CardRequest.kt
tdchow: Sounds good to me 👍

sshropshire:
@sarahkoop @scannillo agreed yeah I reverted back to `CardRequest` in that PR. iOS and Android are now in alignment.

PR #31, opened 2021-12-14
Created Simple PayPal Button for checkout payments

andrespelaezp:

### Summary of changes

 - Created PayPalButton and PayPalCreditButton styles for payment

 ### Checklist

 - [ ] Added a changelog entry

-


Demo/src/main/java/com/paypal/android/ui/payment/paymentbutton/PayPalButton.kt
```diff
@@ -0,0 +1,233 @@
...
+ *      android:id="@+id/paymentButton"
+ *      android:layout_width="match_parent"
+ *      android:layout_height="wrap_content" />
+ * ```
+ *
+ * Optionally you can provide the following attributes: `paypal_color`, `paypal_label`,
+ * `payment_button_shape`, and `payment_button_size`.
+ *
+ * Within your Activity, Fragment, or View that contains the [PayPalButton] instance you can invoke
+ * the [setup] function:
```
jcnoriega: this is copied from NXO right? Please remember to update this docs to match what we are going to implement on Northstar

andrespelaezp: It is indeed, thank you for the reminder 😄

Demo/src/main/java/com/paypal/android/ui/payment/paymentbutton/PaymentButton.kt
```diff
@@ -0,0 +1,462 @@
...
+     * Updates the color of the Payment Button with the provided [PaymentButtonColor].
+     */
+    abstract var color: C
+
+    /**
+     * Provides the current [PaymentButtonEligibilityStatus]. This value can change after the button
+     * has been rendered, those changes can be observed on by using either
+     * [onEligibilityStatusChanged] or [paymentButtonEligibilityStatusChanged].
+     */
+    private var eligibilityStatus: PaymentButtonEligibilityStatus = PaymentButtonEligibilityStatus.Loading
```
tdchow: I don't think we'll want to keep any of the eligibility logic in the payment buttons. Would you be able to remove everything eligibility related?

tdchow (commented):
Looking good! I think we'll need to doublecheck where the payment button classes are located. It looks like they're currently in the `Demo` module. I believe they should belong in the `PayPal` module so that they'll be packaged in the SDK.

Demo/src/main/java/com/paypal/android/ui/payment/paymentbutton/PaymentButton.kt
andrespelaezp: you are completely right, changes are done now 👍

Demo/src/main/java/com/paypal/android/ui/PaymentMethodsFragment.kt
```diff
@@ -78,6 +84,10 @@ class PaymentMethodsFragment : Fragment() {
         navigate(PaymentMethodsFragmentDirections.actionPaymentMethodsFragmentToCardFragment())
     }
 
+    private fun launchPayPalButtonFragment() {
```
sarahkoop: Do we need a separate fragment for the buttons? Could we replace the current demo app PayPal flow button with these new buttons instead?

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/PayPalButton.kt
```diff
@@ -0,0 +1,232 @@
...
+                SILVER.value -> SILVER
+                else -> throw createFormattedIllegalArgumentException("PayPalButtonColor", values().size)
+            }
+        }
+    }
+}
+
+/**
+ * Defines the labels available for payment buttons. If no label is provided then it will
+ * default to [PAYPAL] which will default to no label and only displaying the PayPal wordmark. For
```
sarahkoop:
```suggestion
 * default to [PAYPAL] which will not display a label and will only display the PayPal wordmark. For
```

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/PayPalButton.kt
```diff
@@ -0,0 +1,232 @@
...
+ * other labels they will have the label value itself along with a position either at the start or
+ * the end of the button.
+ *
+ * @see PAYPAL this is the default label which does not display a label at all.
+ * PayPalButton.
```
sarahkoop:
```suggestion
```
Not sure if we need that note

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/PayPalButton.kt
```diff
@@ -0,0 +1,232 @@
...
+ */
+enum class PayPalButtonLabel(
+    val value: Int,
+    val position: Position? = null,
+    private val stringResId: Int? = null
+) {
+    PAYPAL(value = 0),
+    CHECKOUT(
```
sarahkoop, line 186: Should each of these labels have a doc string telling merchants when the particular label should be used?

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/PaymentButton.kt
```diff
@@ -0,0 +1,281 @@
...
+    private val colorLuminance: PaymentButtonColorLuminance
+        get() = color.luminance
+
+    /**
+     * Updates the color of the Payment Button with the provided [PaymentButtonColor].
+     */
+    abstract var color: C
+
+    /**
+     * Provides the current value of the prefix text which is displayed before the button's wordmark.
```
sarahkoop: Nitpick - can we remove "Proivdes the current" on each of theses docs and just have "The prefix text..." etc for these strings? The extra text seems unnecessary.

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/PaymentButton.kt
```diff
@@ -0,0 +1,281 @@
...
+            background = field
+        }
+
+    private val colorLuminance: PaymentButtonColorLuminance
+        get() = color.luminance
+
+    /**
+     * Updates the color of the Payment Button with the provided [PaymentButtonColor].
+     */
+    abstract var color: C
```
sshropshire: qq: Are generics required here? There is an added complexity with generic types that will result in an increased maintenance cost. I'm wondering if there's a simpler solution that we haven't considered.

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/PayPalCreditButton.kt
```diff
@@ -0,0 +1,101 @@
...
+ *      android:layout_width="match_parent"
+ *      android:layout_height="wrap_content" />
+ * ```
+ *
+ * Optionally you can provide the following attributes: `paypal_credit_color`,
+ * `payment_button_shape`, and `payment_button_size`.
+ *
+ */
+@RequiresApi(Build.VERSION_CODES.M)
+class PayPalCreditButton @JvmOverloads constructor(
```
sarahkoop, line 29: Should we have some tests for these button classes? Not sure exactly what approach we want to take for testing UI elements, but we should probably have some coverage on the default cases and functionality.

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/PaymentButton.kt
```diff
@@ -0,0 +1,281 @@
...
+import com.google.android.material.shape.ShapeAppearanceModel
+import com.paypal.android.checkout.R
+
+@Suppress("TooManyFunctions")
+@RequiresApi(Build.VERSION_CODES.M)
+abstract class PaymentButton<C : PaymentButtonColor> @JvmOverloads constructor(
+    context: Context,
+    attributeSet: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : LinearLayout(context, attributeSet, defStyleAttr) {
```
sshropshire, line 30: accessibility q: Does the screen reader see each PaymentButton as a button?

Demo/src/main/java/com/paypal/android/ui/PaymentMethodsFragment.kt
andrespelaezp: Since PaymentMethodsFragment uses Compose to create the view, I would need to construct my view as Composable too. I already spent some time doing this without success but I agree with the fact of combining these two, so I will figure something out.

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/PaymentButton.kt
andrespelaezp: Sounds good to me

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/PaymentButton.kt
andrespelaezp: I left them for now since I wasn't sure about a future enhancement (customization) to the buttons, but I agree, I don't think is needed

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/PaymentButton.kt
andrespelaezp: Not as a Button but as a Layout, but they are indeed seen independently

Demo/src/main/java/com/paypal/android/ui/PaymentMethodsFragment.kt
tdchow: There's actually interop between Compose and Android views. You should be able to keep the button classes as is and add them to the existing Compose view. See https://developer.android.com/jetpack/compose/interop/interop-apis#views-in-compose

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/PaymentButton.kt
sshropshire: We could potentially use an [ImageSpan](https://developer.android.com/reference/android/text/style/ImageSpan) and set the button's text to a spannable string. Then all buttons can extend `Button` and will be accessible to screen readers.

Demo/src/main/java/com/paypal/android/ui/PaymentMethodsFragment.kt
andrespelaezp: @tdchow nice! thanks for the heads up, worked like a charm

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/PaymentButton.kt
andrespelaezp: Thanks for the comment @sshropshire, I made sure the screen reader was able to get the inside of the layout by adding the contentDescription in each button class, so no refactor was needed (thanks to @tdchow for the advice)

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/error/Exceptions.kt
```diff
@@ -0,0 +1,7 @@
+package com.paypal.android.checkout.paymentbutton.error
+
+fun createFormattedIllegalArgumentException(enumName: String, enumValues: Int): IllegalArgumentException {
```
tdchow: I believe this function is only used within the SDK, could we make this function internal?

PayPalCheckout/src/main/java/com/paypal/android/checkout/paymentbutton/PayPalButton.kt
tdchow: I think a small kdoc outlining each of the label options will be helpful.

andrespelaezp: @sarahkoop @tdchow I added some doc, let me know if that's enough or if I should include more information

tdchow (approved):
Looks good!

ndmunoz10 (approved):
lgtm

PR #35, opened 2022-02-01
PayPal Web Integration

sarahkoop:
### Summary of changes

 - Remove Native Checkout SDK and implement a web flow for PayPal
 - Use browser switch for web flow
 - Update PayPalClient integration to use a listener pattern for results
 - Add PayPalError class and re-organize errors
 - General code/doc string cleanup throughout 

 ### Checklist

- @sarahkoop 
- @jcnoriega 
- @andrespelaezp


PayPalCheckout/src/main/java/com/paypal/android/checkout/BrowserSwitchHelper.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.json.JSONObject
+
+
+internal class BrowserSwitchHelper(private val urlScheme: String) {
+
+    private val redirectUriPayPalCheckout = String.format("%s://x-callback-url/paypal-sdk/paypal-checkout", urlScheme)
+
+    private fun buildPayPalCheckoutUri(@NonNull orderId: String?, @NonNull config: CoreConfig): Uri {
+        val baseURL = when (config.environment) {
+            Environment.LIVE -> "https://www.paypal.com"
```
sarahkoop, line 16: In iOS we pulled these URLs into an Environment extension in the PayPal module - this doesn't exist in Kotlin, do we think these are ok to live in this helper class or should they go elsewhere?

PayPalCheckout/src/main/java/com/paypal/android/checkout/PayPalCheckoutListener.java
```diff
@@ -2,6 +2,9 @@
 
 import androidx.annotation.NonNull;
 
-public interface PayPalCheckoutResultCallback {
+/**
+ * Interface for receiving results of the PayPal payment flow.
+ */
+public interface PayPalCheckoutListener {
```
sarahkoop: We have some mis-aligned naming throughout this PR. We need to get consensus on what we want to call classes in this module, and the whole module (PayPal, PayPalCheckout ...)

PayPalCheckout/src/main/java/com/paypal/android/checkout/PayPalClient.kt
```diff
@@ -1,44 +1,96 @@
...
+        configuration: CoreConfig,
+        urlScheme: String
+    ) : this(activity, configuration, BrowserSwitchClient(), BrowserSwitchHelper(urlScheme))
+
+    private var browserSwitchResult: BrowserSwitchResult? = null
+    var listener: PayPalCheckoutListener? = null
+        set(value) {
+            field = value
+            browserSwitchResult?.also {
+                handleBrowserSwitchResult()
```
sarahkoop, line 43: This handles a browser switch immediately when the listener is set - this allows merchants to set a listener after their Activity has been resumed, and the browser switch has been delivered, and still receive the result. How should should this behavior be documented?

PayPalCheckout/src/main/java/com/paypal/android/checkout/PayPalClient.kt
```diff
@@ -1,44 +1,100 @@
...
+    private var browserSwitchResult: BrowserSwitchResult? = null
+
+    var listener: PayPalCheckoutListener? = null
+        /**
+         * @param value a [PayPalCheckoutListener] to receive results from the PayPal flow
+         */
+        set(value) {
+            field = value
+            browserSwitchResult?.also {
+                handleBrowserSwitchResult()
```
sarahkoop, line 45: This handles a browser switch immediately when the listener is set - this allows merchants to set a listener after their Activity has been resumed, and the browser switch has been delivered, and still receive the result. How should should this behavior be documented?

PayPalCheckout/src/main/java/com/paypal/android/checkout/error/PayPalError.kt
```diff
@@ -0,0 +1,23 @@
...
+    )
+
+    // 1. Result did not contain the expected data.
+    val malformedResultError = CoreSDKError(
+        code = Code.MALFORMED_RESULT.ordinal,
+        errorDescription = "Result did not contain the expected data. Payer ID or Order ID is null."
+    )
+}
+
+internal enum class Code {
```
sarahkoop, line 21: These codes are not accessible to merchants - we return the error as a `CoreSDKError` with a description, but the code will just be an Int to the merchant. I think this is the same in iOS. Does this make sense? In BT we just return a description in `BraintreeException`. Do we need these codes? Moving them all to the CoreSDKError would make them visible to the merchant, but wouldn't support modularity.

PayPalCheckout/src/main/res/color/paypal_dark_blue.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
```
sarahkoop, line 1: We were missing some of the Button customization classes (they had been included from Native Checkout) so adding them now that Native Checkout was removed.

Core/src/main/java/com/paypal/android/core/CoreConfig.kt
```diff
@@ -1,5 +1,13 @@
 package com.paypal.android.core
 
+/**
+ * Configuration for SDK requests.
+ *
+ * @property clientId your client ID from the PayPal Developer Portal
+ * @property clientSecret
```
cgdibble, line 7: Does this need a description akin to `clientId` above?

cgdibble (approved):

Core/src/main/java/com/paypal/android/core/CoreConfig.kt
jcnoriega: there is a TODO in this class, that consist on deleting the `clientSecret`. We added it at the beginning as a temporary fix por authenticating api calls until `ordersV2` didnt need it anymore. I think they have fix it, so we might have to write a ticket for this.

PayPalCheckout/src/main/java/com/paypal/android/checkout/PayPalClient.kt
sshropshire: I'm not sure but we should be able to fully encapsulate this and all they'll have to do is set a listener right after instantiation. We can re-visit this once it's implemented and we can use BT as a guide.

PayPalCheckout/src/main/java/com/paypal/android/checkout/PayPalLifeCycleObserver.kt
```diff
@@ -0,0 +1,12 @@
+package com.paypal.android.checkout
+
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleObserver
+import androidx.lifecycle.OnLifecycleEvent
+
+class PayPalLifeCycleObserver(private val payPalClient: PayPalClient) : LifecycleObserver {
+    @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
+    fun resume() {
+        payPalClient.handleBrowserSwitchResult()
```
sshropshire, line 10: I haven't tested this but apparently, the lifecycle gets notified after `onCreate()` returns so if we can get merchants to call `setListener()` immediately after instantiation within onCreate we should be good.

PayPalCheckout/src/main/java/com/paypal/android/checkout/error/PayPalError.kt
sshropshire: There's also this if we need it [Library Annotation](https://developer.android.com/studio/write/annotations#libraries) 🙌

build.gradle
```diff
@@ -61,9 +61,10 @@ buildscript {
             "jsonAssert"                  : "org.skyscreamer:jsonassert:1.4.0",
             "striktMockk"                 : "io.strikt:strikt-mockk:0.30.1",
             "striktCore"                  : "io.strikt:strikt-core:0.30.1",
+            "jsonMock"                    : "org.json:json:20211205",
 
-            // PayPal
-            "paypalNativeCheckout"        : "com.paypal.checkout:android-sdk:0.5.2"
+            // Browser Switch
+            "browserSwitch"               : "com.braintreepayments.api:browser-switch:2.1.1"
```
sshropshire: nit: can we move this up above the test dependencies?

sshropshire (approved):

nyteshade:
Please do not merge these until I have spoken to Aki and David

PayPalCheckout/src/main/java/com/paypal/android/checkout/PayPalLifeCycleObserver.kt
sarahkoop: But do we want to have that requirement on them? Would it be better to allow them to set the listener whenever?

build.gradle
sarahkoop: done 👍

PR #39, opened 2022-02-03
Align Card API with iOS

sshropshire:
### Summary of changes

 - Replace `CardClient#approveOrder` callback pattern with a coroutine style suspend function. 

 ### Checklist

### DIFF
--- a/Card/src/main/java/com/paypal/android/card/CardResult.kt
+++ b/Card/src/main/java/com/paypal/android/card/CardResult.kt
@@ -1,17 +1,5 @@
 package com.paypal.android.card
 
 import com.paypal.android.core.OrderStatus
-import com.paypal.android.core.PayPalSDKError
 
-sealed class CardResult {
-
-    class Success(
-        val orderID: String,
-        val status: OrderStatus
-    ) : CardResult()
-
-    class Error(
-        val payPalSDKError: PayPalSDKError,
-        val correlationID: String?
-    ) : CardResult()
-}
+data class CardResult(val orderID: String, val status: OrderStatus)
--- a/Card/src/main/java/com/paypal/android/card/CardResult.kt
+++ b/Card/src/main/java/com/paypal/android/card/CardResult.kt
@@ -2,4 +2,7 @@ package com.paypal.android.card
 
 import com.paypal.android.core.OrderStatus
 
+/**
+ * A result returned by [CardClient] when an order was successfully approved with a [Card].
+ */
 data class CardResult(val orderID: String, val status: OrderStatus)
--- a/Card/src/test/java/com/paypal/android/card/CardClientUnitTest.kt
+++ b/Card/src/test/java/com/paypal/android/card/CardClientUnitTest.kt
@@ -21,7 +21,7 @@ class CardClientUnitTest {
     private val orderID = "sample-order-id"
 
     private val cardAPI = mockk<CardAPI>(relaxed = true)
-    private val cardResult = CardResult.Success("orderId", OrderStatus.APPROVED)
+    private val cardResult = CardResult("orderId", OrderStatus.APPROVED)
 
     private val testCoroutineDispatcher = TestCoroutineDispatcher()
 
--- a/Card/src/test/java/com/paypal/android/card/CardClientUnitTest.kt
+++ b/Card/src/test/java/com/paypal/android/card/CardClientUnitTest.kt
@@ -29,7 +29,7 @@ class CardClientUnitTest {
 
     @Before
     fun beforeEach() {
-        sut = CardClient(cardAPI, testCoroutineDispatcher)
+        sut = CardClient(cardAPI)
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -44,15 +43,16 @@ card.number = "4111111111111111"
...
-cardClient.checkoutWithCard(ORDER_ID, card) { result, error _>
-    error?.let {
-        // handle checkout error
-        return
-    }
-    result?.let {
-        val orderID = it.orderID 
-        // Send orderID to your server to process the payment
-    }
+lifecycleScope.launch {
```
sarahkoop, line 46: Just for a sanity check - what does invoking a suspend function that throws an error look like in java?

jcnoriega: i think that for Java we might want to wrap the suspend function and use a callback. A merchant that uses Java would call `approveOrder` with a callback and we jsut wrap the suspend functions inside of it.
From what i read, callings Kotlin suspend functions from Java is not that trivial

sarahkoop: Should that be a part of this PR? Thinking about the next task of doing the Java demo app - will that be possible with the code as is?

sshropshire: We do have a separate spike planned for providing callback extensions. We may end up with a Kotlin client that can be invoked from Java.

Card/src/main/java/com/paypal/android/card/CardResult.kt
```diff
@@ -1,17 +1,5 @@
...
-        val orderID: String,
-        val status: OrderStatus
-    ) : CardResult()
-
-    class Error(
-        val payPalSDKError: PayPalSDKError,
-        val correlationID: String?
-    ) : CardResult()
-}
+data class CardResult(val orderID: String, val status: OrderStatus)
```
sarahkoop, line 8: Let's add a doc string to this

Card/src/test/java/com/paypal/android/card/CardClientUnitTest.kt
```diff
@@ -21,15 +21,15 @@ class CardClientUnitTest {
     private val orderID = "sample-order-id"
 
     private val cardAPI = mockk<CardAPI>(relaxed = true)
-    private val cardResult = CardResult.Success("orderId", OrderStatus.APPROVED)
+    private val cardResult = CardResult("orderId", OrderStatus.APPROVED)
 
     private val testCoroutineDispatcher = TestCoroutineDispatcher()
```
sarahkoop:
```suggestion
```
Don't think we need this anymore

Core/src/main/java/com/paypal/android/core/PayPalSDKError.kt
```diff
@@ -2,5 +2,6 @@ package com.paypal.android.core
 
 data class PayPalSDKError(
```
sarahkoop, line 6: Doc string here

Core/src/main/java/com/paypal/android/core/PayPalSDKError.kt
```diff
@@ -2,5 +2,6 @@ package com.paypal.android.core
...
     val code: Int?,
-    val errorDescription: String?
+    val errorDescription: String?,
```
sarahkoop: Should this be optional? I'd think we always need a description

Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt
```diff
@@ -80,17 +80,14 @@ class CardViewModel @Inject constructor(
...
-                        Log.d(TAG, "${result.status}")
-                    }
-                    is CardResult.Error -> {
-                        Log.e(TAG, "ERROR")
-                        Log.e(TAG, result.payPalSDKError.errorDescription.orEmpty())
-                    }
-                }
+
+            try {
+                val result = cardClient.approveOrder(request)
```
sarahkoop, line 85: Much cleaner 🔥 🔥

sarahkoop (commented):
Can we update the `docs/Card/integration-guide.md` with these changes too?

Card/src/test/java/com/paypal/android/card/CardClientUnitTest.kt
sshropshire: Yeah we can remove since we're mocking this co-routine. This can sometimes be useful but I forget why.

Core/src/main/java/com/paypal/android/core/PayPalSDKError.kt
sshropshire: You're right I think both `code` and `errorDescription` can be marked non-optional. I'll update.

sshropshire:
> Can we update the `docs/Card/integration-guide.md` with these changes too?

Done!

sarahkoop (approved):
Looks good to me 🌟

jcnoriega (approved):

andrespelaezp (approved):

PR #41, opened 2022-02-08
Align PayPal API with iOS

sshropshire:
### Summary of changes

 - Remove `PayPalCheckoutResultCallback`
 - Add `PayPalListener`
 - Make `PayPalCheckoutResult` a data class
 - Remove `PayPalCheckout.Failure` class
 - Remove `PayPalCheckout.Cancelled` class
 - Move `PayPalClient` from `ViewModel` to `Fragment` for now

 ### Checklist


jcnoriega:
the listener implementation looks like what we did on web fallback, so i think it looks good.

Demo/src/main/java/com/paypal/android/ui/paypal/PayPalFragment.kt
```diff
@@ -55,22 +72,8 @@ class PayPalFragment : Fragment() {
...
-            }
-        }
-
-        view.findViewById<View>(R.id.payPalCreditButton).setOnClickListener {
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
-                launchNativeCheckout()
-            }
-        }
+        paypalClient = PayPalClient(application, coreConfig, returnUrl)
+        paypalClient.listener = this
```
jcnoriega, line 75: im not that convinced by this. I think that we should remove as much logic as posible from `Fragments/Activities`. I think the `PayPalClient` should live on the `viewmodel` layer rather than the UI one. But it falls more on the Demo app side rather than the sdk

sshropshire: Made this change because `PayPalViewModel` was starting to feel less like a ViewModel and more like a Controller. We got caught in this trap in the Braintree demo app too.

sshropshire: I'm open to reverting to a ViewModel centric style, but my concern is the logic in ViewModels will become more complex in the Demo app, which could make it harder to maintain.

sarahkoop: My two cents would be we should keep anything that requires Context in the Fragment/Activity. I don't think the View Model should be aware of Android Context.

PayPalCheckout/src/main/java/com/paypal/android/checkout/PayPalListener.kt
```diff
@@ -0,0 +1,9 @@
+package com.paypal.android.checkout
+
+import com.paypal.android.core.PayPalSDKError
+
+interface PayPalListener {
```
sarahkoop, line 8: doc strings

sarahkoop (approved):
This looks good to me. I think we are still missing some doc strings and integration guide updates, but we can hold off on those until we've updated the web PR.

jcnoriega (approved):

PR #43, opened 2022-02-17
DTNOR-390: Complete release script

jcnoriega:
Ticket: https://engineering.paypalcorp.com/jira/browse/DTNOR-390

### Summary of changes
This PR adds the complete release script for Android. The script takes a version to release, and then executes the following steps:
- Build
- Unit Test each module (in parallel)
- Update versionName and versionCode in build.gradle
- Publish every module (in parallel)

- jcnoriega

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,10 @@
 name: Publish Release
-on: [workflow_dispatch]
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to release'
+        required: true
 env:
   SIGNING_KEY_FILE_PATH: /home/runner/secretKey.gpg
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -104,6 +135,15 @@ jobs:
         with:
           signing_key_file: ${{ secrets.SIGNING_KEY_FILE }}
           signing_file_path: ${{ env.SIGNING_KEY_FILE_PATH }}
+      - name: Set github user
+        # We set these credentials because we have no way of getting the users (that started the workflow) email
+        # Reference: https://github.com/actions/checkout/issues/13
+        run: |
+          git config user.name github-actions[bot]
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+      - name: Pull updated files
```
sarahkoop, line 136: Do we need this pull step for any of these tasks? Unclear what we'd be pulling here, I thought this action runs off of whatever branch it's run from.

.github/workflows/release.yml
```diff
@@ -86,9 +91,35 @@ jobs:
...
+    name: Change Version
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - name: Set github user
+          # We set these credentials because we have no way of getting the users (that started the workflow) email
+        # Reference: https://github.com/actions/checkout/issues/13
+        run: |
+          git config user.name github-actions[bot]
```
sarahkoop: Should we maybe put this as blocked until we can get an actual account for this? Also, I think we still need to update the Sonatype user in the secrets here to a group account. I don't think we'd want to publish using this script until both of those are complete.

jcnoriega: @sarahkoop the sonatype account is updated! its the user `braintree` in 1Password under Sonatype for Android publishing. Or do you mean creating a new one that its not `braintree`? perhaps a paypal one?

sarahkoop: Ah I missed that ok great!

.github/workflows/release.yml
jcnoriega: @sarahkoop the workflow modifies the `build.gradle` to update the version name and code, then commits it and creates a tag. But the publishing is done on a diferent `job`. Each `job` runs on the state in which the branch is when the workflow starts (if a change a file in a `job`, that change is not translated to other `jobs`). That the reason why it pulls, to have the new `versionCode` and `versionName`

jcnoriega: `Each job runs in a fresh instance of the virtual environment specified by runs-on.` this phrase perhaps explains it better than me lol

andrespelaezp (approved):

.github/workflows/release.yml
sarahkoop: That makes sense - thanks for explaining!

sarahkoop (approved):

PR #44, opened 2022-02-23
Integration with Java

jcnoriega:
### Summary of changes

A callback has been added to `Card.approveOrder` in order to facilitate integrations with merchants using Java. The callback pattern is the desired choice to use across the app

 ### Checklist

 - [ ] Added a changelog entry

- jonoriega
- sshropshire


sarahkoop:
This looks like a good approach to me! Do we have any concerns about it?

sarahkoop:
This looks like a good approach to me! Do we have any concerns about it?

sshropshire:
@sarahkoop there is a bit of a choice paradox if we keep them in the same module. Auto complete in Java still shows the coroutine method. There is a chance that some merchants may try to invoke the coroutine version and that'll send them down a rabbit hole.

sarahkoop:
It seems like the Jetpack libraries are Java support first and then have `-ktx` extension libraries. We could go that route so that only merchants who include the `ktx` library get the coroutine functionality.  But that would mean our core SDK is using the more legacy features.  This could be the best future-proof path though.

I think we could have auto-complete include both methods and rely on merchants reading the documentation before using a method. I suppose we might get more inbounds, but hopefully they would be quick to redirect to the docs and close.

Card/build.gradle
```diff
@@ -55,6 +55,9 @@ dependencies {
 
     androidTestImplementation deps.androidxJUnit
     androidTestImplementation deps.androidxEspressoCore
+
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:1.3.9'
```
sarahkoop: Nitpick but let's put this with the rest of the `deps`

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -18,4 +22,26 @@ class CardClient internal constructor(private val cardAPI: CardAPI) {
      */
     suspend fun approveOrder(request: CardRequest): CardResult =
```
sarahkoop, line 23: Should we call out in this method docs to only use with Kotlin integration?

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -18,4 +22,26 @@ class CardClient internal constructor(private val cardAPI: CardAPI) {
...
+                val result = approveOrder(request)
+                callback.success(result)
+            } catch (e: PayPalSDKError) {
+                callback.failure(e)
+            }
+        }
+    }
+}
+
+interface ApproveOrderCallback {
```
sarahkoop: Maybe this should live in its own file and have doc strings? Not sure what convention we want to follow for callbacks in this project.

sarahkoop (approved):
A few comments, but looks good!

andrespelaezp (approved):
Looking good!

PR #46, opened 2022-03-08
Magnes integration

jcnoriega:
### Summary of changes

 -  This PR integrates Magnes to northstar SDK, wrapping it inside PayPalDataCollector

 ### Checklist

 - [ ] Added a changelog entry

-  @jcnoriega 
-  @sshropshire


Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt
```diff
@@ -77,10 +78,13 @@ class CardViewModel @Inject constructor(
         val card = Card(number, monthString, yearString)
         card.securityCode = _securityCode.value ?: ""
 
+        System.setProperty("magnes.debug.mode", true.toString()) // check this
```
sarahkoop: Should this line or the comment be removed?

PayPalDataCollector/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
sarahkoop, line 1: I think this and the consumer-rules can probably be deleted.

PayPalDataCollector/src/androidTest/java/com/paypal/android/paypaldatacollector/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import org.junit.Assert.*
+
+/**
+ * Instrumented test, which will execute on an Android device.
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+@RunWith(AndroidJUnit4::class)
+class ExampleInstrumentedTest {
```
sarahkoop: Should we remove this or update it to PayPalDataCollectorTest with a TODO to write integration tests?

PayPalDataCollector/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,15 @@
...
+
+    <!-- following permission is optional -->
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
+    <uses-permission android:name="android.permission.GET_ACCOUNTS" />
+    <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.READ_PHONE_STATE" />
+    <!-- for reading GSF ID -->
```
sarahkoop: Are all these required? Looks like on BT we only use `INTERNET` and `ACCESS_NETWORK_STATE`

PayPalDataCollector/src/main/java/com/paypal/android/paypaldatacollector/PayPalDataCollector.kt
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    constructor(coreConfig: CoreConfig) : this(
+        coreConfig,
+        MagnesSDK.getInstance(),
+        UUIDHelper()
+    )
+
+    /**
+     * Gets a Client Metadata ID at the time of payment activity. Once a user initiates a PayPal payment
```
sarahkoop: This will be used for other payments than just `PayPal` in the future right? Maybe we should make this more generic. Also maybe when referring to `PayPal` as the processor we should say the SDK or the server depending? Seems a little confusing

PayPalDataCollector/src/main/java/com/paypal/android/paypaldatacollector/PayPalDataCollector.kt
```diff
@@ -0,0 +1,73 @@
...
+     * from their device, PayPal uses the Client Metadata ID to verify that the payment is
+     * originating from a valid, user-consented device and application. This helps reduce fraud and
+     * decrease declines. This method MUST be called prior to initiating a pre-consented payment (a
+     * "future payment") from a mobile device. Pass the result to your server, to include in the
+     * payment request sent to PayPal. Do not otherwise cache or store this value.
+     *
+     * @param context Android Context
+     * @param clientMetadataId The desired pairing ID, trimmed to 32 characters.
```
sarahkoop: Is pairing ID a known term? Not sure what that means

PayPalDataCollector/src/main/java/com/paypal/android/paypaldatacollector/PayPalDataCollector.kt
```diff
@@ -0,0 +1,73 @@
...
+                .setAppGuid(uuidHelper.getInstallationGUID(context))
+            magnesSDK.setUp(magnesSettingsBuilder.build())
+            val result = magnesSDK.collectAndSubmit(
+                appContext,
+                clientMetadataId,
+                additionalData
+            )
+            result.paypalClientMetaDataId
+        } catch (e: InvalidInputException) {
+            // Either clientMetadataId or appGuid exceeds their character limit
```
sarahkoop, line 64: Should this error be surfaced somehow or do we want to fail silently on data collection?

jcnoriega: Braintree makes it fail silently. Perhaps the rationale behind that is that data collection is not a super important part of the merchant expierence? We could just make the method throw the exception and thats it.

Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt
jcnoriega: this is the only way to make it log. Perhaps we could make `PayPalDataCollector` have a method like `setLogging(boolean)`. The only thing is that in the logs they will see the `Magnes` tag, do we want the merchants to know that we use magnes?

PayPalDataCollector/src/androidTest/java/com/paypal/android/paypaldatacollector/ExampleInstrumentedTest.kt
jcnoriega: this is autogenerated when i created the module, i'll delete it

PayPalDataCollector/src/main/AndroidManifest.xml
jcnoriega: i'll check, this is all on the magnes docs, but you maybe right about not needing all

PayPalDataCollector/src/main/java/com/paypal/android/paypaldatacollector/PayPalDataCollector.kt
sarahkoop: Ah you're right - failing silently is probably the way to go 👍

PayPalDataCollector/src/main/AndroidManifest.xml
sshropshire: Yeah if we don't need access to fine grained location then that'll keep it from popping up on the install sheet in Google Play for all of our merchants.

sshropshire: I agree with @sarahkoop the less permissions the better.

PayPalDataCollector/src/main/java/com/paypal/android/paypaldatacollector/PayPalDataCollector.kt
jcnoriega: im honestly not sure, doesnt the whole SDK fall under PayPal? From the merchant side, im guessing that Northstar is PayPal

PayPalDataCollector/proguard-rules.pro
jcnoriega: @sarahkoop just noticed that all modules have this and are referenced on their gradle files as well, should i delete those as well?

sarahkoop: I guess we should keep them if we are planning to add specific ProGuard rules. If they are empty not sure we need them 🤷

Demo/src/main/java/com/paypal/android/ui/card/CardViewModel.kt
sarahkoop: I guess I was mostly wondering about the //check this comment. Are we telling ourselves or the merchant to check this?

jcnoriega: yeah my bad, that was more for us. I added the function `setLogging`, to give merchants a chance to see if data collection fails.

PayPalDataCollector/proguard-rules.pro
jcnoriega: right now `minifyEnabled = false`. So i thinks this prompts up a more broader discussion if we want to ofuscate code eventually. Mainly to reduce the weight of the sdk, because the code is public anyway

sshropshire (approved):

sarahkoop (approved):

PR #49, opened 2022-03-16
Complete workflow for release and snapshots

jcnoriega:
### Summary of changes

 - Complete release process. The release script is executed manually and it will build, test, publish and bump version with tag. After that, it will create the next version with -SNAPSHOT.
 The second script will execute after the first one finishes, to publish the snapshot version. It will also publish snapshots every time the main branch is changed

 ### Checklist

 - [ ] Added a changelog entry

- @jcnoriega
- @sshropshire


.github/workflows/release.yml
```diff
@@ -131,11 +109,9 @@ jobs:
         with:
           signing_key_file: ${{ secrets.SIGNING_KEY_FILE }}
           signing_file_path: ${{ env.SIGNING_KEY_FILE_PATH }}
-      - name: Set github user
```
sarahkoop, line 134: Remind me why this is being removed? We are creating a shared GitHub user for releases right?

.github/workflows/release.yml
```diff
@@ -224,17 +196,44 @@ jobs:
...
-          signing_key_file: ${{ env.SIGNING_KEY_FILE_PATH }}
\ No newline at end of file
+          signing_key_file: ${{ env.SIGNING_KEY_FILE_PATH }}
+
+  releasing_finished:
+    needs: [ publish_checkout, publish_card, publish_core, publish_data_collector ]
+    name: All Unit Test finished
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "Unit test finished"
```
sarahkoop:
```suggestion
      - run: echo "Release finished"
```

.github/workflows/release.yml
```diff
@@ -224,17 +196,44 @@ jobs:
...
           sonatype_pwd: ${{ secrets.SONATYPE_NEXUS_PASSWORD }}
           signing_key_id: ${{ secrets.SIGNING_KEY_ID }}
           signing_key_pwd: ${{ secrets.SIGNING_KEY_PASSWORD }}
-          signing_key_file: ${{ env.SIGNING_KEY_FILE_PATH }}
\ No newline at end of file
+          signing_key_file: ${{ env.SIGNING_KEY_FILE_PATH }}
+
+  releasing_finished:
+    needs: [ publish_checkout, publish_card, publish_core, publish_data_collector ]
+    name: All Unit Test finished
```
sarahkoop:
```suggestion
    name: Releases Finished
```

.github/workflows/release.yml
jcnoriega: We do have a shared account.
This part of the code is being moved forward in the script, at the end to push the bump version commit.

.github/workflows/release.yml
```diff
@@ -224,17 +196,44 @@ jobs:
...
+    name: Bump Version
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - name: Set github user
+        uses: ./.github/actions/set_github_user
+      - name: Update Version
+        run: |
+          ./gradlew -PversionParam=${{ github.event.inputs.version }} changeReleaseVersion
```
sarahkoop, line 231: I think this should run before the releases right? We want the updated version number to be in the gradle files when released?

jcnoriega: @sarahkoop well here is where it gets kinda tricky:
Every `job` that publishes a module, updates the gradle file with the new version and releases. As each job runs on a diferent environment, after it completes, the changes are lost. So, once all the publishing is done, we do the "oficial" change of version in the `bump_version job` that also commits its and tags it. We decided to move the tagging and commiting to the end of the workflow, so if the publishing fails, we dont bump the version on the repo. (before it was the other way around, we commited, and then we published)

sarahkoop: Ok that makes sense

sarahkoop (approved):

.github/workflows/release.yml
```diff
@@ -224,17 +196,44 @@ jobs:
...
           signing_file_path: ${{ env.SIGNING_KEY_FILE_PATH }}
-      - name: Set github user
-        uses: ./.github/actions/set_github_user
-      - name: Pull updated files
+      - name: Change Version
         run: |
-          git pull
-      - name: Publish PayPal DataCollector
+          ./gradlew -PversionParam=${{ github.event.inputs.version }} changeReleaseVersion
+      - name: PPublish PayPal DataCollector
```
sshropshire:
```suggestion
      - name: Publish PayPal DataCollector
```

.github/workflows/release.yml
```diff
@@ -224,17 +196,44 @@ jobs:
...
+          signing_key_file: ${{ env.SIGNING_KEY_FILE_PATH }}
+
+  releasing_finished:
+    needs: [ publish_checkout, publish_card, publish_core, publish_data_collector ]
+    name: Releases Finished
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "Release finished"
+
+  #Once all releases are done, we bump version, tag it and prepare next
```
sshropshire:
```suggestion
  # Once all releases are done, we bump version, tag it and prepare next
```

sshropshire (approved):

PR #54, opened 2022-03-29
Remove Core dependency from PayPalDataCollector

jcnoriega:
### Summary of changes

 - This PR removes the Core dependency from PayPalDataCollector. We expose a PayPalDataCollectorEnvironment that transforms to Magnes Environment.

 ### Checklist

 - [ ] Added a changelog entry

- @jcnoriega

### DIFF
--- a/PayPalDataCollector/src/main/java/com/paypal/android/paypaldatacollector/PayPalDataCollectorEnvironment.kt
+++ b/PayPalDataCollector/src/main/java/com/paypal/android/paypaldatacollector/PayPalDataCollectorEnvironment.kt
@@ -1,0 +1,13 @@
+package com.paypal.android.paypaldatacollector
+
+enum class PayPalDataCollectorEnvironment {
+    LIVE,
+    SANDBOX,
+    STAGING
+}
+
+internal fun getMagnesEnvironment(environment: PayPalDataCollectorEnvironment) = when (environment) {
+    PayPalDataCollectorEnvironment.LIVE -> lib.android.paypal.com.magnessdk.Environment.LIVE
+    PayPalDataCollectorEnvironment.STAGING -> lib.android.paypal.com.magnessdk.Environment.STAGE
+    PayPalDataCollectorEnvironment.SANDBOX -> lib.android.paypal.com.magnessdk.Environment.SANDBOX
+}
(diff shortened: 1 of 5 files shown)


PayPalDataCollector/src/main/java/com/paypal/android/paypaldatacollector/PayPalDataCollectorEnvironment.kt
```diff
@@ -0,0 +1,13 @@
+package com.paypal.android.paypaldatacollector
+
+enum class PayPalDataCollectorEnvironment {
```
sarahkoop, line 6: doc strings for this class

sshropshire: Actually would we expect merchants to use this? Might be a good time to introduce those [@RestrictTo](https://developer.android.com/studio/write/annotations#libraries) annotations.

sarahkoop (approved):

PayPalDataCollector/src/main/java/com/paypal/android/paypaldatacollector/PayPalDataCollectorEnvironment.kt
jcnoriega: @sshropshire we have to expose the environment somehow. Exposing the `Magnes` `Enviornment` class directly seems to me that defeats the purpose of this module, that essentially wraps `Magnes`.
How would you use the `@RestrictTo` in this case?

sshropshire: If we can reduce the amount of documentation we have to maintain I think it's a win. `RestrictTo` calls out to merchants via the linter that this is library code, and shouldn't be used by the merchant. It's still technically accessible, but with a warning.

sshropshire: This may be a broader conversation though. We would need to confirm that NXO uses the same `groupId` that we do for our artifacts. Since Kotlin doesn't have a `package-private` feature like Java, I think it's in our best interest to draw the line between library code and merchant facing API. We'll have to do the same thing on iOS.

sshropshire (approved):

PayPalDataCollector/src/main/java/com/paypal/android/paypaldatacollector/PayPalDataCollectorEnvironment.kt
jcnoriega: @sshropshire ok gotcha. But it seems to me that this doesnt apply here as this class is part of the merchant facing API.

PR #55, opened 2022-04-05
Adds PayPalCredit and PayLater funding options

jcnoriega:
### Summary of changes

 -  This PR adds the `PayPalWebCheckoutFunding` to the web checkout flow. That value can be either `CREDIT` or `PAY_LATER`

- @jcnoriega

### DIFF
--- a/PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFunding.kt
+++ b/PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFunding.kt
@@ -1,0 +1,5 @@
+package com.paypal.android.checkoutweb
+
+enum class PayPalWebCheckoutFunding(val value: String) {
+    CREDIT("credit"), PAY_LATER("paylater"), DEFAULT("")
+}
--- a/PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFunding.kt
+++ b/PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFunding.kt
@@ -1,8 +1,0 @@
-package com.paypal.android.checkoutweb
-
-/**
- * Enum class to specify the type of funding for an order
- */
-enum class PayPalWebCheckoutFunding(val value: String) {
-    CREDIT("credit"), PAY_LATER("paylater"), DEFAULT("")
-}
--- a/PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFunding.kt
+++ b/PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFunding.kt
@@ -1,5 +1,8 @@
 package com.paypal.android.checkoutweb
 
+/**
+ * Enum class to specify the type of funding for an order
+ */
 enum class PayPalWebCheckoutFunding(val value: String) {
     CREDIT("credit"), PAY_LATER("paylater"), DEFAULT("")
 }
--- a/PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFundingSource.kt
+++ b/PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFundingSource.kt
@@ -2,15 +2,18 @@ package com.paypal.android.checkoutweb
 
 /**
  * Enum class to specify the type of funding for an order.
+ * For more information go to: https://developer.paypal.com/docs/checkout/pay-later/us/
  */
 enum class PayPalWebCheckoutFundingSource(val value: String) {
     /**
      * PAYPAL_CREDIT will launch the web checkout flow and display PayPal Credit funding to eligible customers
+     * Eligible costumers receive a revolving line of credit that they can use to pay over time.
      */
     PAYPAL_CREDIT("credit"),
 
     /**
-     * PAY_LATER will launch the web checkout flow and display Pay Later offers to eligible customers.
+     * PAY_LATER will launch the web checkout flow and display Pay Later offers to eligible customers,
+     * which include short-term, interest-free payments and other special financing options.
      */
     PAY_LATER("paylater"),
 
--- a/PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFundingSource.kt
+++ b/PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFundingSource.kt
@@ -1,0 +1,8 @@
+package com.paypal.android.checkoutweb
+
+/**
+ * Enum class to specify the type of funding for an order
+ */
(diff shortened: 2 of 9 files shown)


PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFunding.kt
```diff
@@ -0,0 +1,8 @@
+package com.paypal.android.checkoutweb
+
+/**
+ * Enum class to specify the type of funding for an order
+ */
+enum class PayPalWebCheckoutFunding(val value: String) {
```
sarahkoop: Wondering about naming on this - would `PayPalWebCheckoutFundingSource` or `PayPalWebCheckoutFundingType` be more clear for this, or are those options too long?

PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFunding.kt
```diff
@@ -0,0 +1,8 @@
...
+    CREDIT("credit"), PAY_LATER("paylater"), DEFAULT("")
```
sarahkoop: Let's add doc strings to each of these types to explain what each funding source means, and what the default is.

PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFunding.kt
jcnoriega: it could be `FundingSource`. As im not a native english speaker, does the concept of "order's funding" make sense? My point being that if adding `Source` or `Type` adds more info to the merchant.

Also, i think we need to maybe re visit the naming convention of adding the module name as a prefix, because class name start to get really long :(

jaxdesmarais: I think `PayPalWebCheckoutFundingSource` makes the most sense here and lines up with what we call a funding source in other SDKs (like NXO). The naming is quite long but I think we are going to want to lean towards more verbose naming to avoid naming collisions with NXO.

PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutRequest.kt
```diff
@@ -4,5 +4,6 @@ package com.paypal.android.checkoutweb
  * Creates an instance of a PayPalRequest.
  *
  * @param orderID The ID of the order to be approved.
+ * @param funding specify funding (credit, paylater or default)
  */
-data class PayPalWebCheckoutRequest(val orderID: String)
+data class PayPalWebCheckoutRequest @JvmOverloads constructor (val orderID: String, val funding: PayPalWebCheckoutFundingSource = PayPalWebCheckoutFundingSource.PAYPAL)
```
sarahkoop:
```suggestion
data class PayPalWebCheckoutRequest @JvmOverloads constructor (val orderID: String, val fundingSource: PayPalWebCheckoutFundingSource = PayPalWebCheckoutFundingSource.PAYPAL)
```

PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFundingSource.kt
```diff
@@ -4,5 +4,18 @@ package com.paypal.android.checkoutweb
...
  */
 enum class PayPalWebCheckoutFundingSource(val value: String) {
-    CREDIT("credit"), PAY_LATER("paylater"), PAYPAL("paypal")
+    /**
+     * CREDIT will launch the web checkout flow with credit funding selected
+     */
+    CREDIT("credit"),
+
+    /**
+     * PAY_LATER will launch the web checkout flow with pay later selected
```
sarahkoop:
```suggestion
     * PAY_LATER will launch the web checkout flow and display Pay Later offers to eligible customers after they have logged into PayPal.
```
I think these still need a little more clarity. I took this from Braintree docs, but is this accurate about what happens in the flow if they select this funding source?

PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFundingSource.kt
```diff
@@ -4,5 +4,18 @@ package com.paypal.android.checkoutweb
...
+     */
+    PAY_LATER("paylater"),
+
+    /**
+     * PAYPAL will launch the web checkout default flow
```
sarahkoop:
```suggestion
     * PAYPAL will launch the web checkout for a one-time PayPal Checkout flow.
```

Is this accurate? Do we only offer one time checkout through the default or is vaulting possible here too?

PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFundingSource.kt
```diff
@@ -4,5 +4,18 @@ package com.paypal.android.checkoutweb
  * Enum class to specify the type of funding for an order
  */
 enum class PayPalWebCheckoutFundingSource(val value: String) {
-    CREDIT("credit"), PAY_LATER("paylater"), PAYPAL("paypal")
+    /**
+     * CREDIT will launch the web checkout flow with credit funding selected
```
sarahkoop: I'm not sure what the wording on this one should be - is this also only for eligible customers? Is there any PayPal documentation on this?

PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFundingSource.kt
jcnoriega: Yes, and its kinda the same for `credit`.

PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFundingSource.kt
jcnoriega: I would word it similar to `PAY_LATER`, is it just displays the credit options. But i tried to get some docs but couldnt find any

PayPalWebCheckout/src/main/java/com/paypal/android/checkoutweb/PayPalWebCheckoutFundingSource.kt
jcnoriega: Im not sure. Passing no value for the `fundingSource` query param, lanuches the same flow as passing `paypal`, so im guessing its just the default. I do have to log in every time.

sarahkoop (approved):
Looks good to me! Is there a card to update the integration guide with this new param or should that be in this PR?

jcnoriega:
@sarahkoop i'll do it in this PR, thanks for mentioning it!

andrespelaezp (approved):
Looks great!

PR #56, opened 2022-04-06
[DO NOT MERGE] Spike 3DS

sshropshire:
### Summary of changes

 - Spike to initiate a 3DS Verification Flow
 - This requires an `application_context` when creating an Order via _ppcp_merchant_server_
 - For now, we're forwarding these parameters to the merchant server
 - In the future, we should find a way to set `return_url` and `cancel_url` when calling `confirm-payment-source` to avoid requiring a merchant to provide a server-side integration in order to perform 3DS Verification

 ### Checklist


Card/src/main/java/com/paypal/android/card/CardAPIRequestFactory.kt
```diff
@@ -35,4 +35,43 @@ internal class CardAPIRequestFactory {
         val path = "v2/checkout/orders/$orderID/confirm-payment-source"
         return APIRequest(path, HttpMethod.POST, body)
     }
+
+    fun createAuthorizeWith3DSVerificationRequest(orderID: String, card: Card): APIRequest {
+        val cardNumber = card.number.replace("\\s".toRegex(), "")
```
crookedneighbor, line 40: Should this be changed to remove any non-number? And should this logic for cleaning up the card number be moved to a function that can be used anywhere it needs to be run? (I assume non-3ds card flows will also need to do it)

crookedneighbor:
Would love to see what the API looks like for the merchant. Do we have any mock up docs for what a merchant would do to integrate with 3ds?

Card/src/main/java/com/paypal/android/card/CardAPIRequestFactory.kt
sshropshire: We may be missing a spec for this. For now all it does is strip out spaces, but we should have local validation in place to provide instant feedback.

sshropshire:
> Would love to see what the API looks like for the merchant. Do we have any mock up docs for what a merchant would do to integrate with 3ds?

Not at the moment. Currently, 3DS requires `return_url` and `cancel_url` to be forwarded to the create order POST, which would require us to provide a server side example. The assumption is that we'll be able to provide these values to `confirm-payment-source` in the near future and it'll just work.

sshropshire:
Closing in favor of #62.

PR #58, opened 2022-04-26
[DONT MERGE] ThreeDSecure Verify

sshropshire:
### Summary of changes

 - This PR as 3DS as optional parameter to the Card module. We also added Client side integration, but we will keep it hidden until phase 2.
 - Unit tests will be added in a separate PR.

 ### Checklist

 - [ ] Added a changelog entry

-@jcnoriega
-@sshropshire


Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -1,42 +1,155 @@
...
     /**
-     * Confirm [Card] payment source for an order. Use this method for Kotlin integrations
+     *  CardClient constructor
      *
-     * @param request [CardRequest] for requesting an order approval
+     *  @param [activity] Activity that launches the card client
+     *  @param [configuration] Configuration parameters for client
+     *  @param [coroutineContext] CoroutineContext to execute
      */
-    suspend fun approveOrder(request: CardRequest): CardResult =
```
sshropshire, line 23: qq: We're removing `suspend` methods?

Card/src/main/java/com/paypal/android/card/CardLifeCycleObserver.kt
```diff
@@ -0,0 +1,19 @@
...
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.LifecycleOwner
+
+internal class CardLifeCycleObserver(private val cardClient: CardClient) : DefaultLifecycleObserver {
+
+    private var isFirstEventHandled = false
+
+    override fun onResume(owner: LifecycleOwner) {
+        super.onResume(owner)
+        // We want to ignore any previous onResume events before this class was instantiated
```
sshropshire, line 13: Was there an actual bug with multiple `onResume` events? Browser switch should only deliver once.

Card/src/main/java/com/paypal/android/card/CreateOrderRequestFactory.kt
```diff
@@ -0,0 +1,65 @@
...
+                            ,
+                            "application_context": {
+                                "return_url": "${it.returnUrl}",
+                                "cancel_url": "${it.cancelUrl}"
+                            }
+                        """
+            } ?: ""
+        }
+            }
+        """.trimIndent()
```
sshropshire, line 51: 🧶 : Can we fix the indentation here?

Card/src/main/java/com/paypal/android/card/CardLifeCycleObserver.kt
jcnoriega: Its an issue with `lifecycle` in general. Like when I add an `Observer` to a lifecycle, it will immediately emmit the latest event, which was `onResume`. So i was getting `onResume` twice.

Card/src/main/java/com/paypal/android/card/CardClient.kt
jcnoriega: I thought it was the only way, because 3DS does a browser switch, so we cant just return a value.

jcnoriega: @sshropshire i did add a `CoroutineContext` to the client, i think we should give developers the option to choose in which thread our client is going to do its request. Right now its in the `constructor`, but maybe it should be a param for each function

sshropshire: > I thought it was the only way, because 3DS does a browser switch, so we cant just return a value.

Ok yeah that makes sense. Then we should remove the callback parameter too and replace it with a listener.

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -1,42 +1,155 @@
...
+
+    /**
+     * Creates order and confirms [Card] payment source for said order. Its private until phase 2
+     *
+     * @param orderRequest [String] order to create
+     * @param cardRequest [CardRequest] for requesting an order approval
+     * @param threeDSecureRequest [threeDSecureRequest] to run transaction with 3DS
+     * @param callback [ApproveOrderCallback] callback for responses
+     */
+    private fun createAndApproveOrder(
```
sshropshire, line 86: This `private` method is no longer used. We should delete it.

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -1,42 +1,155 @@
...
+        threeDSecureRequest: ThreeDSecureRequest? = null,
+        callback: ApproveOrderCallback
+    ) {
+        CoroutineScope(coroutineContext).launch {
+            val orderId = createOrder(orderRequest, threeDSecureRequest)
+            confirmPaymentSource(orderId, cardRequest, threeDSecureRequest, callback)
+        }
+    }
+
+    private suspend fun createOrder(
```
sshropshire, line 108: I think this can be removed as well. It's only used by the method above.

Card/src/main/java/com/paypal/android/card/api/CardAPI.kt
```diff
@@ -1,23 +1,52 @@
...
+        return performRequest(
+            CreateOrderRequestFactory.createRequest(orderRequest, threeDSecureRequest),
+            CreateOrderRequestFactory::parseResponse
+        )
+    }
+
+    private suspend fun <R> performRequest(
+        apiRequest: APIRequest,
+        parseResult: (body: String, correlationId: String?) -> R
+    ): R {
```
sshropshire, line 57: Do we need generics here?

Card/src/main/java/com/paypal/android/card/CardClient.kt
jcnoriega: this implementation is for phase 2, i talked to @demerino and we agreed to just hide the implementation from the release, then we just expose it on phase 2

Card/src/main/java/com/paypal/android/card/CardClient.kt
jcnoriega: same awnser es `createAndApproveOrder` being private

Card/src/main/java/com/paypal/android/card/api/CardAPI.kt
jcnoriega: pefroming a request is kinda mechanical, so the idea is that you supply an `APIRequest` to the method, and them a function that parses the result and return the correct `response`. In `confirm_payment_source` it returns a `ConfirmPaymentSourceResponse` and in create order a `CreateOrderResponse`. I think it will be helpfull when we add more api requests to the module

Card/src/main/java/com/paypal/android/card/CardClient.kt
sshropshire: Oh word that's cool. I pulled down the branch just now though we may need to update `CardFragment` in that case. There's a compilation error I can't get the app to build.

sshropshire:
Closing in favor of #62.

PR #59, opened 2022-05-02
Add PayLater button

jcnoriega:
### Summary of changes

 - PayLater button added to PayPalUI
 
Screen Shot 2022-05-02 at 12 24 16

 ### Checklist

 - [ ] Added a changelog entry

- @jcnoriega


PayPalUI/src/main/java/com/paypal/android/ui/paymentbutton/PayLaterButton.kt
```diff
@@ -0,0 +1,56 @@
...
+    attributeSet: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : PayPalButton(context, attributeSet, defStyleAttr) {
+
+    /**
+     * Provides the label of the [PayLaterButton]. This value will always be [PAY_LATER], attempting
+     * to set the value to anything else will result in that value being ignored.
+     */
+    override var label: PayPalButtonLabel = PayPalButtonLabel.PAY_LATER
+        set(value) {
```
kgangineni, line 36: Couldn't understand this part, 
letting only one value be set, I see similar logic in super class too, any reason for this approach instead of `val`

kgangineni, line 37: Couldn't understand this part, 
letting only one value be set, I see similar logic in super class too, any reason for this approach instead of `val`

PayPalUI/src/main/java/com/paypal/android/ui/paymentbutton/PayLaterButton.kt
```diff
@@ -0,0 +1,56 @@
...
+            if (value == PayPalButtonLabel.PAY_LATER) {
+                field = value
+                updateLabel(field)
+            }
+        }
+
+    override val wordmarkDarkLuminanceResId: Int
+        get() = R.drawable.logo_paypal_monochrome
```
sshropshire:
```suggestion
    override val wordmarkDarkLuminanceResId: Int = R.drawable.logo_paypal_monochrome
```
🧶 : Since these values don't change we can simplify these property declarations, unless Kotlin syntax requires this for overloads.

PayPalUI/src/main/java/com/paypal/android/ui/paymentbutton/PayLaterButton.kt
```diff
@@ -0,0 +1,56 @@
...
+ * <PayPalLater
+ *      android:id="@+id/payLaterButton"
+ *      android:layout_width="match_parent"
+ *      android:layout_height="wrap_content" />
+ * ```
+ *
+ * Optionally you can provide the following attributes: `paypal_color`, `paypal_label`,
+ * `payment_button_shape`, and `payment_button_size`.
+ */
+@RequiresApi(Build.VERSION_CODES.M)
```
sarahkoop: Curious what in here requires API 23? Should we be considering raising the min SDK of the entire SDK to 23? It feels odd to have a button that isn't available to all SDK users.

PayPalUI/src/main/java/com/paypal/android/ui/paymentbutton/PayLaterButton.kt
jcnoriega: The `PayPalButton` supports many labels, like "Checkout", "Pay Now", etc. and the label can be changed on the fly. BUT, `PayLater` is a subtype of `PayPalButton`, that only takes an inmutable labe: "Pay later", thats why overrides the super class which is a `var`, and forces it to be `Pay Later`

PayPalUI/src/main/java/com/paypal/android/ui/paymentbutton/PayLaterButton.kt
jcnoriega: i think we are taking this from NXO, but i tested it and we dont seem to need it. I've removed it

Demo/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,10 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     package="com.paypal.android">
+
+    <!-- We add this so that we can test on API 21, because NXO doesn't support below 23 (for now)  -->
+    <uses-sdk tools:overrideLibrary="com.paypal.android.checkout"/>
```
sarahkoop, line 7: Can this be removed since the whole SDK is supporting 21 now?

PayPalUI/src/main/java/com/paypal/android/ui/paymentbutton/PayLaterButton.kt
```diff
@@ -0,0 +1,52 @@
...
+
+import android.content.Context
+import android.os.Build
+import android.util.AttributeSet
+import androidx.annotation.RequiresApi
+import com.paypal.android.ui.R
+
+
+/**
+ * PayLaterButton provides a PayPal PayLater button with the ability to modify the [color], [label], [shape],
```
sarahkoop: Can you modify the label on this button or no? The docs below make it seem like this can't be changed

sarahkoop (approved):
Added a few more nitpick questions but otherwise looks good

PayPalUI/src/main/java/com/paypal/android/ui/paymentbutton/PayLaterButton.kt
jcnoriega: No you cant. The PayLater Button is exactly like a PayPalButton, but with a fix label. Fixing the doc now!

kgangineni (approved):
LGTM

PR #60, opened 2022-05-06
Creates test section for styling buttons

kgangineni:

### Summary of changes
Test section for styling buttons
testSection

 - 

 ### Checklist

 - [ ] Added a changelog entry

- @kgangineni 

-

### DIFF
--- a/Demo/src/main/java/com/paypal/android/ui/paypal/PayPalButtonsFragment.kt
+++ b/Demo/src/main/java/com/paypal/android/ui/paypal/PayPalButtonsFragment.kt
@@ -91,60 +91,40 @@ class PayPalButtonsFragment : Fragment() {
     }
 
     private fun initPaymentButtonSizeSpinner() {
-        binding.paymentButtonSizeSpinner.adapter = ArrayAdapter(
-            requireContext(),
-            android.R.layout.simple_list_item_1,
-            PaymentButtonSize.values()
-        )
+        binding.paymentButtonSizeSpinner.adapter = simpleArrayAdapter(PaymentButtonSize.values())
         binding.paymentButtonSizeSpinner.onItemSelectedListener = itemSelectionListener
     }
 
+    private fun <T> simpleArrayAdapter(items: Array<T>) = ArrayAdapter(
+        requireContext(),
+        android.R.layout.simple_list_item_1,
+        items
+    )
+
     private fun initPaymentButtonShapeSpinner() {
-        binding.paymentButtonShapeSpinner.adapter = ArrayAdapter(
-            requireContext(),
-            android.R.layout.simple_list_item_1,
-            PaymentButtonShape.values()
-        )
+        binding.paymentButtonShapeSpinner.adapter = simpleArrayAdapter(PaymentButtonShape.values())
         binding.paymentButtonShapeSpinner.onItemSelectedListener = itemSelectionListener
     }
 
     private fun initPaymentButtonLabelSpinner() {
         binding.paymentButtonLabelSpinner.adapter = when (selectedButtonFundingType) {
-            ButtonFundingType.PAYPAL -> ArrayAdapter(
-                requireContext(),
-                android.R.layout.simple_list_item_1,
-                PayPalButtonLabel.values()
-            )
-            else -> ArrayAdapter(
-                requireContext(),
-                android.R.layout.simple_list_item_1,
-                listOf("N/A")
-            )
+            ButtonFundingType.PAYPAL -> simpleArrayAdapter(PayPalButtonLabel.values())
+            else -> simpleArrayAdapter(arrayOf("N/A"))
         }
         binding.paymentButtonLabelSpinner.onItemSelectedListener = itemSelectionListener
     }
 
     private fun initFundingTypesSpinner() {
-        binding.paymentButtonFundingSpinner.adapter = ArrayAdapter(
-            requireContext(),
-            android.R.layout.simple_list_item_1,
-            ButtonFundingType.values()
-        )
+        binding.paymentButtonFundingSpinner.adapter = simpleArrayAdapter(ButtonFundingType.values())
         binding.paymentButtonFundingSpinner.onItemSelectedListener = itemSelectionListener
     }
 
     private fun initPaymentButtonColorSpinner() {
         binding.paymentButtonColorSpinner.adapter = when (selectedButtonFundingType) {
-            ButtonFundingType.PAYPAL, ButtonFundingType.PAY_LATER -> ArrayAdapter(
-                requireContext(),
-                android.R.layout.simple_list_item_1,
+            ButtonFundingType.PAYPAL, ButtonFundingType.PAY_LATER -> simpleArrayAdapter(
                 PayPalButtonColor.values()
             )
-            ButtonFundingType.PAYPAL_CREDIT -> ArrayAdapter(
-                requireContext(),
-                android.R.layout.simple_list_item_1,
-                PayPalCreditButtonColor.values()
-            )
+            ButtonFundingType.PAYPAL_CREDIT -> simpleArrayAdapter(PayPalCreditButtonColor.values())
         }
         binding.paymentButtonColorSpinner.onItemSelectedListener = itemSelectionListener
     }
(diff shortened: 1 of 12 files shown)


kgangineni:
https://github.com/paypal/Android-SDK/pull/59 this needs to be merged first

PayPalUI/src/main/java/com/paypal/android/ui/paymentbutton/PayLaterButton.kt
```diff
@@ -0,0 +1,56 @@
...
+ * <PayPalLater
+ *      android:id="@+id/payLaterButton"
+ *      android:layout_width="match_parent"
+ *      android:layout_height="wrap_content" />
+ * ```
+ *
+ * Optionally you can provide the following attributes: `paypal_color`, `paypal_label`,
+ * `payment_button_shape`, and `payment_button_size`.
+ */
+@RequiresApi(Build.VERSION_CODES.M)
```
sshropshire: qq: is there a reason these buttons require API level 23?

Demo/src/main/java/com/paypal/android/ui/paypal/PayPalButtonsFragment.kt
```diff
@@ -0,0 +1,184 @@
...
+            binding.paymentButtonShapeSpinner.selectedItem.toString()
+        )
+
+    private val selectedSize: PaymentButtonSize
+        get() = PaymentButtonSize.valueOf(
+            binding.paymentButtonSizeSpinner.selectedItem.toString()
+        )
+
+    private val binding: FragmentPaypalButtonsStylingBinding by lazy {
+        FragmentPaypalButtonsStylingBinding.inflate(layoutInflater)
```
sshropshire, line 61: 🧶  : I'm looking at [this](https://developer.android.com/topic/libraries/data-binding/generated-binding) on Google's developer site:
Screen Shot 2022-05-09 at 4 39 48 PM

We should prefer the `FragmentPaypalButtonsStylingBinding.inflate(layoutInflater, viewGroup, false)` inflate variant here to be safe.

Demo/src/main/java/com/paypal/android/ui/paypal/PayPalButtonsFragment.kt
```diff
@@ -0,0 +1,184 @@
...
+        initPaymentButtonShapeSpinner()
+        initPaymentButtonSizeSpinner()
+    }
+
+    private fun initPaymentButtonSizeSpinner() {
+        binding.paymentButtonSizeSpinner.adapter = ArrayAdapter(
+            requireContext(),
+            android.R.layout.simple_list_item_1,
+            PaymentButtonSize.values()
+        )
```
sshropshire: We could extract these `ArrayAdapter` factory methods into a single companion object to make the business logic easier to follow.

Demo/src/main/res/layout/fragment_paypal_buttons_styling.xml
```diff
@@ -0,0 +1,130 @@
...
+
+    <Spinner
+        android:id="@+id/paymentButtonFundingSpinner"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:prompt="@string/select_funding"
+        android:spinnerMode="dialog"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/labelFunding" />
```
sshropshire, line 26: 🧶  : Can we migrate to Material Design's `TextInputLayout` here to be consistent with `CardFragment`?

PayPalUI/src/main/java/com/paypal/android/ui/paymentbutton/PayLaterButton.kt
kgangineni: not sure of this,
@jcnoriega any idea

Demo/src/main/java/com/paypal/android/ui/paypal/PayPalButtonsFragment.kt
kgangineni: @sshropshire this is needed, when we're attaching to parent, in this case we're not doing, so either method will have same impact, used this one because it can be put separately in val rather than in `onCreate()`

Demo/src/main/java/com/paypal/android/ui/paypal/PayPalButtonsFragment.kt
kgangineni: @sshropshire I wouldn't want to put this in `companion object` because this is not static or util functionality, we can put in separate `val` now we're not re-using anywhere else in the code so I guess not needed

Demo/src/main/res/layout/fragment_paypal_buttons_styling.xml
kgangineni: `TextInputLayout` gives nice animation effects when user's typing, in this case spinner opens a dialog so it will not be visible anyway

kgangineni:
hi @sshropshire @jcnoriega @sarahkoop 
any idea why `CardClientUnitTest` keeps failing in CI but it passes when I run in local

PayPalUI/src/main/java/com/paypal/android/ui/paymentbutton/PayLaterButton.kt
```diff
@@ -40,13 +39,16 @@ class PayLaterButton @JvmOverloads constructor(
...
+    override val wordmarkDarkLuminanceResId: Int
+        get() = R.drawable.logo_paypal_monochrome
 
-    override val wordmarkLightLuminanceResId: Int = R.drawable.logo_paypal_color
+    override val wordmarkLightLuminanceResId: Int
+        get() = R.drawable.logo_paypal_color
 
-    override val fundingType: PaymentButtonFundingType = PaymentButtonFundingType.PAY_LATER
+    override val fundingType: PaymentButtonFundingType
+        get() = PaymentButtonFundingType.PAY_LATER
```
jcnoriega: @kgangineni i think here is a merge conflict issue. I fixed this on the preivous PR, removing the `get()` accesors and simply assign it to the variable. Can you revert this change?

PayPalUI/src/main/java/com/paypal/android/ui/paymentbutton/PayLaterButton.kt
```diff
@@ -1,14 +1,12 @@
...
 import android.content.Context
-import android.os.Build
 import android.util.AttributeSet
-import androidx.annotation.RequiresApi
 import com.paypal.android.ui.R
 
 
 /**
- * PayLaterButton provides a PayPal PayLater button with the ability to modify the [color], [shape],
+ * PayLaterButton provides a PayPal PayLater button with the ability to modify the [color], [label], [shape],
```
jcnoriega: same merge issue, remove `label` from docs because PayLaterButton doesnt accept label changing

Demo/src/main/java/com/paypal/android/ui/paypal/PayPalButtonsFragment.kt
sshropshire: Interesting. Is this a pattern y'all had on PP credit? I'm not opposed, just looking to understand. I've never seen inflation without specifying the view group and whether or not to attach to parent.

The template (without view binding) generated in Android projects uses [this method](https://developer.android.com/reference/android/view/LayoutInflater#inflate(int,%20android.view.ViewGroup,%20boolean)) and honestly just feels right since the method signatures line up. Do you feel there's a specific advantage to having it in a `val`?

Demo/src/main/res/layout/fragment_paypal_buttons_styling.xml
sshropshire: Got it. The interface is nearly identical to a Spinner ~(checking this [stack overflow post](https://stackoverflow.com/a/56152098))~. It's a weird API but you get material design and the typing will be disabled with this style.

sshropshire: Here's a [better snippet](https://github.com/paypal/Android-SDK/blob/5d536b3955dfe87df24a347a70caa2025add8eb6/Demo/src/main/res/layout/fragment_card.xml#L10) in the context of this project. It also uses an `ArrayAdapter` so probably uses a `Spinner` internally.

jcnoriega (approved):

sshropshire (approved):
:fire:

PR #62, opened 2022-05-27
ThreeDSecure Verify

sshropshire:
### Summary of changes

 - Allow Strong Consumer Authentication via `CardClient::approveOrder()`
 - Add `ThreeDSecureRequest` property to `CardRequest`
 - Perform a conditional Browser Switch when a 3DS authentication challenge is received

 ### Checklist

### Authors

- @sshropshire
- @jcnoriega


Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -45,102 +46,61 @@ class CardClient internal constructor(
...
 
-    /**
-     * Creates order and confirms [Card] payment source for said order. Its private until phase 2
-     *
-     * @param orderRequest [String] order to create
-     * @param cardRequest [CardRequest] for requesting an order approval
-     * @param threeDSecureRequest [threeDSecureRequest] to run transaction with 3DS
-     * @param coroutineContext [CoroutineContext] to specify in which context to run api calls
-     */
-    private fun createAndApproveOrder(
```
jcnoriega: This is definetly going to be in phase 2. If our "policy" is to keep the main branch as clean as possible, im going to put a PR for this just to have it ready when we need it.

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -45,102 +46,61 @@ class CardClient internal constructor(
...
-     * @param coroutineContext [CoroutineContext] to specify in which context to run api calls
      */
     @JvmOverloads
     fun approveOrder(
+        activity: FragmentActivity,
         orderId: String,
         cardRequest: CardRequest,
-        threeDSecureRequest: ThreeDSecureRequest? = null,
-        coroutineContext: CoroutineContext? = Dispatchers.Default
+        context: CoroutineContext = Dispatchers.IO
```
jcnoriega: do we want default to be `IO`? if the merchant calls this from the activity and wants to make UI changes on each callback update, its going to throw `wrong thread exception`. The merchant has to force any UI update by calling `runOnMainThread{}`

I think that the simplest integration, without specifiying context, should be on `Main`

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -156,47 +116,33 @@ class CardClient internal constructor(
...
-//                                getOrderResponse.orderId,
-//                                getOrderResponse.orderStatus,
-//                                liabilityShift?.let { ThreeDSecureResult(it) }
-//                            )
-//                        )
-                    } catch (e: PayPalSDKError) {
-                        approveOrderListener?.failure(e)
-                    }
-                    clearClient()
+            CoroutineScope(Dispatchers.IO).launch {
```
jcnoriega: the `CoroutineContext` here needs to match the one in `confirmPaymentSource`, otherwise the merchant might want to launch the first one in `Main` and we are switching to `IO`, increasing the chances of them having `wrong thread exception`

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
```diff
@@ -146,123 +127,49 @@ class CardFragment : Fragment() {
     }
 
     private fun onCardFieldSubmit() {
-        val cardNumber = binding.cardNumberInput.text.toString()
-        val expirationDate = binding.cardExpirationInput.text.toString()
-        val securityCode = binding.cardSecurityCodeInput.text.toString()
-        val isServerSideIntegration = binding.serverSideIntegrationChkbox.isChecked
-        val isThreeDSecure = binding.threedsChkbox.isChecked
+        viewLifecycleOwner.lifecycleScope.launch {
```
jcnoriega, line 166: wasnt this going to be inside the client for our first iteration? so that the merchant doesnt have to sepcify the `context` and we automatically bind it to the `FragmentActivity`

Card/src/main/java/com/paypal/android/card/CardClient.kt
sshropshire: Yeah that's the thought here. We want to decrease the scope of the PR to keep it as focused as possible.

Card/src/main/java/com/paypal/android/card/CardClient.kt
sshropshire: Yeah [Dispatchers.IO](https://developer.android.com/kotlin/coroutines/coroutines-adv#main-safety) is meant for network requests and `Dispatchers.Default` is meant for CPU intensive work.

Also, if a merchant is using the fragment lifecycle scopes from Jetpack those are already associated with `Dispatchers.Main`, so there shouldn't be an issue with updating the UI in a traditional use case.

Card/src/main/java/com/paypal/android/card/CardClient.kt
sshropshire: Oh wait I see what you mean here since we're using a listener callback we need to deliver the result on `Dispatchers.Main`?

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
sshropshire: Create order uses coroutines this is more of an implementation detail for the Demo app. The SDK does use the listener pattern, but we're using coroutines to create an order via the merchant server.

Card/src/main/java/com/paypal/android/card/ApproveOrderListener.kt
```diff
@@ -5,34 +5,33 @@ import com.paypal.android.core.PayPalSDKError
 
 /**
  * Implement this callback to receive results from [CardClient].
- * Recommended for Java Integrations
  */
 interface ApproveOrderListener {
```
kgangineni, line 10: `ApproveOrderListener` is a bit confusing,
I see it has functions for both approve & cancel, so may be just `OrderListener` or a different interface for cancellation and other things

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -24,15 +23,17 @@ import kotlin.coroutines.CoroutineContext
  * Use this client to approve an order with a [Card].
  */
 class CardClient internal constructor(
-    private val activity: FragmentActivity,
+    activity: FragmentActivity,
```
kgangineni, line 24: do we specifically need `FragmentActivity`? if not this can be just `Activity`

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -45,102 +46,61 @@ class CardClient internal constructor(
...
+            approveOrderListener?.onApproveOrderSuccess(result)
             clearClient()
         } else {
             // we launch the 3DS flow
             val options = BrowserSwitchOptions()
                 .url(Uri.parse(confirmPaymentSourceResponse.payerActionHref))
                 .returnUrlScheme("com.paypal.android.demo")
             this.confirmPaymentSourceResponse = confirmPaymentSourceResponse
-            approveOrderListener?.threeDSecureLaunched()
+            approveOrderListener?.onApproveOrderThreeDSecureWillLaunch()
```
kgangineni: `approveOrderListener` if it's null, then user will not get any result, so this being null is accepted or do we want to mark it not null

Card/src/main/java/com/paypal/android/card/CardClient.kt
jcnoriega: Right, the listener callbacks are going to be called in `IO` and that is going to produce an exception.

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -156,47 +116,33 @@ class CardClient internal constructor(
         confirmPaymentSourceResponse: ConfirmPaymentSourceResponse?
     ) {
         confirmPaymentSourceResponse?.also { response ->
```
kgangineni: `let` seems better than `also` here,
`also` is when we want to get the back the object after performing operation, here we're not assigning the result to any val

Card/src/main/java/com/paypal/android/card/CardClient.kt
kgangineni: `CoroutineScope` is used when we want to save it, and when it goes out of scope we can cancel all operations within that scope, here we're not saving scope and cancelling anywhere, so i don't see need for scope here
let me know your thoughts

Card/src/main/java/com/paypal/android/card/CardLifeCycleObserver.kt
```diff
@@ -1,21 +1,23 @@
...
 
-    //private var isFirstEventHandled = false
-
     override fun onResume(owner: LifecycleOwner) {
         super.onResume(owner)
-        // We want to ignore any previous onResume events before this class was instantiated
-        //if (isFirstEventHandled) {\
+
+        val activity = when (owner) {
+            is FragmentActivity -> owner
```
kgangineni, line 14: can this be just `Activity`

Card/src/main/java/com/paypal/android/card/api/CardAPI.kt
```diff
@@ -50,7 +38,6 @@ internal class CardAPI(
         )
     }
 
-
     private suspend fun <R> performRequest(
         apiRequest: APIRequest,
         parseResult: (body: String, correlationId: String?) -> R
```
kgangineni: don't get this part of code
we're passing a function as param and returning the result of this function as result of the whole function, why not go sequential
can you explain

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
```diff
@@ -35,14 +30,17 @@ import com.paypal.android.ui.card.validation.CardFormatter
...
-import java.util.UUID
 import javax.inject.Inject
 
 @AndroidEntryPoint
-class CardFragment : Fragment() {
+class CardFragment : Fragment(), ApproveOrderListener {
+
+    companion object {
+        const val TAG = "CardFragment"
+        const val APP_RETURN_URL = "com.paypal.android.demo://example.com/returnUrl"
```
kgangineni, line 41: `APP_RETURN_URL` name sounds like it's gonna be for whole app if that's the case lets put in different file something like `Constants` if not no change required

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
kgangineni, line 166: `lifecycleScope` by default launches in `MainThread` looking at function `createOrder()` seems like there's network operation, in that case it shouldn't be called in main thread

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
```diff
@@ -297,9 +200,39 @@ class CardFragment : Fragment() {
...
+            val authText = it.authenticationResult?.let { auth ->
+                val threeDtext = "\nLiability shift: ${auth.liabilityShift}," +
+                        "Enrollment: ${auth.threeDSecure?.enrollmentStatus}," +
+                        "Authentication: ${auth.threeDSecure?.authenticationStatus}"
+                threeDtext
+            }
+            text + authText
+        } ?: ""
+
+        val deepLink = result.deepLinkUrl?.toString() ?: ""
```
kgangineni: nit: this can be replaced with `result.deepLinkUrl?.toString().orEmpty()`

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
```diff
@@ -297,9 +200,39 @@ class CardFragment : Fragment() {
...
+        val joinedText = listOf(statusText, paymentSourceText, deepLink).joinToString("\n")
```
kgangineni: name is `joinedText` what it actually contains a `list`
would like to know what it's doing

Demo/src/main/res/layout/fragment_card.xml
```diff
@@ -4,7 +4,9 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:layout_marginHorizontal="16dp">
+    android:layout_marginStart="16dp"
```
kgangineni: nit: can `16dp` go into a `dimens.xml`?

Demo/src/main/res/layout/fragment_card.xml
```diff
@@ -90,23 +92,14 @@
...
-
     <androidx.appcompat.widget.AppCompatCheckBox
         android:id="@+id/threeds_chkbox"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:text="@string/with_three_d_secure"
         android:checked="true"
         android:layout_marginTop="8dp"
-        app:layout_constraintTop_toBottomOf="@id/server_side_integration_chkbox"
+        app:layout_constraintTop_toBottomOf="@id/card_security_code"
```
kgangineni, line 102: imho: I think ui element names can be camel cased, for reason that now we're using view binding, these are translated into kotlin variable names

Card/src/main/java/com/paypal/android/card/ApproveOrderListener.kt
sshropshire: We do have a `<METHOD_NAME>Listener` naming convention in place here. The listener is mapped to the `CardClient::approveOrder()` method in this case.

Card/src/main/java/com/paypal/android/card/CardClient.kt
sshropshire: We got an inbound a while back from the GH community requesting [FragmentActivity](https://github.com/braintree/braintree_android/issues/290) to be used throughout the SDK. We're following the same convention here.

Card/src/main/java/com/paypal/android/card/CardClient.kt
sshropshire: Yeah I agree. We can probably make `ConfirmPaymentSourceResponse` non-null here.

Card/src/main/java/com/paypal/android/card/CardClient.kt
sshropshire: The primary reason we have scope is to execute the coroutine. We may want to keep a job around for merchants using the SDK in Java. Is there another alternative?

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
sshropshire: Joined text is a `String` all we're doing here is adding new lines in between each string.

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -156,47 +123,33 @@ class CardClient internal constructor(
...
+                    val getOrderResponse =
+                        cardAPI.getOrderInfo(GetOrderRequest(response.orderId))
+                    approveOrderListener?.onApproveOrderSuccess(
+                        CardResult(
+                            getOrderResponse.orderId,
+                            getOrderResponse.orderStatus,
+                            confirmPaymentSourceResponse.paymentSource,
+                            browserSwitchResult.deepLinkUrl
+                        )
+                    )
```
jcnoriega:
```suggestion
    withContext(Dispatchers.Main) {
                    approveOrderListener?.onApproveOrderSuccess(
                        CardResult(
                            getOrderResponse.orderId,
                            getOrderResponse.orderStatus,
                            confirmPaymentSourceResponse.paymentSource,
                            browserSwitchResult.deepLinkUrl
                        )
                    )
        }
```

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -156,47 +123,33 @@ class CardClient internal constructor(
...
+                } catch (e: PayPalSDKError) {
+                    approveOrderListener?.onApproveOrderFailure(e)
```
jcnoriega, line 110: same
```suggestion
    withContext(Dispatchers.Main) {
                    approveOrderListener?.onApproveOrderFailure(e)
    }
```

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -156,47 +123,33 @@ class CardClient internal constructor(
...
                 }
+                clearClient()
             }
         }
     }
 
     private fun deliverCancellation() {
-        approveOrderListener?.cancelled()
+        approveOrderListener?.onApproveOrderCanceled()
```
jcnoriega, line 117: same
```suggestion
    withContext(Dispatchers.Main) {
        approveOrderListener?.onApproveOrderCanceled()
    }        
```

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -45,102 +48,66 @@ class CardClient internal constructor(
...
             browserSwitchClient.start(activity, options)
         }
     }
 
-    internal fun handleBrowserSwitchResult() {
+    internal fun handleBrowserSwitchResult(activity: FragmentActivity) {
         val browserSwitchResult = browserSwitchClient.deliverResult(activity)
         if (browserSwitchResult != null && approveOrderListener != null && confirmPaymentSourceResponse != null) {
-            approveOrderListener?.threeDSecureFinished()
+            approveOrderListener?.onApproveOrderThreeDSecureDidFinish()
```
jcnoriega, line 88:
```suggestion
    withContext(Dispatchers.Main) {
            approveOrderListener?.onApproveOrderThreeDSecureDidFinish()
    }
```

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
sshropshire: I agree, let's remove the `APP_` prefix to make a better name. I'd like to keep this value in the same file to keep it close to the call site.

Card/src/main/java/com/paypal/android/card/CardClient.kt
kgangineni: that's interesting, how about repos that doesn't use Fragments?
they will not be having `FragmentActivity`

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
sshropshire: Yeah this is a good call. We definitely need to make sure we're doing this on `IO`.

Demo/src/main/res/layout/fragment_card.xml
sshropshire: I think `snake_case` is more common for android layout ids like `R.id.snake_case` and view binding will convert to camel case automatically. Here's a [big discussion on SO](https://stackoverflow.com/questions/1832290/android-id-naming-convention-lower-case-with-underscore-vs-camel-case) on the topic.

The only negative with camel case is we'd have to update the entire app to fit this convention.

Card/src/main/java/com/paypal/android/card/CardClient.kt
sshropshire: `AppCompatActivity` is the Google recommended, and it is a subclass of `FragmentActivity`. We could support `Activity` I'm not opposed, but in the past we've used `FragmentActivity`.

I'd honestly prefer if a merchant came and requested that `Activity` be allowed as a parameter to avoid having to support integrations that are using `Activity` since Google has shifted its focus toward the `androidx` variant as well.

Card/src/main/java/com/paypal/android/card/CardClient.kt
```diff
@@ -1,42 +1,119 @@
...
+                confirmPaymentSource(activity, cardRequest)
             } catch (e: PayPalSDKError) {
-                callback.failure(e)
+                approveOrderListener?.onApproveOrderFailure(e)
+            }
+        }
+    }
+
+    private suspend fun confirmPaymentSource(activity: FragmentActivity, cardRequest: CardRequest) {
+        val response = cardAPI.confirmPaymentSource(cardRequest)
```
jcnoriega, line 64: so are we going to use `Dispatchers.Main` for all our api requests? is it safe to asume that the api request are going to be called in`IO`? perhaps wrapping our requests in `withContext(Dispatchers.IO)`? or was this messing a lot with the unit tests?

sshropshire: Yeah after doing the tests it became clear that `Dispatchers.Main` was the correct route all along. We should be good now since the actual networking does happen on `Dispatchers.IO` in the `Http` class.

Card/src/main/java/com/paypal/android/card/CardLifeCycleObserver.kt
```diff
@@ -0,0 +1,23 @@
...
+    override fun onResume(owner: LifecycleOwner) {
+        super.onResume(owner)
+
+        val activity = when (owner) {
+            is FragmentActivity -> owner
+            is Fragment -> owner.requireActivity()
+            else -> null
+        }
+
+        Log.e("LIFECYCLE OBSERVER", "ON RESUME")
```
sarahkoop:
```suggestion
```
Not sure if we need the logs, but probably shouldn't be logging that as an error

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
```diff
@@ -43,30 +51,83 @@ class CardFragment : Fragment() {
...
+                val joinedText = listOf(statusText, paymentSourceText, deepLink).joinToString("\n")
+                updateStatusText(joinedText)
+            }
+
+            override fun onApproveOrderFailure(error: PayPalSDKError) {
+                updateStatusText("CAPTURE fail: ${error.errorDescription}")
+            }
+
+            override fun onApproveOrderCanceled() {
+                updateStatusText("USER CANCELLED")
```
sarahkoop:
```suggestion
                updateStatusText("USER CANCELED")
```

sarahkoop (approved):

Card/src/main/java/com/paypal/android/card/CardClient.kt
sshropshire: update: I tried changing to `Activity` and it broke compilation. Followed up through the docs and it turns out `activity.lifecycle` is a property on an Activity subclass [ComponentActivity](https://developer.android.com/reference/androidx/activity/ComponentActivity).

andrespelaezp (approved):

kgangineni (approved):

PR #64, opened 2022-06-08
Paypal Test Cards

sshropshire:
### Summary of changes

 - This PR refactors the Card UI to support additional test cards from [developer.paypal.com](https://developer.paypal.com/docs/checkout/advanced/customize/3d-secure/test/) 

 ### Checklist


Demo/src/main/java/com/paypal/android/ui/testcards/TestCardsViewModel.kt
```diff
@@ -0,0 +1,148 @@
+package com.paypal.android.ui.testcards
+
+import androidx.lifecycle.ViewModel
+import com.paypal.android.card.Card
+import com.paypal.android.data.card.TestCard
+import com.paypal.android.data.card.TestCardGroup
+
+class TestCardsViewModel : ViewModel() {
```
jcnoriega, line 8: @sshropshire  I think the structure of the PR is good, but all these cards are not working, so now that I found one that works, we can update this list with cards that work. Should we merge this and update the list in a different PR?

sshropshire: Yeah i'm down for whatever. The main part of this PR was the UI. The good thing is we can update this file as we get more stable test cards and the UI will update automatically.

jcnoriega (approved):

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
```diff
@@ -168,6 +164,10 @@ class CardFragment : Fragment() {
         }
     }
 
+    private fun showTestCards() {
+        findNavController().navigate(R.id.action_cardFragment_to_testCardFragment)
```
kgangineni, line 168: nit: since we've enabled viewBinding, directions can be referred from `CardFragmentDirections` instead of using `r.id`

Demo/src/main/java/com/paypal/android/ui/testcards/TestCardsFragment.kt
```diff
@@ -0,0 +1,65 @@
...
+    ): View {
+        binding = FragmentTestCardsBinding.inflate(inflater, null, false)
+        return binding.root
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        val items = mutableListOf<TestCardsItem>()
+        for (group in viewModel.testCardGroups) {
```
kgangineni, line 43: nit: this logic can be in viewModel

kgangineni (approved):
LGTM

PR #67, opened 2022-07-08
Remove Client Secret and support for Full Scoped Access Token

jcnoriega:

### Summary of changes

 This PR removes the client's `secret` from the app and SDK. It also changes `CoreConfig` significantlly:
- `clientId` is only needed in `PayPalNativeCheckout` and `EligiblityAPI`, so if its not needed across all modules, then it doesnt belong in `CoreConfig`. Same goes for `accessToken`.

Also added an `AuthHandler` to manage authorization header.

- @jcnoriega 
- @sshropshire


crookedneighbor (commented):
There are no hardcoded client secrets in this repo, correct? Want to make sure we don't need to rotate any keys.

jcnoriega:
@crookedneighbor no, there arent!

Core/src/main/java/com/paypal/android/core/AuthHandler.kt
```diff
@@ -0,0 +1,18 @@
+package com.paypal.android.core
+
+class AuthHandler private constructor(private val authHeaderValue: String) {
+
+    fun getAuthHeader() = authHeaderValue
+
+    companion object {
+
+        fun fromClientId(clientId: String): AuthHandler {
```
sshropshire: Do we need this? We should only need ~a client token~ an FSAT now.

jcnoriega: true

Demo/build.gradle
```diff
@@ -21,7 +21,6 @@ android {
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
 
         buildConfigField("String", "CLIENT_ID", paypalProperties["CLIENT_ID"] ?: "\"\"")
```
sshropshire, line 23: We can remove this too now.

Demo/src/main/java/com/paypal/android/DemoActivity.kt
```diff
@@ -31,9 +31,7 @@ class DemoActivity : AppCompatActivity() {
 
         PreferenceManager.setDefaultValues(this, R.xml.preferences, false)
 
-        val authCredentialsMissing =
-            BuildConfig.CLIENT_ID.isEmpty() || BuildConfig.CLIENT_SECRET.isEmpty()
-        if (authCredentialsMissing) {
+        if (BuildConfig.CLIENT_ID.isEmpty()) {
```
sshropshire: +1. Since client id is no longer required.

Demo/src/main/java/com/paypal/android/api/services/AuthApi.kt
```diff
@@ -29,14 +29,20 @@ interface AuthApi {
...
          *
          * OkHttp provides a convenient "Basic" function which handles the heavy lifting for us.
          */
-        val payPalAuthorization: String = Credentials.basic(BuildConfig.CLIENT_ID, BuildConfig.CLIENT_SECRET)
+        val payPalAuthorization: String
 
         var authToken: AuthToken? = null
+
+        init {
+            val basicAuth = "${BuildConfig.CLIENT_ID}:"
```
sshropshire: +1. This entire class could be removed.

Demo/src/main/java/com/paypal/android/di/DemoModule.kt
```diff
@@ -15,11 +15,10 @@ object DemoModule {
     @Provides
     fun provideEligibilityAPI(
         coreConfig: CoreConfig
-    ): EligibilityAPI = EligibilityAPI(coreConfig)
+    ): EligibilityAPI = EligibilityAPI(coreConfig, com.paypal.android.BuildConfig.CLIENT_ID)
```
sshropshire: +1

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
```diff
@@ -169,11 +125,54 @@ class CardFragment : Fragment() {
     }
 
     private suspend fun createOrder() {
+
+        val accessToken = payPalDemoApi.fetchAccessToken().value
+        val configuration = CoreConfig()
```
sshropshire: We could make `accessToken` a `CoreConfig` parameter to make authentication available to every SDK module via the core configuration.

PayPalNativeCheckout/src/main/java/com/paypal/android/checkout/PayPalClient.kt
```diff
@@ -23,7 +23,7 @@ class PayPalClient(application: Application, coreConfig: CoreConfig, returnUrl:
     init {
         val config = CheckoutConfig(
             application = application,
-            clientId = coreConfig.clientId,
+            clientId = clientId,
```
sshropshire, line 34: +1 client ID can be removed here.

Demo/build.gradle
jcnoriega: @sshropshire the `clientId` is needed for the ElgibilityAPI. I think its a good practice to still have a `.properties` file with only the `CLIENT_ID`.

Demo/src/main/java/com/paypal/android/di/DemoModule.kt
jcnoriega: i dont know what you mean. The `clientId` is requiered for `EligibilityAPI`

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
jcnoriega: the only problem I see is that right now is only needed for `Card`. If we leave it on `CoreConfig`,the merchant might get confused and try to always put a token even though its not necessary

PayPalNativeCheckout/src/main/java/com/paypal/android/checkout/PayPalClient.kt
jcnoriega: `clientId` is requiered for PayPalNativeCheckout

Demo/build.gradle
sshropshire: Ok yeah I see the docs now. If only GraphQL could accept a client token instead of a `clientId` 🤔 .

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
sshropshire: The token will most likely be necessary for `v1`. A design goal would be to use the `accessToken` as the primary authentication mechanism, but it may not be that simple it sounds like GraphQL is a bit different than some of the PayPal API.

PayPalNativeCheckout/src/main/java/com/paypal/android/checkout/PayPalClient.kt
sshropshire: Got it yeah.

Demo/src/main/java/com/paypal/android/api/services/PayPalDemoApi.kt
```diff
@@ -1,23 +1,20 @@
...
+    @POST("/orders")
+    suspend fun fetchOrderId(@Body orderRequest: CreateOrderRequest?): Order
 
-    @POST("/order")
-    suspend fun fetchOrderId(
-        @Query("countryCode") countryCode: String,
-        @Body jsonObject: JsonObject
-    ): Order
+    @POST("/orders")
+    suspend fun fetchOrderId(@Body jsonObject: JsonObject): Order
```
kgangineni: `createOrder` for method name instead of `fetchOrderId`? because we have different access and endpoints for get and creating orders
same for above method

Core/src/main/java/com/paypal/android/core/CoreConfig.kt
```diff
@@ -1,7 +1,5 @@
 package com.paypal.android.core
 
 data class CoreConfig @JvmOverloads constructor(
-    val clientId: String,
```
kgangineni, line 4: feel we can leave `clientId` here because it's a config thing would like to hear from others

andrespelaezp (approved):
LGTM

kgangineni (approved):

sshropshire (approved):
🌊

PR #74, opened 2022-07-26
Remove Client ID

sshropshire:
### Summary of changes

 - Fetch Access Token from Sample Merchant Server
 - Remove Client ID from `CoreConfig`
 - Use Access Token to fetch Client ID internally

 ### Checklist

### Authors

- @andrespelaezp
- @sshropshire


PayPalNativeCheckout/src/main/java/com/paypal/android/checkout/PayPalClient.kt
```diff
@@ -14,36 +14,35 @@ import com.paypal.checkout.error.OnError
...
-                0,
-                "Client Id should not be null or empty"
-            )
-        }
+    /**
+     * Initiate a PayPal checkout for an order.
+     *
+     * @param orderId the id of the order
+     */
+    suspend fun checkout(orderId: String) {
```
jcnoriega, line 37: wondering if we should just use a `CoroutineScope` inisde the `checkout` function. It does feel "weird" to use a suspend functions for what essentially is a UI flow from the merchants point of view. Thinks also about `java` integrations...

sshropshire: Yeah that was honestly the path of least resistance. We were already calling checkout in a coroutine so it felt ok. We can also circle back once we start integrating NXO again.

Core/src/main/java/com/paypal/android/core/API.kt
```diff
@@ -14,4 +14,17 @@ class API internal constructor(
             httpRequestFactory.createHttpRequestFromAPIRequest(apiRequest, configuration)
         return http.send(httpRequest)
     }
+
+    suspend fun getClientId(): String {
+        val apiRequest = APIRequest(
+            path = "v1/oauth2/token",
+            method = HttpMethod.GET,
+            body = null
```
kgangineni: nit: `APIRequest` can have `body=null` since get requests will not have

Core/src/main/java/com/paypal/android/core/API.kt
```diff
@@ -14,4 +14,17 @@ class API internal constructor(
...
+        )
+        val httpRequest =
+            httpRequestFactory.createHttpRequestFromAPIRequest(apiRequest, configuration)
+        val response = http.send(httpRequest)
+        val json = PaymentsJSON(response.body!!)
```
kgangineni: throw error when body is null?
` if (bodyResponse.isNullOrBlank()) {
                throw APIClientError.noResponseData(correlationID)
            }`

Core/src/main/java/com/paypal/android/core/API.kt
sshropshire: 👍 that's a good call I'll update.

PayPalNativeCheckout/src/main/java/com/paypal/android/checkout/PayPalClient.kt
jcnoriega: sure, its out of the scope of this PR, will get back to it on the nxo pr

Core/src/main/java/com/paypal/android/core/API.kt
sshropshire: Also a good call! I added in some error handling logic with tests.

kgangineni (approved):
👍

jcnoriega (approved):

PR #82, opened 2022-08-10
Add Return URL to Confirm Order Call

sshropshire:
### Summary of changes

 - `confirm-payment-source` endpoint now supports `return_url` and `cancel_url` parameters
 - This PR removes 3DS from order creation and forwards those values from within the SDK

 ### Checklist

- @sshropshire

### DIFF
--- a/Card/src/main/java/com/paypal/android/card/CardRequestFactory.kt
+++ b/Card/src/main/java/com/paypal/android/card/CardRequestFactory.kt
@@ -30,18 +30,25 @@ internal class CardRequestFactory {
                 cardJSON.put("billing_address", billingAddressJSON)
             }
 
+            val bodyJSON = JSONObject()
             threeDSecureRequest?.also {
                 val verificationJSON = JSONObject()
                     .put("method", it.sca.name)
                 val attributesJSON = JSONObject()
                     .put("verification", verificationJSON)
                 cardJSON.put("attributes", attributesJSON)
+
                 // add return and cancel url when its supported
+                val applicationContextJSON = JSONObject()
+                    .put("return_url", it.returnUrl)
+                    .put("cancel_url", it.cancelUrl)
+                bodyJSON.put("application_context", applicationContextJSON)
             }
 
             val paymentSourceJSON = JSONObject().put("card", cardJSON)
-            val bodyJSON = JSONObject().put("payment_source", paymentSourceJSON)
-            val body = bodyJSON.toString()
+            bodyJSON.put("payment_source", paymentSourceJSON)
+
+            val body = bodyJSON.toString().replace("\\/", "/")
 
             val path = "v2/checkout/orders/$orderID/confirm-payment-source"
             APIRequest(path, HttpMethod.POST, body)
--- a/Card/src/main/java/com/paypal/android/card/CardRequestFactory.kt
+++ b/Card/src/main/java/com/paypal/android/card/CardRequestFactory.kt
@@ -31,7 +31,7 @@ internal class CardRequestFactory {
             }
 
             val bodyJSON = JSONObject()
-            threeDSecureRequest?.also {
+            threeDSecureRequest?.let {
                 val verificationJSON = JSONObject()
                     .put("method", it.sca.name)
                 val attributesJSON = JSONObject()
--- a/Card/src/main/java/com/paypal/android/card/CardRequestFactory.kt
+++ b/Card/src/main/java/com/paypal/android/card/CardRequestFactory.kt
@@ -38,7 +38,6 @@ internal class CardRequestFactory {
                     .put("verification", verificationJSON)
                 cardJSON.put("attributes", attributesJSON)
 
-                // add return and cancel url when its supported
                 val returnURLJSON = JSONObject()
                     .put("return_url", it.returnUrl)
                     .put("cancel_url", it.cancelUrl)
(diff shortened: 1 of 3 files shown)


jcnoriega (approved):

Card/src/main/java/com/paypal/android/card/CardRequestFactory.kt
```diff
@@ -30,18 +30,25 @@ internal class CardRequestFactory {
...
+            val bodyJSON = JSONObject()
             threeDSecureRequest?.also {
                 val verificationJSON = JSONObject()
                     .put("method", it.sca.name)
                 val attributesJSON = JSONObject()
                     .put("verification", verificationJSON)
                 cardJSON.put("attributes", attributesJSON)
+
                 // add return and cancel url when its supported
+                val applicationContextJSON = JSONObject()
```
kgangineni: nit: `contextUrLJson` or `returnUrlJson` just because we have applicationContext meaning different thing in andorid

kgangineni (approved):

Card/src/main/java/com/paypal/android/card/CardRequestFactory.kt
sshropshire: That's a good call yeah. Went with `returnURLJSON` to differentiate 👍.

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
```diff
@@ -223,13 +222,6 @@ class CardFragment : Fragment() {
             ),
             payee = Payee(emailAddress = "anpelaez@paypal.com")
         )
-
-        if (shouldRequestThreeDSecure) {
-            createOrderRequest.applicationContext = ApplicationContext(
-                returnURL = APP_RETURN_URL,
-                cancelURL = APP_CANCEL_URL
-            )
-        }
```
scannillo, line 232: I notice this step isn't in our [docs](https://github.com/paypal/Android-SDK/tree/main/docs/Card#3-create-an-order). Was the plan always to remove this from the order creation step?

Card/src/main/java/com/paypal/android/card/CardRequestFactory.kt
```diff
@@ -30,18 +30,25 @@ internal class CardRequestFactory {
...
                     .put("method", it.sca.name)
                 val attributesJSON = JSONObject()
                     .put("verification", verificationJSON)
                 cardJSON.put("attributes", attributesJSON)
+
                 // add return and cancel url when its supported
+                val returnURLJSON = JSONObject()
+                    .put("return_url", it.returnUrl)
+                    .put("cancel_url", it.cancelUrl)
+                bodyJSON.put("application_context", returnURLJSON)
```
scannillo, line 44: We should update our `CardRequestFactoryUnitTests` to make sure these URLs are properly set in the request body to `confirm-payment-source`.

mattwylder: Where do `returnUrl` and `cancelUrl` come from? It seems like we used to expect the merchant to set them in the Fragment, but now we're getting them elsewhere?

mattwylder: Also, what do we mean by "when it's supported" in the comment? When do we expect it to be supported and when do we not?

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
sshropshire: Originally we were using LSATs and `confirm-payment-source` for this call, but then we had to reach out to that team because there was a bug where we couldn't update the return url after order creation. Once they fixed it we verified with this PR to allow the Client SDK to set the return and cancel URLs.

Temporarily we set them on the sample merchant server as a workaround and we planned on migrating to this method once the fix was in.

Card/src/main/java/com/paypal/android/card/CardRequestFactory.kt
sshropshire: Yeah i'm not sure why it isn't showing in the diff I guess we had some duplicated logic. We still do still [set them in CardFragment](https://github.com/paypal/Android-SDK/blob/8ae947a7b8f01f92e104e47971a1cb1f9135b188/Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt#L201).

sshropshire: And the "when it's supported" is referencing the bug. That should be a TODO I can remove that now.

scannillo (approved):
Do we have a ticket to do the same on iOS?

sshropshire:
> Do we have a ticket to do the same on iOS?

We don't have a ticket yet, but this should also be implemented on iOS.

We kind of got rushed into doing it on Android because the team who implemented this feature for us needed explicit confirmation of it working.

Card/src/main/java/com/paypal/android/card/CardRequestFactory.kt
sshropshire: > We should update our `CardRequestFactoryUnitTests` to make sure these URLs are properly set in the request body to `confirm-payment-source`.

Updated the `CardRequestFactoryUnitTests` to reflect the new change.

PR #83, opened 2022-08-11
NXO Demo: ViewModel and UseCases

jcnoriega:

### Summary of changes

 -  This PR adds the viewmodel and use cases for the NXO integration demo. The UI will be done in another PR

 ### Checklist

 - [ ] Added a changelog entry

- @jcnoriega


sshropshire (approved):

Demo/src/main/java/com/paypal/android/PayPalConfigConstants.kt
```diff
@@ -0,0 +1,5 @@
+package com.paypal.android
+
+class PayPalConfigConstants {
+    val returnUrl = "${BuildConfig.APPLICATION_ID}://paypalpay"
```
kgangineni, line 6: do we still need it, in iOS it works without returnUrl

Demo/src/main/java/com/paypal/android/api/services/PayPalApi.kt
```diff
@@ -2,15 +2,22 @@ package com.paypal.android.api.services
 
 import com.google.gson.JsonObject
 import com.paypal.android.api.model.Order
+import com.paypal.android.data.vault.VaultSessionId
 import retrofit2.http.Body
 import retrofit2.http.Headers
 import retrofit2.http.POST
 
-interface OrdersV2Api {
+interface PayPalApi {
```
kgangineni, line 11: `PaymentTokensAPI`? because this is what it does

Demo/src/main/java/com/paypal/android/di/NetworkModule.kt
```diff
@@ -36,12 +36,12 @@ object NetworkModule {
...
 
     @Provides
-    fun providePaypalApi(): PayPalDemoApi {
+    fun provideSDKSampleServerApi(): SDKSampleServerApi {
         return provideApi(provideRetrofitService(SAMPLE_SERVER_BASE_URL))
     }
 
     @Provides
-    fun provideOrdersV2Api(): OrdersV2Api {
+    fun providePayPalApi(): PayPalApi {
```
kgangineni, line 44: change here, if renamed `PayPalApi`

Demo/src/main/java/com/paypal/android/viewmodels/PayPalNativeViewModel.kt
```diff
@@ -0,0 +1,170 @@
...
+@HiltViewModel
+class PayPalNativeViewModel @Inject constructor(
+    private val getBillingAgreementTokenUseCase: GetBillingAgreementTokenUseCase,
+    private val getAccessTokenUseCase: GetAccessTokenUseCase,
+    private val getOrderIdUseCase: GetOrderIdUseCase,
+    private val getApprovalSessionIdActionUseCase: GetApprovalSessionIdActionUseCase,
+    application: Application
+) : AndroidViewModel(application) {
+
+    private val getOrderUseCase: GetOrderUseCase = GetOrderUseCase()
```
kgangineni: this can be injected as constructor param

Demo/src/main/java/com/paypal/android/viewmodels/PayPalNativeViewModel.kt
```diff
@@ -0,0 +1,170 @@
...
+
+    private val payPalConstants = PayPalConfigConstants()
```
kgangineni: same here can be injected

Demo/src/main/java/com/paypal/android/viewmodels/PayPalNativeViewModel.kt
```diff
@@ -0,0 +1,170 @@
...
+    private fun initPayPalClient(accessToken: String) {
+        payPalClient = PayPalClient(
+            getApplication(),
+            CoreConfig(accessToken),
+            payPalConstants.returnUrl
+        )
+        payPalClient.listener = payPalListener
+    }
+
+    sealed class ViewState {
```
kgangineni: this looks like MVI, in that case, LGTM is this the way forward for demo app?
`NativeCheckoutViewState` to be specific on the class and can be in different file

Demo/src/main/java/com/paypal/android/api/services/PayPalApi.kt
jcnoriega: it also has `postOrderCheckout` which is part of the `ordersV2` api. I thought of it this way: we have the `paypal` endpoints that will go under here (until it gets too big), and we have the `sample_merchant_server` endpoints which go on the other interface.

Demo/src/main/java/com/paypal/android/viewmodels/PayPalNativeViewModel.kt
jcnoriega: Honestly i took this from the nxo repo as refernence. I dont mind whatever pattern we decide. And i dont think we should decide on one specific pattern, perhaps for diferent demos we could try diferent patterns and then decide which one is better.

Demo/src/main/java/com/paypal/android/api/services/PayPalApi.kt
sshropshire: I could see us reaching a point where we start to migrate interaction with these endpoints into the SDK to support new features.

Demo/src/main/java/com/paypal/android/viewmodels/PayPalNativeViewModel.kt
sshropshire: What are some alternatives to this pattern?

Demo/src/main/java/com/paypal/android/PayPalConfigConstants.kt
jcnoriega: yes, android integration still needs a returnUrl

Demo/src/main/java/com/paypal/android/viewmodels/PayPalNativeViewModel.kt
jcnoriega: @kgangineni i had to remove them all from the constructor because `detekt` was failing with `LongParameterList`, so I inject them now as variables

Demo/src/main/java/com/paypal/android/usecase/GetApprovalSessionIdActionUseCase.kt
```diff
@@ -0,0 +1,47 @@
...
+    private val payPalApi: PayPalApi
+) {
+
+    suspend operator fun invoke(): String? = withContext(Dispatchers.IO) {
+        val jsonRequest = JsonParser.parseString(APPROVAL_SESSION_ID_REQUEST) as JsonObject
+        val vaultSessionId = payPalApi.postApprovalSessionId(jsonRequest)
+
+        val approvalSessionIdLink = vaultSessionId
+            .links
+            ?.find { vaultLink -> vaultLink?.rel == "approve" }
```
sshropshire:
```suggestion
        val approvalSessionIdLink =
            vaultSessionId.links?.find { vaultLink -> vaultLink?.rel == "approve" }
```

Demo/src/main/java/com/paypal/android/viewmodels/PayPalNativeViewModel.kt
jcnoriega: well technically MVVM is diferent, you post "data" to the view, not an entire `ViewState`. In MVI, the view subscribes to one stream of view states and thats it. On MVVM, you subscribe to diferent data streams

sshropshire: Yeah there are so many different patterns on Android. MVP was the main recommendation not too long ago. I'd vote for whatever scales the best. As we add more features to Northstar maintenance costs will go up in the Demo app.

jcnoriega: Agree. We could use the Demo app as a guinea pig to try diferent patterns. Like if someone wants to do the Venmo demo with full MVVM, then go ahead. We will evenutally figure out which patterns suits our interests

Demo/src/main/java/com/paypal/android/viewmodels/PayPalNativeViewModel.kt
kgangineni: in these type of cases, it is better to suppress detekt because this is genuine case of we requiring more paramters I see u already made changes to all dependencies to be `lateinit var` that's good too

kgangineni (approved):

PR #84, opened 2022-08-12
Create CoreCoroutineExceptionHandler

jcnoriega:

### Summary of changes

 -  This PR creates a CoreCoroutineExceptionHandler that works as a last-resort mechanism for global "catch all" behavior. Essentially, within a corotuine scope that isnt handled, will end up on the CoreCoroutineExceptionHandler, transforming that error into a PayPalSDKError

 ### Checklist

 - [ ] Added a changelog entry

-


sshropshire (approved):

kgangineni (approved):
👍

PR #85, opened 2022-08-15
NXO Demo: UI complete

jcnoriega:

### Summary of changes

 - This PR completed the UI for Native Checkout Demo
 

 ### Checklist

 - [ ] Added a changelog entry

-


andrespelaezp (approved):

Demo/src/main/java/com/paypal/android/api/services/PayPalApi.kt
```diff
@@ -18,6 +19,7 @@ interface PayPalApi {
     @Headers("Accept: application/json")
     @POST("/v2/vault/payment-tokens")
     suspend fun postApprovalSessionId(
+        @Header("Authorization") accessToken: String,
```
kgangineni, line 22: Q: why is this needed?

kgangineni (approved):
LGTM

Demo/src/main/java/com/paypal/android/api/services/PayPalApi.kt
jcnoriega: `v2/vault/payment-tokens` requires and Access Token to work

PR #88, opened 2022-08-22
NXO Demo: add patch order after shipping change

jcnoriega:

### Summary of changes

 - This PR patches an order after a change in shipping address

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/Demo/src/main/java/com/paypal/android/usecase/GetBillingAgreementTokenWithoutPurchaseUseCase.kt
+++ b/Demo/src/main/java/com/paypal/android/usecase/GetBillingAgreementTokenWithoutPurchaseUseCase.kt
@@ -1,0 +1,52 @@
+package com.paypal.android.usecase
+
+import com.google.gson.JsonObject
+import com.google.gson.JsonParser
+import com.paypal.android.api.services.PayPalApi
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import javax.inject.Inject
+
+class GetBillingAgreementTokenWithoutPurchaseUseCase
+    @Inject constructor(private val payPalApi: PayPalApi) {
+
+    suspend operator fun invoke(accessToken: String): String = withContext(Dispatchers.IO) {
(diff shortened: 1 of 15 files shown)


Demo/src/main/java/com/paypal/android/viewmodels/PayPalNativeViewModel.kt
```diff
@@ -75,7 +92,43 @@ class PayPalNativeViewModel @Inject constructor(
...
+                ),
+                PatchAmount.Replace(
+                    purchaseUnitReferenceId = "PUHF",
+                    amount = getAmount(
+                        value = order.purchaseUnitList.first().amount.value,
+                        shippingValue = updatedShippingAmount ?: "0.00"
+                    )
+                )
+            )
+            shippingChangeActions.patchOrder(patchRequest) {
```
scannillo, line 129: What does it mean for a merchant to patch an order? Basically to amend an existing order?

jcnoriega: I think so. Its for when a shipping address is changed or a shipping method. The order needs to be updated with the new value.

Demo/src/main/java/com/paypal/android/usecase/GetBillingAgreementTokenWithoutPurchaseUseCase.kt
```diff
@@ -0,0 +1,52 @@
+package com.paypal.android.usecase
+
+import com.google.gson.JsonObject
+import com.google.gson.JsonParser
+import com.paypal.android.api.services.PayPalApi
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import javax.inject.Inject
+
+class GetBillingAgreementTokenWithoutPurchaseUseCase
```
scannillo, line 10: What is a billing agreement without purchasing? Is this the same as vault without purchase?

scannillo:
Is Northstar Android out for beta release like iOS? If so - this needs a CHANGELOG entry.

PR #94, opened 2022-10-13
Remove Native Checkout exit survey

mattwylder:
### Reasons for Change
On iOS & Android, we want a customer to be able to cancel out of the PayPal flow without needing to complete an exit survey. This occurs by default in the NXO SDK.
Currently, when they try to exit the Native PayPal flow - they are required to complete an "exit survey", increasing friction for them to change to a different payment method.

### Summary of changes

 - Disable the exit survey before Native Checkout starts.

 ### Checklist

~ - [ ] Added a changelog entry~

- @mattwylder @sshropshire @jcnoriega


jaxdesmarais (approved):
👻

jcnoriega (approved):

PR #95, opened 2022-10-13
NXO: bump checkout to 0.8.2

jcnoriega:
### Summary of changes

 Bump NXO to 0.8.2, and updated accordingly:
 - Remove return url
 - Can not patch order if flow was started with an order Id. Reported the bug.
 - Updated documentation and demo app

 ### Checklist

 - [ ] Added a changelog entry

- @jcnoriega


sshropshire (approved):

Demo/src/main/java/com/paypal/android/ui/paypal/PayPalFragment.kt
```diff
@@ -110,8 +109,8 @@ class PayPalFragment : Fragment(), PayPalWebCheckoutListener {
             try {
                 binding.statusText.setText(R.string.creating_order)
 
-                val orderJson = JsonParser.parseString(OrderUtils.orderWithShipping) as JsonObject
-                val order = sdkSampleServerApi.createOrder(orderJson)
+                val orderRequest = OrderUtils.createOrderBuilder("100.0")
```
jaxdesmarais: nit - can we make this amount like ~$5 or less? Just on the off chance that someone ever completes a transaction in prod accidentally it'll make sure it's a much smaller amount.

Demo/src/main/java/com/paypal/android/usecase/GetOrderIdUseCase.kt
```diff
@@ -15,13 +13,11 @@ class GetOrderIdUseCase@Inject constructor(
...
         withContext(Dispatchers.IO) {
-            val orderRequestString = when (shippingPreferenceType) {
-                ShippingPreferenceType.GET_FROM_FILE -> OrderUtils.orderWithShipping
-                ShippingPreferenceType.NO_SHIPPING -> OrderUtils.orderWithoutShipping
-                ShippingPreferenceType.SET_PROVIDED_ADDRESS -> OrderUtils.orderWithProvidedAddress
-            }
-            val jsonOrder = JsonParser.parseString(orderRequestString) as JsonObject
-            val result = sdkSampleServerApi.createOrder(jsonOrder)
+            val order = OrderUtils.createOrderBuilder(
+                "100.0",
```
jaxdesmarais: same comment here on just doing a smaller amount

jaxdesmarais (approved):
👻

Demo/src/main/java/com/paypal/android/utils/OrderUtils.kt
```diff
@@ -0,0 +1,332 @@
...
+    fun getAmount(
+        currency: CurrencyCode = CurrencyCode.USD,
+        value: String = "0.01",
+        shippingValue: String = "0.00"
+    ): Amount {
+        val valueAsFloat = value.toFloat()
+        val shippingValueAsFloat = shippingValue.toFloat()
+        val totalValue = valueAsFloat + shippingValueAsFloat
+
+        return Amount.Builder()
```
mattwylder, line 94: Question from ignorance: is the Builder pattern a Kotlin thing?

jcnoriega: No. The "Amount" class is taken from the NXO folks.
The builder pattern is a common thing on `java`. My guess is that they implemented the builder pattern to make it easier for merchants using `java`.
Kotlin by itself doesnt need a builder pattern, because it has optional parameters. `java` doesnt have that

PR #98, opened 2022-10-18
Add debug and release implementation

jcnoriega:

### Summary of changes

 - This PR adds debug and release implementation to the Demo app, allowing to test SNAPSHOT versions. Another PR is going to be made so that we can test released versions as well.

 ### Checklist

 - [ ] Added a changelog entry

- @jcnoriega

### DIFF
--- a/Demo/build.gradle
+++ b/Demo/build.gradle
@@ -59,12 +59,18 @@ android {
 }
 
 dependencies {
-    implementation project(':Card')
-    implementation project(':PayPalDataCollector')
-    implementation project(':PayPalNativeCheckout')
-    implementation project(':PayPalWebCheckout')
-    implementation project(':PayPalUI')
-    implementation project(':Venmo')
+    debugImplementation project(':Card')
+    debugImplementation project(':PayPalDataCollector')
+    debugImplementation project(':PayPalNativeCheckout')
+    debugImplementation project(':PayPalWebCheckout')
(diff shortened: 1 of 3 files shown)


kgangineni (approved):

Demo/build.gradle
```diff
@@ -59,12 +60,18 @@ android {
...
+    debugImplementation project(':PayPalNativeCheckout')
+    debugImplementation project(':PayPalWebCheckout')
+    debugImplementation project(':PayPalUI')
+    debugImplementation project(':Venmo')
+
+    releaseImplementation deps.card
+    releaseImplementation deps.paypalUI
+    releaseImplementation deps.paypalWebCheckout
+    releaseImplementation deps.paypalNativeCheckout
+    releaseImplementation deps.paypalDataCollector
```
sshropshire, line 74: q: Is there a reason these need to duplicated? I notice there's an extra Venmo dependency between the two but overall these look similar.

mattwylder (approved):

Demo/build.gradle
jcnoriega: the idea is that if under Demo you change the build variant to `release` it will pull the `releaseImplementation` dependencies. Which in this case point to the `SNAPSHOT` version. The idea in the future, is to somehow be able to test in the Demo app the `snapshot` versions, the `released` versions and the local ones (local ones are used when you are developing)

PR #103, opened 2022-11-16
Card: Remove ThreeDSecureRequest and only ask for return_url

jcnoriega:

### Summary of changes
[Jira ticket](https://engineering.paypalcorp.com/jira/browse/DTNOR-635)
 -  We dont need to ask merchants for `return` and `cancel` urls, we can only ask for `return` and set the same url in `ApplicationContext`. So we remove `ThreeDSecureRequest` and simply ask for it in `CardReques`. Also update browserSwitch library to latest.
 - Also some test clean up: removing deprecated and unnecessary coroutine code from tests
 - Remove `Demo` module from static analysis and build

 ### Checklist

 - [x] Added a changelog entry

- @jcnoriega

### DIFF
--- a/Card/src/main/java/com/paypal/android/card/CardRequest.kt
+++ b/Card/src/main/java/com/paypal/android/card/CardRequest.kt
@@ -1,9 +1,17 @@
 package com.paypal.android.card
 
-import com.paypal.android.card.threedsecure.ThreeDSecureRequest
+import com.paypal.android.card.threedsecure.SCA
 
-data class CardRequest(
+/**
+ * A card request to process a card payment
+ * @property orderID The order ID to to process the payment
+ * @property card   Card used for payment
+ * @property returnUrl Url to return to app after SCA challenge finishes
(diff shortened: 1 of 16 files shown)


Card/src/main/java/com/paypal/android/card/CardRequest.kt
```diff
@@ -1,9 +1,17 @@
...
+ * @property returnUrl Url to return to app after SCA challenge finishes
+ * @property sca Specify to always launch 3DS or only when required. Defaults to `SCA.SCA_WHEN_REQUIRED`.
+ */
+data class CardRequest @JvmOverloads constructor(
     val orderID: String,
     val card: Card,
-    var threeDSecureRequest: ThreeDSecureRequest
+    val returnUrl: String,
+    val sca: SCA = SCA.SCA_WHEN_REQUIRED
 )
```
scannillo, line 17: I like this change! Now that 3DS is non-optional, and there was only 2 params on it, this is more explicit.

scannillo (approved):
🎸

sshropshire (approved):

PR #104, opened 2022-11-18
Initial FPTI Networking Capability

scannillo:
### Reason for changes

 - See summary of https://github.com/paypal/iOS-SDK/pull/91

### Summary of changes

- Add analytics specific method on `HttpRequestFactory` to construct `HttpRequest`
- Add `AnalyticsEventData` data class to model JSON post structure
- Add `DeviceInspector` class to perform all  app `Context` and build specific information extraction 

 ### Checklist

 - ~Added a changelog entry~ Will update CHANGELOG once events are sent.

### Authors
@sshropshire @scannillo @kgangineni


Core/src/androidTest/java/com/paypal/android/core/HttpIntegrationTest.kt
```diff
@@ -9,6 +9,7 @@ import org.junit.Assert.assertEquals
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
+import java.net.URL
```
scannillo, line 12:
```suggestion
```

Core/src/main/java/com/paypal/android/core/API.kt
```diff
@@ -1,13 +1,30 @@
...
+import android.content.Context
+import android.util.Log
+import androidx.annotation.VisibleForTesting
+import com.paypal.android.core.analytics.AnalyticsEventData
+import com.paypal.android.core.analytics.DeviceInspector
+import java.util.UUID
+
 class API internal constructor(
     private val configuration: CoreConfig,
+    private val applicationContext: Context,
```
jcnoriega: `API` doesnt need the `applicationContext`, why dont we pass in `DeviceInspector` constructor?

jcnoriega:
Also why are not the PR checks being run? maybe the PR needs to be assinged for it to work?

Core/src/main/java/com/paypal/android/core/API.kt
sshropshire: It's used in the `sendAnalyticsEvent()` method. It was better for testability.

sshropshire:
> Also why are not the PR checks being run? maybe the PR needs to be assinged for it to work?

I'm not sure about this actually. They aren't active on this branch for some reason.

Core/src/main/java/com/paypal/android/core/API.kt
jcnoriega: im just wondering why can it not live inside `DeviceInspector`, where it is actually used, does it affect testability that much? Feels like an uneccessary variable inside `API`

sshropshire: We had it like that at first, but then `DeviceInspector` had its own properties like `appName` and `appId` which felt kind of odd, since those properties also exist on `DeviceData`.

`DeviceInspector#inspect(Context)` prevents `DeviceInspector` from duplicating those properties by returning all device metadata in a single data class.

Core/src/main/java/com/paypal/android/core/API.kt
```diff
@@ -1,13 +1,30 @@
...
+    private val deviceInspector: DeviceInspector
 ) {
 
-    constructor(configuration: CoreConfig) :
-            this(configuration, Http(), HttpRequestFactory())
+    constructor(configuration: CoreConfig, context: Context) :
+            this(
+                configuration,
+                context.applicationContext,
+                UUID.randomUUID().toString().replace("-", ""),
```
kgangineni, line 25: this creates new `uuid` everytime, `API` is not singleton if merchant creates multiple instances of API then there will be multiple session ids

Core/src/main/java/com/paypal/android/core/API.kt
```diff
@@ -41,4 +58,19 @@ class API internal constructor(
         }
         return clientID
     }
+
+    suspend fun sendAnalyticsEvent(name: String) {
```
kgangineni, line 64: can this be `private` it will not be used by merchants

Core/src/main/java/com/paypal/android/core/analytics/DeviceInspector.kt
```diff
@@ -0,0 +1,63 @@
...
+    constructor() : this(
+        clientSDKVersion = BuildConfig.PAYPAL_SDK_VERSION,
+        sdkInt = Build.VERSION.SDK_INT,
+        deviceManufacturer = Build.MANUFACTURER,
+        deviceModel = Build.MODEL,
+        deviceProduct = Build.PRODUCT,
+        deviceFingerprint = Build.FINGERPRINT
+    )
+
+    fun inspect(context: Context): DeviceData {
```
kgangineni: `context` here sent as constructor param, saved in `WeakReference`, then we can get rid of `context` in `API`
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.native.ref/-weak-reference/

Core/src/main/java/com/paypal/android/core/analytics/DeviceInspector.kt
```diff
@@ -0,0 +1,63 @@
...
+        val packageManager = context.packageManager
+
+        val appName = try {
+            val applicationInfo = packageManager.getApplicationInfo(context.packageName, 0)
+            packageManager.getApplicationLabel(applicationInfo).toString()
+        } catch (ignored: PackageManager.NameNotFoundException) {
+            ""
```
kgangineni: need to define default app name for failure scenarios

scannillo: We intentionally wanted the value to be for example `"merchantAppName": ""` in the Analytics JSON. I think this is more explicit when inspecting the metadata than having us inject some string value like `"merchantAppName": "UnknownApp"`. 

We also did this on iOS - what do folks think?

Core/src/main/java/com/paypal/android/core/API.kt
sshropshire: This is a good point. How can we make it so that a merchant only instantiates one `API` instance?

Core/src/main/java/com/paypal/android/core/API.kt
sshropshire: I tried both `private` and `internal` but it broke compilation since it's currently used by `CardAPI`.

Core/src/main/java/com/paypal/android/core/analytics/DeviceInspector.kt
sshropshire: We currently call `Context#getApplicationContext()` to store a long lived reference. Do we need a `WeakReference` here since we're storing a reference to the Application's context?

Core/src/main/java/com/paypal/android/core/analytics/DeviceInspector.kt
sshropshire: We could add something explicit i'm indifferent.

Core/src/main/java/com/paypal/android/core/API.kt
scannillo: I think this is a good callout, but a 🏈 punt for another PR / JIRA ticket.

I think the end desired behavior is that we create a new sessionID only if a new API was instantiated with a different Auth credential, otherwise - we use the same sessionID.

Core/src/main/java/com/paypal/android/core/analytics/DeviceInspector.kt
```diff
@@ -0,0 +1,63 @@
...
+
+        val appId = context.packageName
+        val clientOS = "Android API $sdkInt"
+
+        val isSimulator = "google_sdk".equals(deviceProduct, ignoreCase = true) ||
+                "sdk".equals(deviceProduct, ignoreCase = true) ||
+                "Genymotion".equals(deviceManufacturer, ignoreCase = true) ||
+                deviceFingerprint.contains("generic")
+
+        return DeviceData(
```
scannillo, line 56: Instead of needing an additional type `DeviceData`, can we instead have getters on `DeviceInspector` for these properties?

Just thinking if there's a way we could simplify a bit. I think `DeviceData` vs `DeviceInspector` is a bit confusing.

Core/build.gradle
```diff
@@ -15,12 +15,12 @@ android {
         versionName modules.sdkVersionName
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         consumerProguardFiles "consumer-rules.pro"
+        buildConfigField "String", "PAYPAL_SDK_VERSION", "\"$versionName\""
     }
 
     buildTypes {
 
         release {
-            version = android.defaultConfig.versionName //had to add this to get the version in the build
```
scannillo, line 23: Was this meant to be removed?

Core/src/main/java/com/paypal/android/core/API.kt
jcnoriega: The merchant shouldnt use the API class, its meant for internal use only, we have it public because if not other modules cant access it. If they DO use it, they do so at their own risk (we should add this to the docs).
Even so, i think the `API` class should end up being agnostic of session IDs, analytics events, etc. It does seem to me that the `sessiongID`, should be tied to the future `FPTIClient` that will be bounded to a feature client. Meaning, that a sessionId should be created when the `FPTIClient` is instantiated, and the `FPTIClient` will be instantiated whenever a feature client (e.g `CardClient`) is instantiated. Something like: 

```kotlin
class CardClient {
    private val fptiClient: FPTIClient
    
    init {
        fptiClient = FPTIClient()
    }
    
}

class FPTIClient {

    private val sessionID = UUID.randomUUID().toString().replace("-", "")

}
```

If we are going to create an `FTPIClient` on a diferent PR later on, then i think that the `sessionID` should live in the feature client for now, and pass it as parameter to the analytics call

Core/src/main/java/com/paypal/android/core/analytics/DeviceInspector.kt
sshropshire: @jcnoriega brought up the same concern [here](https://github.com/paypal/Android-SDK/pull/104/files/4185eff824ffaabe45b1eaad76afe127f8510dc5#r1026630726). Overall I agree, except that it has potential to make testing more difficult.

To make it simpler, I would remove the `DeviceInspector` class and add a `DeviceData` constructor that takes a `Context`.

In the current design, something about having properties on the device inspector feels odd since it only has one job to "inspect" a given context. I'd be down to pair with y'all to get it merged in quicker and align our design principles.

Core/src/main/java/com/paypal/android/core/API.kt
sshropshire: Yeah we've always had an implicit definition for sessionID. I think it would make sense to update this in a separate PR and make the creation of a session ID more explicit.

Core/build.gradle
sshropshire: Probably so that isn't a common property for this gradle block.

Core/src/main/java/com/paypal/android/core/analytics/DeviceInspector.kt
scannillo: I _personally_ think `"merchantAppName": ""` or maybe even `"merchantAppName": "N/A"` is the most clear that no data was sent for a specific field. What do folks think? Whatever we choose - let's align on all fields and across iOS.

sshropshire: I like "N/A." I think it makes sense to have a default value for the rare case where we're unable to determine the merchant's app name.

Core/src/main/java/com/paypal/android/core/API.kt
sshropshire: Also willing to compromise here. The biggest thing is mocking constructors is difficult. I'm up for `DeviceInspector` with properties.

Core/src/main/java/com/paypal/android/core/analytics/DeviceInspector.kt
sshropshire: Hey @kgangineni I removed this method in favor of a single `DeviceInspector` class.

Core/src/main/java/com/paypal/android/core/analytics/DeviceInspector.kt
sshropshire: @scannillo @jcnoriega ended up removing `DeviceData`.

Core/src/test/java/com/paypal/android/core/HttpRequestFactoryUnitTest.kt
```diff
@@ -1,130 +1,167 @@
...
-import org.robolectric.ParameterizedRobolectricTestRunner
 import org.robolectric.RobolectricTestRunner
+import org.skyscreamer.jsonassert.JSONAssert
 import java.net.URL
 
-@RunWith(Enclosed::class)
+@RunWith(RobolectricTestRunner::class)
 class HttpRequestFactoryUnitTest {
 
-    @RunWith(ParameterizedRobolectricTestRunner::class)
```
sshropshire, line 16: Ended up simplifying this test by removing the parameterized portion.

Core/src/main/java/com/paypal/android/core/API.kt
scannillo: 👋 These are great callouts. We're going to tackle them in follow up PRs.

1) Move to singletons so each access token is associated with only 1 session ID
2) Consolidate analytics logic (sessionID and sending events) into FPTIClient

scannillo: >  If they DO use it, they do so at their own risk (we should add this to the docs).

Great call out - Updatedthe docstring in this PR

Core/src/main/java/com/paypal/android/core/analytics/DeviceInspector.kt
scannillo: 👋 Sorry for the back and forth. I'm realizing now my suggestion to consolidate is an anti-pattern for the SDK and we prefer the `data class` types not to have logic and really only just do formatting if necessary (like `AnalyticsEventData`). Is that an accurate statement?

Core/src/test/java/com/paypal/android/core/HttpRequestFactoryUnitTest.kt
scannillo: Awesome - thanks for tackling that!

Core/src/main/java/com/paypal/android/core/analytics/DeviceInspector.kt
scannillo: @kgangineni - what do you think? Revert back to `DeviceData`?

Core/src/main/java/com/paypal/android/core/API.kt
scannillo: > @jcnoriega that a sessionId should be created when the FPTIClient is instantiated, and the FPTIClient will be instantiated whenever a feature client (e.g CardClient) is instantiated

The problem with this is that a sessionID should be tied to an accessToken, not the initialization of a feature. So it is a 1:1 of accessToken: sessionID. The idea here is the accessToken represents the "checkout session". We want to be able to see within 1 checkout session that a customer maybe attempted BOTH cards and PayPal. Let me know if this makes sense.

jcnoriega: It does, but we have no way of guarantee that, because the access token is created on the merchant side. The only way i can think of, is that the `sessionId` is created in the `CoreConfig` class as an internal static variable. It would also need to store the access token as a static variable as well, so that access token lives across multiple CoreConfigs. Or make CoreConfig a singleton. There are lots of ways to go about this i think.

jcnoriega (approved):

Core/src/main/java/com/paypal/android/core/API.kt
jcnoriega: @scannillo the `sessionId` could be handled in `CoreConfig`, something like this:

```kotlin
class CoreConfig @JvmOverloads constructor(
    val accessToken: String? = null,
    val environment: Environment = Environment.SANDBOX,
) {
    companion object {
        private var sessionID: String? = null
        private var activeAccessToken: String? = null
    }

    init {
        if (accessToken != activeAccessToken) {
            activeAccessToken = accessToken
            sessionID = UUID.randomUUID().toString().replace("-", "")
        }
    }

    fun getSessionId() = sessionID
}
```

scannillo (commented):
✅ Approve

Core/src/main/java/com/paypal/android/core/API.kt
kgangineni: We will create something like `AnalyticsClient` which will be a singleton instance, this will have a session id, generated per instance
this way we will one session id regardless of how many times we create access token

Core/src/main/java/com/paypal/android/core/analytics/DeviceInspector.kt
kgangineni: moved context out from `API` to `DeviceInspector` returning `DeviceData`

kgangineni (approved):

PR #107, opened 2022-12-05
Single session id for analytics

kgangineni:

### Summary of changes
[Jira](https://engineering.paypalcorp.com/jira/browse/DTNOR-644)
 - Single session Id for all FPTI events in same session
 - delegate analytics to AnalyticsClient

 ### Checklist

 - [ ] Added a changelog entry

- @kgangineni @scannillo


sshropshire (approved):

Core/src/main/java/com/paypal/android/core/analytics/AnalyticsClient.kt
```diff
@@ -0,0 +1,31 @@
+package com.paypal.android.core.analytics
+
+import android.util.Log
+import com.paypal.android.core.Http
+import com.paypal.android.core.HttpRequestFactory
+import java.util.UUID
+
+internal class AnalyticsClient(
```
sshropshire: q: is the goal to instantiate a new `AnalyticsClient` for each feature client?

Core/src/main/java/com/paypal/android/core/API.kt
```diff
@@ -1,31 +1,20 @@
...
 class API internal constructor(
     private val configuration: CoreConfig,
-    private val sessionID: String,
     private val http: Http,
     private val httpRequestFactory: HttpRequestFactory,
-    private val deviceInspector: DeviceInspector
 ) {
 
-    constructor(configuration: CoreConfig, context: Context) :
+    constructor(configuration: CoreConfig) :
```
scannillo: Each feature client will need to still pass along a `Context` at some point.

I think we should keep the `context` as a dependency of `API`, so that each feature client can easily just call `API.sendEvent("card.started")`, for example. That method call, will delegate to an `AnalyticsService` so none of the analytics specific logic lives in `API`. Here's how I thought of doing it on iOS (https://github.com/paypal/iOS-SDK/pull/100).

With this current approach, each feature client would need _both_ an `API` dependency and an `AnalyticsClient` dependency. What do you think?

Core/src/main/java/com/paypal/android/core/analytics/AnalyticsClient.kt
scannillo: Similar question [here](https://github.com/paypal/Android-SDK/pull/107#discussion_r1039909042)

kgangineni: @sshropshire moving it to `API`

Core/src/main/java/com/paypal/android/core/analytics/AnalyticsClient.kt
```diff
@@ -0,0 +1,30 @@
...
+        val httpRequest = httpRequestFactory.createHttpRequestForAnalytics(analyticsEventData)
+
+        val response = http.send(httpRequest)
+        if (!response.isSuccessful) {
+            Log.d("[PayPal SDK]", "Failed to send analytics: ${response.error?.message}")
+        }
+    }
+
+    companion object {
+        private val sessionId = UUID.randomUUID().toString()
```
jcnoriega, line 28: this is equal to `static` so we will have only one session ID as long as the merchant app is still in memory

jcnoriega (approved):
I wonder if in the future it would be better to have feature client API. For example, something like `CardAPI` where it has `confirmPaymentSource()` and other `Card` related `API` calls. All of this `api` clients could extend `API`, where the `AnalyticsClient` would live.

Core/src/main/java/com/paypal/android/core/analytics/AnalyticsClient.kt
```diff
@@ -0,0 +1,30 @@
+package com.paypal.android.core.analytics
+
+import android.util.Log
+import com.paypal.android.core.Http
+import com.paypal.android.core.HttpRequestFactory
+import java.util.UUID
+
+internal class AnalyticsClient(
```
scannillo: Nitpick / naming comment - I think we use the word `Client` a lot, and it's meaning is a bit conflated. I opted for `AnalyticsService` on iOS, since the android feature modules aren't interfacing directly with this `Analytics` layer & it is what we use on BT.

I don't feel too strongly, but it would be nice if we can align on the same name and decide what makes something a `Client`.

jcnoriega: Agree, perhaps we could reserve `Client` to the classes the the merchant should interact with.

scannillo:
Let's confirm in stand-up that we like the approach of `1 sessionID: 1 accessToken` (like in our [iOS discussion](https://github.com/paypal/iOS-SDK/pull/100#issuecomment-1348867435)). If so - we'll need to make an update to the `AnalyticsClient` introduced here.

PR #116, opened 2023-01-30
Bump NXO to 0.8.8

jcnoriega:

### Summary of changes

 - Bump nxo to 0.8.8 and add return url back

 ### Checklist

 - [x] Added a changelog entry

- @jcnoriega

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # PayPal Android SDK Release Notes
 
 ## unreleased
-* `PayPalNativePayments`: Bump NX0 to `0.8.8` and add `return_url`
+* `PayPalNativePayments`:
+  *  Bump `PayPal Native Checkout` to `0.8.8` and add `return_url`
 
 ## 0.0.7 (2023-01-25)
 * Rename `PayPalDataCollector` to `FraudProtection`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # PayPal Android SDK Release Notes
 
+## unreleased
+* `PayPalNativePayments`: Bump NX0 to `0.8.8` and add `return_url`
+
 ## 0.0.7 (2023-01-25)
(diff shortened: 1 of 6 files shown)


scannillo:
This seems odd to me that this is again required. Needing merchants to set this URL in their dev portal is a hefty barrier to integrating/developing quickly, also it was removed from iOS. Does the NXO team plan to remove this soon? Was there any reasoning why it was added back?

jcnoriega:
@scannillo i dont know, i was asking about the `return_url` on the `native-xo-sdks-integration` channel, and they informed me that in the latests version they had added it again. This is not the first time that they re add the `return_url`. Maybe they are still finding edge cases? i'll ask

jcnoriega:
@scannillo they removed it in version `0.7.2`, then added it back in `0.7.3`. Removed it again on `0.8.1`, and now they added it back on `0.8.8`

jaxdesmarais:
@scannillo @jcnoriega I asked the [same question here](https://github.com/braintree/braintree_android/pull/662#discussion_r1088033157). Sounds like it's an Android only change.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # PayPal Android SDK Release Notes
 
+## unreleased
+* `PayPalNativePayments`: Bump NX0 to `0.8.8` and add `return_url`
```
jaxdesmarais: Looks like below in the changelog we use this syntax. Not sure we have established a preference, but we likely should so that we are able to stay consistent. 
```suggestion
* `PayPalNativePayments`:
  * Bump NX0 to `0.8.8` and add `return_url`
```

jaxdesmarais: We should also update `NX0` to `PayPal Native Checkout` since merchants won't know what `NXO` stands for

jaxdesmarais (approved):

jcnoriega:
@jaxdesmarais thanks for that link! yeah i think we are going to have to expect this back and forth with the url

PR #118, opened 2023-02-08
ClientID Caching ability; value required for analytics

scannillo:
Equivalent iOS PR - https://github.com/paypal/iOS-SDK/pull/118
[JIRA](https://paypal.atlassian.net/browse/DTNOR-745)

## Reason for changes

This PR requires a clientID be successfully fetched (or cached) before proceeding with any of the 3 payment method flows (card, paypal web, and paypal native).

### Summary of changes

- Rename `API.getClientId()` --> `API.fetchCachedOrRemoteClientID()` + added docstring
     - Store `clientID` in static LRU Cache of max size 10 items; fetch from cache when possible
     - Key = accessToken, value = clientID
- Add new `APIClientError.clientIDNotFoundError` case
- Add clientID fetch to PayPalWebCheckoutClient, CardClient, PayPalNativeClient & updated respective unit test files
- Add `partner_client_id` to analytics payload metadata

### Next steps
- Will create follow-up tickets for doing an early background fetch for clientID once the merchant passes us their accessToken; but that work can happen post-GA.

 ### Checklist

 - ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
@@ -70,8 +70,9 @@ class CardClient internal constructor(
         try {
             cardAPI.fetchCachedOrRemoteClientID()
         } catch (e: PayPalSDKError) {
-            val merchantError = APIClientError.clientIDNotFoundError(e.code, e.correlationID)
-            approveOrderListener?.onApproveOrderFailure(merchantError)
+            approveOrderListener?.onApproveOrderFailure(
+                APIClientError.clientIDNotFoundError(e.code, e.correlationID)
+            )
             return
         }
 
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
@@ -60,13 +60,12 @@ class CardClient internal constructor(
      */
     fun approveOrder(activity: FragmentActivity, cardRequest: CardRequest) {
         CoroutineScope(dispatcher).launch(exceptionHandler) {
+            cardAPI.fetchClientID()
             confirmPaymentSource(activity, cardRequest)
         }
     }
 
     private suspend fun confirmPaymentSource(activity: FragmentActivity, cardRequest: CardRequest) {
-        cardAPI.fetchClientID()
-
         val response = cardAPI.confirmPaymentSource(cardRequest)
         if (response.payerActionHref == null) {
             val result = response.run {
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
@@ -65,6 +65,8 @@ class CardClient internal constructor(
     }
 
     private suspend fun confirmPaymentSource(activity: FragmentActivity, cardRequest: CardRequest) {
+        cardAPI.fetchClientID()
+
         val response = cardAPI.confirmPaymentSource(cardRequest)
         if (response.payerActionHref == null) {
             val result = response.run {
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
@@ -10,8 +10,10 @@ import com.paypal.android.cardpayments.api.CardAPI
 import com.paypal.android.cardpayments.api.GetOrderRequest
 import com.paypal.android.cardpayments.model.CardResult
 import com.paypal.android.corepayments.API
+import com.paypal.android.corepayments.APIClientError
 import com.paypal.android.corepayments.CoreConfig
 import com.paypal.android.corepayments.CoreCoroutineExceptionHandler
+import com.paypal.android.corepayments.PayPalSDKError
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -60,12 +62,19 @@ class CardClient internal constructor(
      */
     fun approveOrder(activity: FragmentActivity, cardRequest: CardRequest) {
         CoroutineScope(dispatcher).launch(exceptionHandler) {
-            cardAPI.fetchClientID()
             confirmPaymentSource(activity, cardRequest)
         }
     }
 
     private suspend fun confirmPaymentSource(activity: FragmentActivity, cardRequest: CardRequest) {
+        try {
+            cardAPI.fetchCachedOrRemoteClientID()
+        } catch (e: PayPalSDKError) {
+            val merchantError = APIClientError.clientIDNotFoundError(e.code, e.correlationID)
+            approveOrderListener?.onApproveOrderFailure(merchantError)
+            return
+        }
+
         val response = cardAPI.confirmPaymentSource(cardRequest)
         if (response.payerActionHref == null) {
             val result = response.run {
--- a/CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
+++ b/CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
@@ -1,6 +1,7 @@
 package com.paypal.android.corepayments
 
 import android.content.Context
+import android.util.LruCache
 import com.paypal.android.corepayments.analytics.AnalyticsService
 import com.paypal.android.corepayments.analytics.DeviceInspector
 
@@ -35,13 +36,19 @@ class API internal constructor(
 
     @Throws(PayPalSDKError::class)
     suspend fun getClientId(): String {
+        clientIDCache.get(configuration.accessToken)?.let {
+            return it
+        }
+
         val apiRequest = APIRequest("v1/oauth2/token", HttpMethod.GET)
         val httpRequest =
             httpRequestFactory.createHttpRequestFromAPIRequest(apiRequest, configuration)
         val response = http.send(httpRequest)
         val correlationID = response.headers["Paypal-Debug-Id"]
         if (response.isSuccessful) {
-            return parseClientId(response.body, correlationID)
+            val clientID = parseClientId(response.body, correlationID)
+            clientIDCache.put(configuration.accessToken, clientID)
+            return clientID
         }
 
         throw APIClientError.serverResponseError(correlationID)
@@ -62,5 +69,9 @@ class API internal constructor(
 
     suspend fun sendAnalyticsEvent(name: String) {
         analyticsService.sendAnalyticsEvent(name)
+    }
+
+    companion object {
+        val clientIDCache = LruCache<String, String>(10)
     }
 }
--- a/CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
+++ b/CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
@@ -5,7 +5,6 @@ import android.util.Log
 import android.util.LruCache
 import com.paypal.android.corepayments.analytics.AnalyticsService
 import com.paypal.android.corepayments.analytics.DeviceInspector
-import java.lang.Exception
 
 /**
  * This class is exposed for internal PayPal use only. Do not use.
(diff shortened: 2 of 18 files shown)


CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
```diff
@@ -33,15 +36,29 @@ class API internal constructor(
...
+
         val apiRequest = APIRequest("v1/oauth2/token", HttpMethod.GET)
         val httpRequest =
             httpRequestFactory.createHttpRequestFromAPIRequest(apiRequest, configuration)
         val response = http.send(httpRequest)
         val correlationID = response.headers["Paypal-Debug-Id"]
         if (response.isSuccessful) {
-            return parseClientId(response.body, correlationID)
+            val clientID = parseClientId(response.body, correlationID)
+            configuration.accessToken?.let { accessToken ->
```
jcnoriega, line 57: all of the `configuration.accessToken?.let { ... }` will go away once #117 is merged right?

CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
```diff
@@ -61,6 +78,15 @@ class API internal constructor(
...
+        try {
+            val clientID = fetchCachedOrRemoteClientID()
+            analyticsService.sendAnalyticsEvent(name, clientID)
+        } catch (e: Exception) {
+            Log.d("[PayPal SDK]", "Failed to send analytics due to missing clientID: ${e.message}")
+        }
+    }
+
+    companion object {
+        val clientIDCache = LruCache<String, String>(10)
```
jcnoriega, line 89: 🔥 🔥 on `LruCache`. Never used it before!

CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
scannillo: Yes, exactly!

PayPalWebPayments/src/test/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClientUnitTest.kt
```diff
@@ -7,53 +7,80 @@ import com.braintreepayments.api.BrowserSwitchClient
...
+
+    @Test
+    fun `start() starts browserSwitchClient with correct parameters`() = runTest {
+        val sut = getPayPalCheckoutClient(testScheduler = testScheduler)
+        val browserSwitchOptions = mockk<BrowserSwitchOptions>(relaxed = true)
+
+        coEvery { api.fetchCachedOrRemoteClientID() } returns "fake-client-id"
+
+        coEvery {
+            browserSwitchHelper.configurePayPalBrowserSwitchOptions(any(), any(), any())
```
scannillo, line 72: I'm not sure why, but the only way to get this test passing was to change from `any(), coreConfig, payPalRequest.fundingSource` params to `any(), any(), any()` for the stub 🤔 - none of the logic around browser switch options changed either

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -65,6 +67,15 @@ class CardClient internal constructor(
     }
 
     private suspend fun confirmPaymentSource(activity: FragmentActivity, cardRequest: CardRequest) {
+        try {
+            cardAPI.fetchCachedOrRemoteClientID()
```
sshropshire, line 71: q: should we fail if we can't get client id here? Confirm payment source may still be able to succeed with the access token.

scannillo: We decided in one of our group meetings w/ David and Walter that we want all flows to fail unless we have a clientID, even if the flow itself doesn't need a clientID.

CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
```diff
@@ -33,15 +35,29 @@ class API internal constructor(
...
+    /**
+     * Retrieves the merchant's clientID either from the local cache, or via an HTTP request if not cached.
+     * @return Merchant clientID.
+     */
     @Throws(PayPalSDKError::class)
-    suspend fun getClientId(): String {
+    suspend fun fetchCachedOrRemoteClientID(): String {
+        configuration.accessToken?.let { accessToken ->
+            clientIDCache.get(accessToken)?.let { cachedClientID ->
+                return cachedClientID
```
sshropshire, line 46: This might not return for the outer suspend function, I can't remember how it works in Kotlin though.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
scannillo: In follow up PRs, we want to fetch the clientID as soon as possible (when merchant gives ur their accessToken) similar to the config on BT. I figure that could be post-GA.

CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
scannillo: Ah, this might be moot once [this PR](https://github.com/paypal/Android-SDK/pull/117) is merged in? We'll be removing that outer `?.let` unwrap since accessToken will be non-optional.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
sshropshire: Oh ok got it. Would `fun prefetchClientID(): Unit` make sense? Or maybe it's OD.

CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
sshropshire: Potentially yeah. It's trickier with the nested closures. Technically those `let` functions do return a value though even though we're not capturing the output here. Something like
```kotlin
var clientID = configuration.accessToken?.let { clientIDCache.get(it) }
```
would be equivalent.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
scannillo: There is a flow where the value of `clientID` is actually needed within the flow (NXO) - [code](https://github.com/paypal/Android-SDK/blob/5e950e2d268fe7c1dd8bd9749b643b11458c38a2/PayPalNativePayments/src/main/java/com/paypal/android/paypalnativepayments/PayPalNativeCheckoutClient.kt#L67).

But yeh, the other 2 flows (card & PP web) ignore the return value in the flow, and only use it for analytics metadata.

Down to rename this if we think there is something more clear!

sshropshire: Yeah I would just to make the intention clear. It makes sense now but as the code evolves it might not be clear that it's a prefetch. Maybe we could have two methods? Like `suspend fun prefetchClientID(): Unit` and `suspend fun fetchClientID(): String`? The caller won't care if it's from a cache or the network, we can do some information hiding to keep the api clean too.

CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
scannillo: I'm a little confused - so you're saying there's a potential that `configuration.accessToken?.let` is returning the `accessToken` from the `fetchCachedOrRemoteClientID()` func, unless we specify a variable for it?

Does Kotlin not always require explicit `return`s?

sshropshire: In some cases Kotlin doesn't require `return`. In this case the compiler won't warn about the statement on line 44 returning a value that isn't used. In Swift we'd be required to put `let _ =` at the beginning of the line to get it to compile.

sshropshire: You are right though this may be less of a concern once that PR is merged.

scannillo: Running through the flow/tests, the function doesn't return the `accessToken` value if it is present. Also if the `accessToken` or `clientIDCache` checks return `null`, it does continue on to execute the actual HTTP fetch below.

So I think we are good here.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
scannillo: Yeh, that's a good call. There is no pre-fetch happening now. But when we do, we should make sure the naming is clear.

I added a note about this conversion in that follow-up [JIRA DTNOR-823](https://paypal.atlassian.net/browse/DTNOR-823).

sshropshire (approved):

PayPalWebPayments/src/test/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClientUnitTest.kt
sshropshire: It's probably because of a [Referential Equality](https://kotlinlang.org/docs/equality.html) check in this case since the custom types don't have any `equals(other: Any?)` methods implemented.

jcnoriega (approved):

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -65,6 +67,15 @@ class CardClient internal constructor(
...
+        } catch (e: PayPalSDKError) {
```
jaxdesmarais, line 72: General Kotlin/Android question here, but is `e` vs more explicit `error` the best practice for the language/platform? I'm still digging into Kotlin as a whole so am not totally sure what the language practices are.

CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
jaxdesmarais: Was just reading the docs on this and seems super cool! 🚀

CorePayments/src/test/java/com/paypal/android/corepayments/APIUnitTest.kt
```diff
@@ -11,8 +11,7 @@ import io.mockk.slot
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.test.runTest
 import org.json.JSONObject
-import org.junit.Assert.assertEquals
-import org.junit.Assert.assertSame
+import org.junit.Assert.*
```
jaxdesmarais, line 14: Basic Q again, I assume the `*` just imports all of `junit.Assert`? If so that's cool and I can see this being useful since some files have what feels like 40 imports. Are these imports generated automagically like if we use X amount from an import it switches to use the `*` or is it a manual process?

jaxdesmarais (approved):
Mainly basic Kotlin questions. 😅 This looks great otherwise!

CorePayments/src/test/java/com/paypal/android/corepayments/APIUnitTest.kt
scannillo: Yeh, the * imports everything within the namespace. We have [this Detekt rule](https://github.com/paypal/Android-SDK/blob/0d46271a16fc3193db01d5f8fa0c33bd1fae13f5/detekt/detekt-config.yml#L786) that does not allow wildcard imports in source files, but does allow them in test files. I am guessing we added it to avoid importing unnecessary classes in the SDK source itself.

The import statements were generated automatically by Android Studio (command + enter on a red class name usually adds the import statement for you). Some of them defaulted to * in the source file, which I had to change to specify the exact class, in order to satisfy the linter.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
scannillo: I'm not sure what best practice is, but we use `e` in the other try-catch blocks in the SDK, so I went with this for consistency. [Kotlin docs examples](https://kotlinlang.org/docs/exceptions.html) also use `e`

CorePayments/src/test/java/com/paypal/android/corepayments/APIUnitTest.kt
jaxdesmarais: Gotcha - that's cool, TIL!

PR #120, opened 2023-02-21
Add analytic events for PPWeb, PP NXO, and Card flows

scannillo:
### Reason

- Add analytics events to each payment method flow; unblock GA release
- Unit tests for analytics to be added in follow up PR - **DTNOR-826**

### Changes
- Send analytics events ([as outlined in this doc](https://paypal.sharepoint.com/:x:/r/sites/BTSDK/_layouts/15/Doc.aspx?sourcedoc=%7BAAA7C353-5744-4F21-AC7D-DF85D8605817%7D&file=Northstar%20Analytics%20Events.xlsx&action=default&mobileredirect=true&cid=1b0afccd-68ab-448f-9760-60ed14df9811)) for PayPal Web Payments, PayPal Native Payments, and Card Payment flows
- Other changes
    - Change component name to `ppcpmobilesdk` for Lighthouse IQ validations
    - Update APIClient.sendAnalyticsEvent() to send analytics in a background task

 ### Checklist

 - [X] Added a changelog entry
 
### Follow ups
- Unit tests  **DTNOR-826**
- Add inspection on 3DS return URL for error details & add corresponding analytic events **DTNOR-827**

### Authors
@scannillo

### DIFF
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
@@ -72,8 +72,7 @@ class CardClient internal constructor(
         try {
             cardAPI.fetchCachedOrRemoteClientID()
         } catch (e: PayPalSDKError) {
-            notifyApproveOrderFailure(APIClientError.clientIDNotFoundError(e.code, e.correlationID))
-            return
+            throw APIClientError.clientIDNotFoundError(e.code, e.correlationID)
         }
 
         val response = cardAPI.confirmPaymentSource(cardRequest)
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
@@ -34,7 +34,7 @@ class CardClient internal constructor(
     private val lifeCycleObserver = CardLifeCycleObserver(this)
 
     private val exceptionHandler = CoreCoroutineExceptionHandler {
-        approveOrderListener?.onApproveOrderFailure(it)
+        notifyApproveOrderFailure(it)
     }
     /**
      *  CardClient constructor
@@ -61,6 +61,8 @@ class CardClient internal constructor(
      * @param cardRequest [CardRequest] for requesting an order approval
      */
     fun approveOrder(activity: FragmentActivity, cardRequest: CardRequest) {
+        cardAPI.sendAnalyticsEvent("card-payments:3ds:started")
+
         CoroutineScope(dispatcher).launch(exceptionHandler) {
             confirmPaymentSource(activity, cardRequest)
         }
@@ -70,9 +72,7 @@ class CardClient internal constructor(
         try {
             cardAPI.fetchCachedOrRemoteClientID()
         } catch (e: PayPalSDKError) {
-            approveOrderListener?.onApproveOrderFailure(
-                APIClientError.clientIDNotFoundError(e.code, e.correlationID)
-            )
+            notifyApproveOrderFailure(APIClientError.clientIDNotFoundError(e.code, e.correlationID))
             return
         }
 
@@ -81,8 +81,9 @@ class CardClient internal constructor(
             val result = response.run {
                 CardResult(orderID, status, paymentSource)
             }
-            approveOrderListener?.onApproveOrderSuccess(result)
+            notifyApproveOrderSuccess(result)
         } else {
+            cardAPI.sendAnalyticsEvent("card-payments:3ds:confirm-payment-source:challenge-required")
             approveOrderListener?.onApproveOrderThreeDSecureWillLaunch()
 
             // launch the 3DS flow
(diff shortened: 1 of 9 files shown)


CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
```diff
@@ -73,12 +75,20 @@ class API internal constructor(
...
-        try {
-            val clientID = fetchCachedOrRemoteClientID()
-            analyticsService.sendAnalyticsEvent(name, clientID)
-        } catch (e: PayPalSDKError) {
-            Log.d("[PayPal SDK]", "Failed to send analytics due to missing clientID: ${e.message}")
+    /**
+     * Sends analytics event to https://api.paypal.com/v1/tracking/events/ via a background task.
+     */
+    fun sendAnalyticsEvent(name: String) {
+        GlobalScope.launch {
```
scannillo, line 82: Do we need `CoroutineScope(dispatcher).launch(exceptionHandler)` here instead? I figured the `GlobalScope.launch` was sufficient since we don't really care about handling the result of the failure of analytics background tasks.

demerino (approved):

CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
sshropshire: This is a good call for this PR. `GlobalScope` will spawn a background thread to execute the analytics send. We will eventually migrate to `WorkManager` it supports co-routines and handles the thread scheduling for us.

sshropshire (approved):

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -70,9 +72,7 @@ class CardClient internal constructor(
         try {
             cardAPI.fetchCachedOrRemoteClientID()
         } catch (e: PayPalSDKError) {
-            approveOrderListener?.onApproveOrderFailure(
-                APIClientError.clientIDNotFoundError(e.code, e.correlationID)
-            )
+            notifyApproveOrderFailure(APIClientError.clientIDNotFoundError(e.code, e.correlationID))
             return
```
jcnoriega, line 76: Let the error be handled by the `coroutineExcpetionHandler`:
```suggestion
            throw APIClientError.clientIDNotFoundError(e.code, e.correlationID)
```

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -70,9 +72,7 @@ class CardClient internal constructor(
         try {
             cardAPI.fetchCachedOrRemoteClientID()
         } catch (e: PayPalSDKError) {
-            approveOrderListener?.onApproveOrderFailure(
-                APIClientError.clientIDNotFoundError(e.code, e.correlationID)
-            )
+            notifyApproveOrderFailure(APIClientError.clientIDNotFoundError(e.code, e.correlationID))
```
jcnoriega, line 75: Let the error be handled by the `coroutineExcpetionHandler`:
```suggestion
            throw APIClientError.clientIDNotFoundError(e.code, e.correlationID)
```
and remove the `return` below it

jcnoriega (approved):

CorePayments/src/main/java/com/paypal/android/corepayments/API.kt
jcnoriega: [Android coroutines best practices](https://developer.android.com/kotlin/coroutines/coroutines-best-practices#global-scope) says that we should avoid `GlobalScope`, because we are "hardcoding" the `Dispatcher`, and it might be dificult to test. 
Lets see what happens when we write the unit test, we might just end up doing another `CoroutineScope` with a dispatcher injected to it

jcnoriega: we could end up with something like this:
```kotlin
    private val externalScope: CoroutineScope = GlobalScope,
    private val defaultDispatcher: CoroutineDispatcher = Dispatchers.Default
) {
    
    fun sendAnalticsEvent() {
        externalScope.launch(defaultDispatcher) {
            //send event
        }
    }
}
```

scannillo: Ok - we can 🏈 punt on in it until the unit test ticket (DTNOR-826)

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
scannillo: @jcnoriega - Hm, making this change results in a `SwallowedException` from Detekt.

scannillo: I reverted this change. I think it boils down to Detekt keeps complaining about a SwallowedException due to the way that CardAPI just wraps the API.fetchCachedOrRemoteClientID() method, which might warrant a larger refactor. Happy to address this in a follow-up PR, though.

PR #128, opened 2023-04-05
Move GetOrder from CardPayments to CorePayments

jcnoriega:

### Summary of changes

In the PayPalNativePayments, we have to get the order from ordersV2 api, to correctly patch a shipping change. But all the models, factories and parsers to perform that request, are on CardPayments. So, all logic related to that request has been moved to CorePayments. Getting the order info from the ordersV2 api is a very generic and usable call, with a high chance of being used in multiple integrations, so moving it to CorePayments seems to be the next logical step. Changes include

- Move order model from CardPayments to CorePayments (Address, PurchaseUnit, Amount, OrderIntent, etc)
- Move from `CardRequestFactory` to a new `CoreRequestFactory` all the logic to create the get order request (we can add other request creation into the factory in the future)
- Move from `CardRequestParser` to a new `CoreRequestParser` all the logic that parses the get order response. 
- Update unit tests accordingly

 ### Checklist

 - [ ] Added a changelog entry

- @jcnoriega


CardPayments/src/main/java/com/paypal/android/cardpayments/CardRequestFactory.kt
```diff
@@ -1,11 +1,11 @@
 package com.paypal.android.cardpayments
 
-import com.paypal.android.cardpayments.api.GetOrderRequest
 import com.paypal.android.corepayments.APIRequest
 import com.paypal.android.corepayments.HttpMethod
+import com.paypal.android.corepayments.api.CoreRequestFactory
 import org.json.JSONObject
 
-internal class CardRequestFactory {
+internal class CardRequestFactory : CoreRequestFactory() {
```
sshropshire, line 8: 🧶 : can we keep `createGetOrderInfoRequest` in this class and rename `CardRequestFactory` to `OrdersRequestFactory` or something? I'm thinking about [this article](https://www.cinqict.nl/blog/dont-repeat-yourself-and-dont-pay-inheritance-tax) and the section on _Inheritance Tax_.

sshropshire: Also starting to feel like request factories will become obsolete soon. That design was temporary, there's definitely a better way to keep this logic in the `CardAPI` class.

CorePayments/src/test/java/com/paypal/android/corepayments/api/CoreRequestParserUnitTest.kt
```diff
@@ -0,0 +1,177 @@
...
+                        "type":"CREDIT",
+                        "authentication_result":{
+                            "liability_shift":"NO",
+                            "three_d_secure":{
+                                "enrollment_status":"U"
+                            }
+                        }
+                    }
+                },
+                "purchase_units":[
```
scannillo, line 47: @jcnoriega - Are we OK to the 1st value in this array as the amount to act on? What if the array contains multiple amounts.

scannillo:
@jcnoriega This PR can be closed, in favor of - https://github.com/paypal/Android-SDK/pull/130 merging into `main` directly, right?

jcnoriega:
@scannillo [#130](https://github.com/paypal/Android-SDK/pull/130) already has all this changes implemented, so sure!

PR #129, opened 2023-04-07
Remove GraphQL Generics

sshropshire:
### Summary of changes

 - This PR removes generics from the `GraphQLClient` and it also removes `GraphQLRequestFactory`
 - Add additional tests to increase `EligibilityAPI` and `GraphQLClient` coverage

 ### Checklist

- @sshropshire


adr/1-remove-graphql-generics/index.md
```diff
@@ -0,0 +1,49 @@
+# Remove GraphQL Generics
```
sshropshire, line 1: 👋  Created an [ADR style](https://github.com/joelparkerhenderson/architecture-decision-record) proposal detailing the refactor [here](https://github.com/paypal/Android-SDK/blob/graphql_remove_generics/adr/1-remove-graphql-generics/index.md).

CorePayments/src/main/java/com/paypal/android/corepayments/ResourceLoader.kt
```diff
@@ -0,0 +1,29 @@
+package com.paypal.android.corepayments
+
+import android.content.Context
+import android.content.res.Resources
+import androidx.annotation.RawRes
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.io.IOException
+
+class ResourceLoader(context: Context) {
```
jcnoriega: qq: whats the benefit of putting queries in resource files? Isnt loading a resource a much more expensive operation as if the query lived in the code?

adr/1-remove-graphql-generics/index.md
jcnoriega: 🔥 🔥

CorePayments/src/main/java/com/paypal/android/corepayments/ResourceLoader.kt
sshropshire: That's a good question there's definitely some tradeoffs. I did try putting in a string, but escaping `$`'s in Kotlin felt kind of weird.

It'll be slightly more expensive, but the latency of the web request should be much larger than the latency from reading the file from disk. In total, the user may not notice. We do need access to `Context` as well which kinda sucks, but I think we'll have access to an `applicationContext` reference in most cases.

There are some schema validation plugins too that we can use to make sure the queries are valid. That'll give us some compile time schema validation that isn't available with plain strings.

jcnoriega (approved):

CorePayments/src/main/java/com/paypal/android/corepayments/ResourceLoader.kt
scannillo: Can we add docstring or rename `ResourceLoader`? At a glance of the file standlone it isn't clear what it's used for

adr/1-remove-graphql-generics/index.md
```diff
@@ -0,0 +1,49 @@
...
+
+We can also follow GraphQL best practices and [provide query variables via JSON][5]. By making each query a `raw` Android resource, we can take advantage of plugins like the [IntelliJ GraphQL Plugin][6] for compile time syntax checking and auto completion of GraphQL queries.
+
+## Consequences
+
+The proposed GraphQL refactor will make the code less DRY in some ways, but the task to create a `JSONObject` for each GraphQL API call isn't overly complex. The tradeoff is increased repetition in the codebase in exchange for more readable code that's easy to follow. A cleaner codebase will reduce the cost of onboarding new contributors.
+
+We will also gain more flexibility by removing abstraction through generics, since the API layer will now have access to the full `HttpRequest` object.
```
scannillo, line 37: 👏

scannillo (approved):
Love the included ADR!

PR #138, opened 2023-05-09
Remove API Client and HTTP Request Factories

sshropshire:
### Summary of changes

 - Remove core `API` class
 - Add `SecureTokenServiceAPI`
 - Add `TrackingEventsAPI`
 - Remove HTTP Request Factories and move request building logic into API layer

 ### Checklist

- @sshropshire


PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -21,7 +22,8 @@ import kotlinx.coroutines.launch
 class PayPalWebCheckoutClient internal constructor(
```
sshropshire, line 22: Detekt throws on this constructor because it has 7 parameters. Would y'all be opposed to lifting this restriction to 10? I'm sure it's helpful but this lint restriction feels like overkill.

gradle.properties
```diff
@@ -18,8 +18,7 @@ android.useAndroidX=true
 # Kotlin code style for this project: "official" or "obsolete":
 kotlin.code.style=official
 
-detektVersion=1.18.0-RC2
-
+detektVersion=1.22.0
```
sshropshire, line 21: Had to update Detekt version to be compatible with JDK 17.

detekt/detekt-config.yml
```diff
@@ -85,7 +85,7 @@ complexity:
     threshold: 10
     includeStaticDeclarations: false
     includePrivateDeclarations: false
-  ComplexMethod:
+  CyclomaticComplexMethod:
```
sshropshire, line 88: A lot of lint rules like this one were either deprecated or migrated to the `libraries` ruleset.

build.gradle
```diff
@@ -146,6 +146,7 @@ task clean(type: Delete) {
 
 dependencies {
     detektPlugins "io.gitlab.arturbosch.detekt:detekt-formatting:${detektVersion}"
+    detektPlugins "io.gitlab.arturbosch.detekt:detekt-rules-libraries:${detektVersion}"
```
sshropshire, line 149: Adding `libraries` ruleset since a lot of rules were migrated out of the core ruleset.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
scannillo: I'm down for 10. One of the downsides to composition in large inits and that is OK.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -45,7 +49,9 @@ class CardClient internal constructor(
     constructor(activity: FragmentActivity, configuration: CoreConfig) :
             this(
                 activity,
-                CardAPI(API(configuration, activity)),
+                OrdersAPI(configuration),
+                SecureTokenServiceAPI(configuration),
+                AnalyticsService(activity.applicationContext, configuration),
```
scannillo: 😍

CardPayments/src/main/java/com/paypal/android/cardpayments/api/OrdersAPI.kt
```diff
@@ -3,47 +3,41 @@ package com.paypal.android.cardpayments.api
...
 
     suspend fun confirmPaymentSource(cardRequest: CardRequest): ConfirmPaymentSourceResponse {
         val apiRequest = requestFactory.createConfirmPaymentSourceRequest(cardRequest)
-        val httpResponse = api.send(apiRequest)
+        val httpResponse = restClient.send(apiRequest)
 
         val error = responseParser.parseError(httpResponse)
         if (error != null) {
-            sendAnalyticsEvent("card-payments:3ds:confirm-payment-source:failed")
+//            sendAnalyticsEvent("card-payments:3ds:confirm-payment-source:failed")
```
scannillo: Looks like we still need to move these to the CardClient

CardPayments/src/test/java/com/paypal/android/cardpayments/OrdersAPIUnitTest.kt
```diff
@@ -244,7 +248,7 @@ class CardAPIUnitTest {
 
     @Test
     fun `when confirmPaymentSource is errors, correlation ID is set in Error`() = runTest {
-        coEvery { api.send(apiRequest) } returns HttpResponse(400, headers, errorBody)
+        coEvery { restClient.send(apiRequest) } returns HttpResponse(400, headers, errorBody)
```
scannillo: This is super readable too from a test perspective 👏

CorePayments/src/main/java/com/paypal/android/corepayments/RestClient.kt
```diff
@@ -1,23 +1,27 @@
...
 
-import com.paypal.android.corepayments.analytics.AnalyticsEventData
 import java.net.URL
 import java.util.Locale
 
-internal class HttpRequestFactory(private val language: String = Locale.getDefault().language) {
+class RestClient internal constructor(
+    private val configuration: CoreConfig,
+    private val http: Http = Http(),
+    private val language: String = Locale.getDefault().language
```
scannillo, line 9: Tangent - we're always setting this to `en_US` on iOS too. Worth looking into if that will restrict usage in other countries at all.

CorePayments/src/main/java/com/paypal/android/corepayments/SecureTokenServiceAPI.kt
```diff
@@ -0,0 +1,49 @@
+package com.paypal.android.corepayments
+
+import android.util.LruCache
+
+class SecureTokenServiceAPI internal constructor(
```
scannillo, line 3: Can we document that this can be found at `go/ppass` under "Secure Token Service (v1)"? just so future devs know too that these API classes are 1:1 with ppass services

CorePayments/src/test/java/com/paypal/android/corepayments/RestClientUnitTest.kt
```diff
@@ -0,0 +1,163 @@
...
+
+        sut = RestClient(sandboxConfig, http, "en_US")
+        sut.send(apiGETRequest)
+
+        val httpRequest = httpRequestSlot.captured
+        assertEquals(URL("https://api.sandbox.paypal.com/sample/path"), httpRequest.url)
+    }
+
+    @Test
+    fun `send() should properly format the url for the staging environment`() = runTest {
```
scannillo, line 56: We don't support staging on iOS. I think we should remove it from Android too. I don't think it's accessible to merchants.

scannillo: We should make a ticket for it 🏈

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsService.kt
```diff
@@ -1,34 +1,64 @@
...
+    private val secureTokenServiceAPI: SecureTokenServiceAPI,
+    private val scope: CoroutineScope
 ) {
 
-    internal suspend fun sendAnalyticsEvent(name: String, clientID: String) {
-        val timestamp = System.currentTimeMillis()
+    constructor(
+        context: Context,
+        coreConfig: CoreConfig,
+        dispatcher: CoroutineDispatcher = Dispatchers.IO
```
scannillo: Does this need to be in the public constructor if no feature clients are setting it?

sshropshire: I'll double check this. It got weird with testing lol.

sshropshire: Yeah I was able to update and make another convenience constructor to discourage changing the coroutine dispatcher.

I do have a feeling this class will change a lot though, and it won't use coroutines once we migrate to `WorkManager` and let it do the heavy lifting with thread scheduling. All the `suspend` method invocations will move into a Work Manager task.

CardPayments/src/main/java/com/paypal/android/cardpayments/api/OrdersAPI.kt
sshropshire: :nod: I'll move these.

CorePayments/src/main/java/com/paypal/android/corepayments/RestClient.kt
sshropshire: Yeah i'm not sure what affect Locale has on PayPal's web services, but thinking about it now I wouldn't be surprised if we're required to allow the merchant to specify in the future.

sshropshire:
ADR: [Remove Core API and HTTP Request Factories](https://github.com/paypal/Android-SDK/blob/dissolve_api_client_and_http_request_factories/adr/2-remove-core-api-and-http-request-factories/index.md)

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
jaxdesmarais: I am also down for bumping this to 10.

CorePayments/src/main/java/com/paypal/android/corepayments/RestClient.kt
jaxdesmarais: Yeah, we allow merchants to specify via `localeCode` on BT iOS and Android. Harder to test other locales, but we can switch this value to another locale and see if it always defaults to that even if device is US based. That will at least show us if users in other countries would always get the US based locale.

CorePayments/src/main/java/com/paypal/android/corepayments/ClientIdRepository.kt
```diff
@@ -0,0 +1,28 @@
...
+        clientIDCache.get(configuration.accessToken)?.let { cachedClientID ->
+            return cachedClientID
+        }
+        val clientID = secureTokenServiceAPI.getClientId()
+        clientIDCache.put(configuration.accessToken, clientID)
+        return clientID
+    }
+
+    companion object {
+        val clientIDCache = LruCache<String, String>(10)
```
jaxdesmarais, line 26: kotlin n00b question - does the 10 at the end set the limit for the size of this cache? Also is the cache destroyed when this class is destroyed or do we need to worry about something being held onto after it should have been destroyed?

CorePayments/src/main/java/com/paypal/android/corepayments/SecureTokenServiceAPI.kt
jaxdesmarais: ➕ 1

jaxdesmarais (approved):
Thanks for putting together the ADR for this! It really helped understand the motivation of the PR as a whole. 👏

CorePayments/src/main/java/com/paypal/android/corepayments/RestClient.kt
sshropshire: Got it ok. Would this go in the `CoreConfig`? We should be able to move this parameter into there.

CorePayments/src/main/java/com/paypal/android/corepayments/ClientIdRepository.kt
sshropshire: q1. Yup! Basically cache will keep 10 entries maximum. When size is exceeded, the entry that was least recently accessed will be evicted from the cache to make room for the newest entry.

q2. The `companion object` is basically where static properties go in Kotlin, so the cache stays in memory for as long as the application does.

jaxdesmarais: Nice - thanks for answering those questions! 😄

KunJeongPark (approved):

CorePayments/src/main/java/com/paypal/android/corepayments/RestClient.kt
sshropshire: Made a ticket: DTNOR-972

jaxdesmarais: Good question - on BT we can pass it into PayPal/MXO (though I don't actually know if MXO respects it). I assume we want to use it for everything? But actually don't know from a product perspective what the goal is, so may be good to clarify with product in the follow up ticket!

PR #137, opened 2023-05-09
Send `orderID` instead of `sessionID` to FPTI analytics

scannillo:
### Reason for changes

- Mirror [this iOS PR](https://github.com/paypal/iOS-SDK/pull/144)

### Summary of changes

- Feature clients
     - Add `AnalyticsService` instance to each feature client
         - (Since this needs an orderID, I made this a nullable param. Open to suggestions)
     - Remove clientID fetch from feature clients
 - `API`
    - Remove `sendAnalyticsEvent()` func
        - (Feature clients now interact directly with `AnalyticsService`)
    - Move fetchClientID functions into new `ClientIDAPI` layer
        - (This moves us toward eventually removing `API` in favor of `<Service>API`s per endpoint that @sshropshire proposed)
    - Remove `Context` dependency from `API`
- `APIUnitTest`
    - Move relevant tests --> `ClientIDAPIUnitTests`
    - Move relevant tests --> `AnalyticsServiceUnitTests`
- Tag changes
    - `session_id` --> `order_id`
    - `merchant_app_environment` --> `merchant_sdk_env`

Please provide feedback!

 ### Checklist

 - [x] Added a changelog entry

### Authors
@scannillo

PR #139, opened 2023-05-09
Send `orderID` instead of `sessionID` to FPTI analytics [simplified]

scannillo:
### Summary 

 - This is a simplified re-do of this PR #137
 - The difference, is this PR just has 1 job: send `order_id` for analytics via the least invasive way possible
    - The previous PR did more architecting & redesign
    - We will hold on moving analytics functionality out of `API` until this PR is merged #138

### Changes

- Send `order_id` instead of `session_id`
- Rename `merchant_app_environment` --> `merchant_sdk_env`
- Leave a TODO for refactoring the `order_id` passing logic
    - As is, it's really ugly and clunky to send `order_id` every time `sendAnalyticsEvent()` is called
    - Addressing that TODO will be easier once #138 is merged in setting the stage for our new `API` pattern

 ### Checklist

 - [X] Added a changelog entry

### Authors
@scannillo


jcnoriega (approved):

CardPayments/src/main/java/com/paypal/android/cardpayments/api/CardAPI.kt
```diff
@@ -35,15 +35,15 @@ internal class CardAPI(
 
         val error = responseParser.parseError(httpResponse)
         if (error != null) {
-            sendAnalyticsEvent("card-payments:3ds:get-order-info:failed")
+            sendAnalyticsEvent("card-payments:3ds:get-order-info:failed", getOrderRequest.orderId)
```
jaxdesmarais, line 38: Should we have a follow up to align the casing of `orderId` to `orderID`?

jaxdesmarais (approved):

CardPayments/src/main/java/com/paypal/android/cardpayments/api/CardAPI.kt
sshropshire: It's probably like that because of how the JSON property is named. I'm starting to wonder if maybe on Android we should go with `Id` as a suffix. We don't have the same restriction as iOS, and it may feel more natural for our Android developers if we use the `Id` suffix. Plus the `Id` suffix lines up naturally with the actual JSON property names. What y'all think?

scannillo: Yeh, Android should prefer `Id`. Aligns with the Java & Kotlin style guides.

Pushed! (We can make a separate PR to rename CardRequest.orderID & AnalyticsEventData.clientID)

PR #142, opened 2023-05-12
Demo App Support for Injection of Order Id and Access Token

sshropshire:
### Summary of changes

- Rename `SDKSampleMerchantServerApi` to `SDKSampleMerchantServerAPI`
- Extract Retrofit setup logic from `NetworkModule` into `SDKSampleMerchantServerAPI`
- Add injected constants for `DEFAULT_ACCESS_TOKEN` and `DEFAULT_ORDER_ID` for E2E Testing
- Stub injected values for access token and order id in `SDKSampleMerchantServerAPI` when they are set

### Checklist

### Authors
- @sshropshire
- @KunJeongPark


Demo/src/main/java/com/paypal/android/api/services/SDKSampleServerAPI.kt
```diff
@@ -0,0 +1,93 @@
...
+import retrofit2.http.POST
+import retrofit2.http.Path
+import java.util.concurrent.TimeUnit
+
+private const val CONNECT_TIMEOUT_IN_SEC = 20L
+private const val READ_TIMEOUT_IN_SEC = 30L
+private const val WRITE_TIMEOUT_IN_SEC = 30L
+
+private val DEFAULT_ACCESS_TOKEN: String? = null
+private val DEFAULT_ORDER_ID: String? = null
```
scannillo: Take it or leave it: adding a comment above these variables to specify what they're for / make it more clear to less experienced devs how to set these.

Ex:
```suggestion
// To hardcode an access token and order ID for this demo app, set the below values:
private val DEFAULT_ACCESS_TOKEN: String? = null // = "your-token"
private val DEFAULT_ORDER_ID: String? = null // = "your-order-id"
```

scannillo (approved):

Demo/src/main/java/com/paypal/android/api/services/SDKSampleServerAPI.kt
```diff
@@ -0,0 +1,94 @@
...
+
+private const val CONNECT_TIMEOUT_IN_SEC = 20L
+private const val READ_TIMEOUT_IN_SEC = 30L
+private const val WRITE_TIMEOUT_IN_SEC = 30L
+
+// To hardcode an access token and order ID for this demo app, set the below values:
+private val DEFAULT_ACCESS_TOKEN: String? = null // = "your-token"
+private val DEFAULT_ORDER_ID: String? = null // = "your-order-id"
+
+class SDKSampleServerAPI(baseUrl: String) {
```
jaxdesmarais, line 28: General question, is there a reason we do `API` over `Api` since we do id as `Id`? I would think the style guide would be the same for both.

jaxdesmarais (approved):

Demo/src/main/java/com/paypal/android/api/services/SDKSampleServerAPI.kt
sshropshire: Yeah I did have the Swift style guide [case conventions section](https://www.swift.org/documentation/api-design-guidelines/#follow-case-conventions) in mind. I do like our decision to use `ID` on iOS, mostly because it's common throughout Apple frameworks.

For Android, `Id` started to feel more natural because of common methods like [findViewById](https://developer.android.com/reference/android/view/View#findViewById(int)) that use the camel case convention.

We do legit need a shared space to put our style guides though, I'd like to iron some of these rules out. Detekt has been wildly strict too, we may be able to reframe the conversation in the context of our linting tools.

sshropshire: We do have a ticket for creating a style guide repo: DTNOR-967

jaxdesmarais: Gotcha - I am fine diverging on iOS and Android. We should first adhere to the platform specific style guides. It was more a question of if we are going with `Id` on Android (which it seems like Android leans towards) why we didn't go with `Api` here?

sshropshire: Just because of the Swift style guide. `ID` is technically an abbreviation, and I believe `id` is a reserved keyword in objc so Apple capitalizes is everywhere. That naming guide is really good, a lot of the advice in there can be applied to any language.

jaxdesmarais: Yeah, I thought that was only for Swift though and on Android we had decided to follow the [Kotlin style guide](https://developer.android.com/kotlin/style-guide#camel_case) for abbreviations? It's probably more confusing to mix the two and use the Swift style guide (`ID / API`) in some places and the Kotlin style guide (`Id / Api`) in others which is what we are doing right now. We can 🏈  on it but worth revisiting and making whichever we choose consistent across the SDK vs right now mixing the two and using both `Id` and `API`.

sshropshire: Oh wow yeah I hadn't noticed. `API` feels more readable though personally, but once we get that style guide repo up we can do pull requests into a style guide for both iOS and Android.

PR #143, opened 2023-05-15
Support Capture and Authorize in Demo App

sshropshire:
### Summary of changes

- In demo app, either capture or authorize an order based on the original intent
- Related to: https://github.com/paypal/iOS-SDK/pull/153

### Checklist

- @sshropshire
- @KunJeongPark


Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
```diff
@@ -119,31 +119,43 @@ class CardFragment : Fragment() {
         findNavController().navigate(R.id.action_cardFragment_to_testCardFragment)
     }
 
+    private fun updateStatusTextWithCardResult(result: CardResult, orderStatus: String?) {
```
scannillo: We'll want to support AUTHORIZE / CAPTURE for all payment types, not just Card. Maybe let's move this toggle to the settings, where the environment toggle lives so it can apply to all payment method demos?

sshropshire: Yeah good callout, we only did Card here. Are y'all feeling the UX of the settings menu though? To me, putting Capture / Authorize as a Drop Down option made more sense here. I feel like it makes our SDK capabilities more discoverable for people who've never used the Demo app.

jaxdesmarais (approved):
👻

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
scannillo: I don't feel too strongly about where the toggle goes. I agree though, that the settings is less obvious. 

But we def wanna support AUTH/CAPTURE for the other 2 payment methods. Can we update this PR to add it for PayPal Web & Native?

sshropshire:
@scannillo I added a group of radio buttons to each payment method demo for now. It doesn't look great, but we've talked about migrating to Jetpack Compose soon. We should be able to get some better styling down the line:

auth-capture-in-ppcp

jcnoriega (approved):

PR #146, opened 2023-05-30
Remove Client ID from project source.

sshropshire:
### Summary of changes

 - Remove Access Token integration in favor of Client ID

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire
- @KunJeongPark


CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -129,24 +117,23 @@ class CardClient internal constructor(
...
-                    val getOrderResponse = checkoutOrdersAPI.getOrderInfo(GetOrderRequest(metadata.orderID))
                     analyticsService.sendAnalyticsEvent(
                         "card-payments:3ds:get-order-info:succeeded",
                         metadata.orderID
                     )
                     val deepLinkUrl = browserSwitchResult.deepLinkUrl
-                    val result = getOrderResponse.run {
-                        CardResult(orderId, orderStatus, paymentSource, deepLinkUrl)
+                    val result = metadata.run {
+                        CardResult(orderID, null, paymentSource, deepLinkUrl)
```
scannillo: Can we leave a TODO to address this always sending `null`? We'll have to align the result type on both iOS and Android

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
```diff
@@ -2,7 +2,6 @@ package com.paypal.android.corepayments.analytics
 
 // Ref: https://blog.klipse.tech/databook/2022/06/22/separate-code-from-data.html
 data class AnalyticsEventData(
-    val clientID: String,
```
scannillo, line 5: We still want to send `clientID` in our analytics. We should be sending the value that the merchant provides. The clientID is our way of tying events to specific merchants.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # PayPal Android SDK Release Notes
 
+## unreleased
+* Breaking Changes
+  * CorePayments
+    * Access token integrations are now deprecated
```
scannillo:
```suggestion
```

Don't think that line is necessary. Also I don't think "deprecated" is appropriate since it was removed entirely.

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
sshropshire: Yup! We still send it [here](https://github.com/paypal/Android-SDK/blob/449aeabf1c1d9d2354616402bdeeeba5f75d9a7a/CorePayments/src/main/java/com/paypal/android/corepayments/TrackingEventsAPI.kt#L36).

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
sshropshire: This is a good call we should remove that property altogether. I'll remove.

sshropshire: Or actually I'm not sure what to do here. Sometimes we have it, sometimes we don't. Maybe putting "UNKNOWN" could work?

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
```diff
@@ -120,9 +120,8 @@ class CardFragment : Fragment() {
     }
 
     private suspend fun createOrder() {
-
-        val accessToken = sdkSampleServerAPI.fetchAccessToken().value
-        val configuration = CoreConfig(accessToken = accessToken)
+        val clientId = "AcXwOk3dof7NCNcriyS8RVh5q39ozvdWUF9oHPrWqfyrDS4AwVdKe32Axuk2ADo6rI_31Vv6MGgOyzRt"
```
jaxdesmarais: Should we extract this somewhere in the demo app so we don't need to define it in each fragment? I could see a case where we want to change this and need to change it in every fragment.

sshropshire: Yeah definitely. There's a sample merchant server PR that needs to get merged. Once it's in, we can switch to fetching it from the merchant server.

jaxdesmarais: Is this what we expect merchants to do? My understanding of client IDs was that they are similar to tokenization keys and can live on the client. Maybe I am misunderstanding their intent/use but I assumed we could do something similar for client ID like we do today with tokenization keys on the demo apps in Braintree.

scannillo: Yeh - client IDs can be stored on the client.

I'm not sure what is more common for merchants to do though (store on client or fetch via server). I am guessing fetch via server since I know they can swap client IDs for various products (like Old Navy v Gap v Banana Republic).

We should check with the solutions folks or NXO though to see which is more common and which we should be advocating for to our merchants in our docs/guides.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
scannillo: Hm, I'd vote to remove it over sending "Unknown"

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
sshropshire: In BT Android we store tokenization keys in [Settings.java](https://github.com/braintree/braintree_android/blob/master/Demo/src/main/java/com/braintreepayments/demo/Settings.java).

Embedding `clientId` as a `BuildConfig` variable works well for merchants with a single set of credentials. For the demo app we may have multiple client ids to test different e2e flows.

sshropshire: It'd be nice to not have this value stored in git though. The sample merchant server `GET /client_id` endpoint is the quickest route.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
sshropshire: Went ahead and removed it.

CorePayments/src/main/java/com/paypal/android/corepayments/RestClient.kt
```diff
@@ -40,8 +40,10 @@ class RestClient internal constructor(
             "Accept-Language" to language
         )
 
-        accessToken?.let { token ->
-            headers["Authorization"] = "Bearer $token"
+        if (clientId != null) {
```
jaxdesmarais: Do we want to throw an error here if the client ID is null?

jaxdesmarais (approved):

CorePayments/src/main/java/com/paypal/android/corepayments/RestClient.kt
sshropshire: This is a good call. It's non-null in `CoreConfig` so this won't happen. I'll refactor to update.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -129,24 +117,23 @@ class CardClient internal constructor(
...
-                    val getOrderResponse = checkoutOrdersAPI.getOrderInfo(GetOrderRequest(metadata.orderID))
                     analyticsService.sendAnalyticsEvent(
                         "card-payments:3ds:get-order-info:succeeded",
                         metadata.orderID
                     )
                     val deepLinkUrl = browserSwitchResult.deepLinkUrl
-                    val result = getOrderResponse.run {
-                        CardResult(orderId, orderStatus, paymentSource, deepLinkUrl)
+                    val result = metadata.run {
+                        CardResult(orderID, paymentSource, deepLinkUrl)
```
jcnoriega: isnt PaymentSource here null? After doing `confirm_payment_source`, we dont get the actual payment source, we were getting it from the `getorderinfo`. So i think that we can only return `orderID` and `deeplinkurl`

sshropshire: Yeah good call, this is `null` in the 3DS flow but not when 3DS is skipped it has a value.

I'm not sure what to do in that case. Client ID's will limit functionality going forward. We may need to reach out to the team that owns the GetOrder API to allow client-id-only access.

sshropshire: After discussion, it sounds like the right move may be to remove the `paymentSource` property from `CardResult`. Merchants can get this information from authorize and capture server-side calls.

scannillo: Agreed - [for iOS](https://github.com/paypal/iOS-SDK/blob/be9ae349d5c1dc3cdd4eab56f57f86003b126589/Sources/CardPayments/Models/CardResult.swift#L12-L16) we'll have to remove both `status` and `paymentSource`.

We should also align these types across platforms - does the merchant really need the [deepLinkURI on Android](https://github.com/paypal/Android-SDK/blob/a3b97cd5ed7d1f7c2639d98e6fb1ea44309639b2/CardPayments/src/main/java/com/paypal/android/cardpayments/model/CardResult.kt#LL13C1-L13C1)?

scannillo: Also, there might be valuable info in the redirectURL we could parse and include on the result. I know @KunJeongPark started that investigation.

sshropshire: Right yeah, if I recall we didn't parse the deep link URL when we first implemented this feature because the callback parameters weren't very uniform.

KunJeongPark: So I tried 4 different sets of test cards from this developer page: https://developer.paypal.com/docs/checkout/advanced/customize/3d-secure/test/
1) supposed to be successful authentication, cardholder enrolled, successful signature verification
2) failed signature, card holder enrolled, successful authentication, unsuccessful signature verification
3) failed authentication, cardholder enrolled, unsuccessful authentication, successful signature verification
4) not enrolled

URL returned from 3DS were not consistent for some cards. For example, the first card one time returned "sdk.ios.paypal://card/success?state=undefined&code=undefined&liability_shift=NO" and another time "sdk.ios.paypal://card/success?error=error&error_description=error&liability_shift=NO".

Results I from getOrderInfo from the cards I tested were the same => liabilityShifted: NO, threeDSecure enrollment_status: "U" and authentication_status:nil. Status back from from getOrderInfo is always "COMPLETED" on these cards.
I am assuming that these test cards are not right or something is wrong with the results that are returned from 3ds. Even if the test card is not right, I would have expected to get same url back on different trials.

Previous logic did not cover checking this url. Is assumption that if there is no error from 
3DS session that authorization is success? 

In call to `getOrderInfo` previously, we sent `CardResult` back to merchant with status of the order and `paymentSource` returned. Are we trying to accurately relay this "status" information sent back to merchant?

jaxdesmarais: We should bring the above behavior to the team that owns that API. We can also ask them what the URL response is that we can expect back for success/failure. From there we can build out logic to parse the URL and query items as @scannillo has mentioned previously and use that to determine the status or return any info we deem appropriate to a merchant. For now it probably makes sense to handle just the success/failure cases and we can add on from there.

scannillo: Dang, that's disappointing that the sandbox URLs don't have much useful info and are inconsistent between test cards. Agreed, we should reach out to whoever owns this to learn more.

I found a convo in Slack from last year where Don Roman recommends using the test card generator for potentially more accurate test cards:
> In sandbox, one can generate test cards using credit card generator. Please login to developer dashboard: https://developer.paypal.com/developer/applications and find the credit card generator on the left panel.

scannillo:
Looks like we still need to remove the GetOrderRequest.kt and GetOrderInfoResponse.kt files. Unless we're keeping it around for potential use later?

scannillo (approved):

sshropshire:
> Looks like we still need to remove the GetOrderRequest.kt and GetOrderInfoResponse.kt files. Unless we're keeping it around for potential use later?

Good call I'll delete. It's always in `git` if we ever need it back.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
sshropshire: The test card generator looks like it's in JavaScript I wonder if we could do something similar. All the test cards start with `403203` and the rest might be a random number that makes it Luhn valid.

PR #153, opened 2023-06-20
Migrate Card to Jetpack Compose

sshropshire:
### Summary of changes

 - Migrate Card Demo Fragment to Jetpack Compose

migrate-card-to-jetpack-compose

 ### Checklist

- @sshropshire


Demo/src/main/java/com/paypal/android/ui/WireframeOptionDropDown.kt
```diff
@@ -0,0 +1,48 @@
...
+import androidx.compose.material3.DropdownMenuItem
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.ExposedDropdownMenuBox
+import androidx.compose.material3.ExposedDropdownMenuDefaults
+import androidx.compose.material3.OutlinedTextField
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+
+@ExperimentalMaterial3Api
```
jaxdesmarais, line 13: General question, since this is listed as "experimental" does that mean it could change/break on us? Not a big deal since it's the demo, just curious.

sshropshire: Yeah these are required in order to compile. Compose is fairly new. This has been required for a lot of the Material3 components I've seen.

sshropshire: I am having similar difficulties with UIAutomator support that we saw when we initially tried Compose. I'm not sure when those will be resolved, but the value added by using Compose outweighs that limitation in my opinion. We need the ability to manually QA our integrations more efficiently.

Demo/src/main/java/com/paypal/android/ui/card/DateString.kt
```diff
@@ -0,0 +1,90 @@
+package com.paypal.android.ui.card
+
+import java.lang.Integer.min
+
+data class DateString(private val rawDate: String) {
```
jaxdesmarais, line 5: It'd be really cool to eventually move some of these helpers into a foundations module that can be shared across BT + PPCP + MXO!

build.gradle
```diff
@@ -8,10 +8,10 @@ buildscript {
     ]
 
     ext.versions = [
-            "kotlin"    : "1.6.21",
-            "compose"   : "1.0.2",
+            "kotlin"    : "1.8.21",
```
jaxdesmarais, line 11: Should these version bumps go into a changelog?

Demo/src/main/java/com/paypal/android/ui/card/DateString.kt
sshropshire: Def! Yeah that would be cool once we come up with some standard patterns on re-using Composables.

Demo/src/main/java/com/paypal/android/ui/WireframeOptionDropDown.kt
jaxdesmarais: Oh 110% agree, was moreso just curious as to what that portion meant and if my assumptions were correct. Agree that the advantages outweigh the disadvantages!

build.gradle
sshropshire: Yeah for sure, we should get into the habit of detailing Kotlin version changes in the CHANGELOG. There's some motion in the Kotlin library ecosystem that's been breaking apps (`jdk-7` and `jdk-8` dependencies are moving to a unified library without a suffix).

build.gradle
```diff
@@ -8,10 +8,10 @@ buildscript {
...
+            "compose"   : "1.4.7",
```
jaxdesmarais, line 12: We should likely add the compose and hilt version bumps in the changelog as well, right?

jaxdesmarais (approved):

build.gradle
sshropshire: This is a good callout! I'll say since it's only for the demo app, we can omit version changes for Compose and Hilt. We don't ship features that use compose or hilt yet, but when that day comes we definitely should add to the CHANGELOG.

Demo/src/main/java/com/paypal/android/ui/selectcard/SelectCardFragment.kt
```diff
@@ -0,0 +1,172 @@
...
+                        onTestCardSelected(card)
+                    }
+                )
+                Divider(color = Color.Black)
+            }
+        }
+    }
+
+    @ExperimentalMaterial3Api
+    @Preview
```
scannillo, line 124: How did Android Studio's preview tool work for you throughout this development? The Xcode one for SwiftUI was originally really buggy but _apparently_ is relatively stable now

sshropshire: It's ok, I'll say it's better than rebuilding every time to see changes, but that's about it. They're still really early on with Compose too, so they'll probably keep improving it over time. It definitely bugs out sometimes though.

Demo/src/main/java/com/paypal/android/ui/card/DateString.kt
scannillo: +1!

scannillo (approved):
Ran through this a few times on a Samsung device - this looks awesome! 👏 

Would definitely love a demo session about basics of constructing views in Compose and any gotchas that you encountered doing this

build.gradle
jaxdesmarais: Ah gotcha - it's hard to tell sometimes what in the build.gradle is for us vs what merchants would see/pull in 😅

sshropshire: Yeah Gradle has a convention where there's a top-level `build.gradle` and a separate `build.gradle` for each module. There's a few other ways to build a gradle project, but the way we have it is fairly standard for Android Studio projects. But it is good to callout out these details because a change in the top-level build.gradle could have an effect on the feature modules we ship.

sshropshire:
> Would definitely love a demo session about basics of constructing views in Compose and any gotchas that you encountered doing this

Totally down! A lot of the same principles apply to SwiftUI. It changes perspective a lot, we should be able to make better SDKs for our merchants by allowing it to evolve alongside Compose.

Demo/src/main/java/com/paypal/android/ui/selectcard/SelectCardFragment.kt
jaxdesmarais: During my sabbatical I used the Swift UI previews a bit and it's way more stable/easier to develop with than it was initially! It's really easy to interact with as well and go through screens all in the previews. 🤓

PR #157, opened 2023-06-28
Vault With Purchase

sshropshire:
### Summary of changes

 - Add `ApproveOrderProgressView` to detail SDK and Sample Merchant Server interaction
 - Remove unused classes:
   - `CreateOrderResponse`
   - `PreferenceConstants`
   - `TestCardGroup`
   - `LinkResponse`
   - `VaultSessionId`

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### Video

https://github.com/paypal/Android-SDK/assets/58225613/56fa50ab-4a2d-471a-a5bf-36f625cb9555


CardPayments/src/main/java/com/paypal/android/cardpayments/CardRequest.kt
```diff
@@ -1,17 +1,24 @@
...
+@Parcelize
 data class CardRequest @JvmOverloads constructor(
     val orderId: String,
     val card: Card,
     val returnUrl: String,
-    val sca: SCA = SCA.SCA_WHEN_REQUIRED
-)
+    val sca: SCA = SCA.SCA_WHEN_REQUIRED,
+    val shouldVault: Boolean = false,
+    val vaultCustomerId: String? = null
```
jaxdesmarais: I know we had initially discussed 2 approaches to vaulting:
1. add everything to the card object then do the same with other modules, so for example vaulting PayPal or Venmo would also require us to have `shouldVault` and `vaultCustomerId`
2. Have and explicit Vault object to hold these parameters (`shouldVault` and `vaultCustomerId`)

I am curious if as these objects grow it will be harder to maintain having these in the individual objects vs a shared `Vault` object that can be used across multiple modules. I'm curious what your thoughts since I know you're a lot more familiar with the underlying APIs than I am!

sshropshire: Yeah I just dropped a link to Jose's document on Vault API in the team chat.

jaxdesmarais: If we have a vault object could we remove the `shouldVault` parameter since we could assume with that object the intention was to vault? What would happen with the customer ID in the current implementation if a merchant passes a customer ID value but leaves `shouldVault` as `false`?

sshropshire: It's up to us. We can either throw or allow it and no-op (I believe the latter is the current behavior).

scannillo:
I am running this on a Samsung device and clicking on "Card" or "PayPal" from the PaymentMethodsFragment works fine for me, but clicking on "PayPal Native" results in a crash. Is anyone else experiencing this? Looks like this is the log:

```
FATAL EXCEPTION: main
Process: com.paypal.android, PID: 25932
java.lang.IllegalArgumentException: CreationExtras must have a value by `SAVED_STATE_REGISTRY_OWNER_KEY`
    at androidx.lifecycle.SavedStateHandleSupport.createSavedStateHandle(SavedStateHandleSupport.kt:91)
    at androidx.lifecycle.AbstractSavedStateViewModelFactory.create(AbstractSavedStateViewModelFactory.kt:90)
    at dagger.hilt.android.internal.lifecycle.HiltViewModelFactory.create(HiltViewModelFactory.java:116)
    at androidx.lifecycle.ViewModelProvider.get(ViewModelProvider.kt:187)
    at androidx.lifecycle.ViewModelProvider.get(ViewModelProvider.kt:153)
    at androidx.lifecycle.ViewModelLazy.getValue(ViewModelLazy.kt:53)
    at androidx.lifecycle.ViewModelLazy.getValue(ViewModelLazy.kt:35)
    at com.paypal.android.ui.paypal.PayPalNativeFragment.getViewModel(PayPalNativeFragment.kt:33)
    at com.paypal.android.ui.paypal.PayPalNativeFragment.onCreateView(PayPalNativeFragment.kt:44)
    at androidx.fragment.app.Fragment.performCreateView(Fragment.java:3104)
    at androidx.fragment.app.FragmentStateManager.createView(FragmentStateManager.java:524)
    at androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:261)
    at androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:1890)
    at androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(FragmentManager.java:1814)
    at androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:1751)
    at androidx.fragment.app.FragmentManager$5.run(FragmentManager.java:538)
    at android.os.Handler.handleCallback(Handler.java:938)
    at android.os.Handler.dispatchMessage(Handler.java:99)
    at android.os.Looper.loopOnce(Looper.java:226)
    at android.os.Looper.loop(Looper.java:313)
    at android.app.ActivityThread.main(ActivityThread.java:8751)
    at java.lang.reflect.Method.invoke(Native Method)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1135)
```

sshropshire:
> I am running this on a Samsung device and clicking on "Card" or "PayPal" from the PaymentMethodsFragment works fine for me, but clicking on "PayPal Native" results in a crash. Is anyone else experiencing this?

Yeah this is a good catch. I'm looking for a fix now. It looks like it bugs out here in [PayPalNativeFragment.kt](https://github.com/paypal/Android-SDK/blob/741325e23f487be2579bd69abfac31ca134f039b/Demo/src/main/java/com/paypal/android/ui/paypal/PayPalNativeFragment.kt#L44).

sshropshire:
@scannillo was able to fix https://stackoverflow.com/a/73675462.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardRequest.kt
jaxdesmarais: Yeah, personally I think it'd be weird to no-op if a merchant provides a customer ID but should vault as false and could see us getting inbounds around that. It sounds like the options we have are:
```swift
// option 1: provide a Vault object and default should vault to true
let cardRequest = CardRequest(
    ...
    vault: Vault() // we generate customer ID
    // OR
    vault: Vault(customerID: "my-customer-id" // merchant provided customer ID
)

// option 2: provide a Vault object and have merchants provide should vault
// should we throw if a merchant tries to pass false
let cardRequest = CardRequest(
    ...
    vault: Vault(shouldVault: true) // we generate customer ID
    // OR
    vault: Vault(shouldVault: true, customerID: "my-customer-id" // merchant provided customer ID
)

// option 3: have vault options on the request directly
let cardRequest = CardRequest(
    ...
    shouldVault: true // we generate a customer ID - is this explicit enough for merchants to know that we generate the ID
    // OR
    shouldVault: true, 
    customerID: "my-customer-id" // merchant provided customer ID - what do we do in this case if a merchant provides an ID but should vault as false
)
```
Wanted to type them out to grock what the different options look like. Personally I still lean towards a vault object since it is more explicit. Not having one makes it a bit harder to handle the what ifs and doesn't make it clear that merchants can pass should vault and have us generate a customer ID for them in the top level object.

sshropshire: Yeah I agree, they all field weird to me. At PPCP stand David mention we can go with Option 3 for now and circle back once we get some merchant feedback.

jaxdesmarais: Ah, gotcha, I didn't realize that was the decision I thought we wanted to mock them out a bit more and discuss. I think this is fine for now then and we can revisit later.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
   * Bump Kotlin version to `1.8.21`
 * CardPayments
   * Make `Card` implement `Parcelable`
+  * Make `CardRequest` implement `Parcelable`
```
jaxdesmarais, line 8: Do we also need to add an entry for the new vault parameters to the changelog?

CardPayments/src/main/java/com/paypal/android/cardpayments/CardRequest.kt
```diff
@@ -1,17 +1,24 @@
...
 import com.paypal.android.cardpayments.threedsecure.SCA
+import kotlinx.parcelize.Parcelize
 
 /**
  * A card request to process a card payment
  * @property orderId The order ID to to process the payment
  * @property card   Card used for payment
  * @property returnUrl Url to return to app after SCA challenge finishes
  * @property sca Specify to always launch 3DS or only when required. Defaults to `SCA.SCA_WHEN_REQUIRED`.
+ * @property shouldVault Set this value to true if the payment method should be vaulted on success
```
jaxdesmarais: We should specify in this doc string that when set to true and a customer ID is not provided we will generate one for the merchant

CardPayments/src/main/java/com/paypal/android/cardpayments/CardRequest.kt
```diff
@@ -1,17 +1,24 @@
...
+ * @property vaultCustomerId Set a customer id for vaulting payment methods when [shouldVault] is true
```
jaxdesmarais: We should call out that this is optional maybe just something like:
```suggestion
 * @property vaultCustomerId Optionally set a customer id for vaulting payment methods when [shouldVault] is true
```

CHANGELOG.md
sshropshire: Yeah def that's a good call. I'll add those in.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardRequest.kt
sshropshire: 👍 yup! I'll reformat this to make it more clear.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardRequest.kt
sshropshire: Ended up combining the last to feedback items into this:

> @property vaultCustomerId Optional. When [shouldVault] is set to true and this property is set, the
 value of this property will be used to associate the vaulting of a payment method with a specific
 customer. If [shouldVault] is set to true and this property is not set, a vault customer ID will be generated.

scannillo:
> @scannillo was able to fix https://stackoverflow.com/a/73675462.

Good find! This works for me.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardRequest.kt
scannillo: I also lean toward a vault object. Something like:

```
let cardRequest = CardRequest(
    ...
    useVault: Vault() // we generate customer ID
    // OR
    useVault: Vault(customerID: "my-customer-id") // merchant provided customer ID
)
```

sshropshire: Yeah we made the decision to go option 3 the other day in stand. I'm down to bring it up again in the team channel, we should get David involved if we decide to switch it up.

Demo/src/main/java/com/paypal/android/ui/approveorderprogress/ApproveOrderProgressFragment.kt
```diff
@@ -0,0 +1,188 @@
...
+        cardClient.approveOrder(requireActivity(), cardRequest)
+    }
+
+    @ExperimentalMaterial3Api
+    @Composable
+    fun ApproveOrderProgressView(events: List<ComposableEvent>) {
+        val listState = rememberLazyListState()
+        LaunchedEffect(events) {
+            // continuously scroll to bottom of the list when event state is updated
+            listState.animateScrollToItem(events.size)
```
scannillo, line 134: This entire scroll view looks awesome!

Demo/src/main/java/com/paypal/android/uishared/components/CardResultView.kt
```diff
@@ -0,0 +1,88 @@
...
+                UriView(uri = result.deepLinkUrl!!)
+            }
+            Spacer(modifier = Modifier.size(24.dp))
+        }
+    }
+}
+
+@Preview
+@Composable
+fun CardResultViewWithDeepLinkPreview() {
```
scannillo, line 65: Would love a demo or mob on how you've been using these Previews!

scannillo:
Do we want to add vaulting documentation to our GH hosted docs ([here](https://github.com/paypal/Android-SDK/tree/main/docs/CardPayments)) as part of this PR?

CardPayments/src/main/java/com/paypal/android/cardpayments/CardRequest.kt
jaxdesmarais: I think we should circle back on it with David - I can post in there

scannillo: If we already decided on 3 we can stick with that! I know sometimes PPCP can get stuck in a lot of back and forth.

scannillo (approved):
Changes look good to me! Left a comment about the docs that need to be addressed in this repo. Whether that happens in this PR or a follow-up.

KunJeongPark (approved):

CardPayments/src/main/java/com/paypal/android/cardpayments/CardRequest.kt
sshropshire: Sounds like option 1 is the route. I'm going to change it to that one.

Demo/src/main/java/com/paypal/android/uishared/components/CardResultView.kt
sshropshire: Most def!

sshropshire:
> Do we want to add vaulting documentation to our GH hosted docs ([here](https://github.com/paypal/Android-SDK/tree/main/docs/CardPayments)) as part of this PR?

@scannillo for this, I'd say we can do the docs in a separate PR so we can capture Vault with Purchase and Vault without Purchase in the same docs story.

PR #159, opened 2023-07-05
Support Multiple Sample Server Merchants 

sshropshire:
### Summary of changes

 - Add `Merchant` enum to Demo app for supported merchants:
   - `DEFAULT`
   - `DIRECT`
   - `CONNECTED_PATH`
   - `MANAGED_PATH`
- Allow `Merchant` to be specified when making Sample Merchant Server API calls

 ### Checklist

- @sshropshire
- @KunJeongPark


Demo/src/main/java/com/paypal/android/api/services/SDKSampleServerAPI.kt
```diff
@@ -26,7 +26,12 @@ private const val WRITE_TIMEOUT_IN_SEC = 30L
 private val DEFAULT_CLIENT_ID: String? = null // = "your-client-id"
 private val DEFAULT_ORDER_ID: String? = null // = "your-order-id"
 
-class SDKSampleServerAPI(baseUrl: String) {
+class SDKSampleServerAPI {
+
+    companion object {
+        // UPCOMING PR: require Merchant enum to be specified via UI layer
```
scannillo: Nitpick so this comment is searchable

```suggestion
        // TODO: - require Merchant enum to be specified via UI layer
```

Demo/src/main/java/com/paypal/android/api/services/SDKSampleServerAPI.kt
```diff
@@ -56,10 +61,18 @@ class SDKSampleServerAPI(baseUrl: String) {
         suspend fun authorizeOrder(@Path("orderId") orderId: String): Order
     }
 
-    private val service: RetrofitService
+    private val serviceMap: Map<Merchant, RetrofitService>
 
     init {
+        val serviceMap = mutableMapOf<Merchant, RetrofitService>()
```
scannillo: What does `serviceMap` mean? Can this be something more descriptive

Demo/src/main/java/com/paypal/android/api/services/Merchant.kt
```diff
@@ -0,0 +1,8 @@
+package com.paypal.android.api.services
+
+enum class Merchant(val baseUrl: String) {
```
scannillo: Another nitpick on naming - just because `Merchant` is vague IMO

```suggestion
enum class MerchantAccountType(val baseUrl: String) {
```

Demo/src/main/java/com/paypal/android/api/services/SDKSampleServerAPI.kt
sshropshire: The equivalent in Swift would be `serviceDictionary`. The lookup key is the `Merchant` enum, and the value is a Retrofit instance that makes HTTP requests using the merchant's associated `baseUrl`.

The sample merchant server has a base route for each supported integration pattern e.g. Direct, Connected Path, Managed path. This way we can support multiple integration methods using a single server.

Demo/src/main/java/com/paypal/android/api/services/Merchant.kt
sshropshire: Yeah that makes sense. We may go with `MerchantIntegration` to avoid baking `Type` into the name. I like this snippet from [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/#name-according-to-roles). We aren't in Swift, but that naming guideline applies to a lot of languages I feel.

Demo/src/main/java/com/paypal/android/api/services/SDKSampleServerAPI.kt
sshropshire: Service is also a concept taken from the [Retrofit docs](https://square.github.io/retrofit/). We could call `RetrofitService` something else, but `SampleMerchantServerService` doesn't feel right even though it's technically what it is. Here's the whole Retrofit integration in [SDKSamplerServerAPI.kt](https://github.com/paypal/Android-SDK/blob/c2ebc91f0abc68174b6236a1ce9b6a17ad9737e7/Demo/src/main/java/com/paypal/android/api/services/SDKSampleServerAPI.kt#L37).

scannillo (approved):

KunJeongPark (approved):

PR #167, opened 2023-07-21
Vault Without Purchase UI Placeholder

sshropshire:
### Summary of changes

 - PR to Add Vault Without Purchase to SDK and Demo app
 - The Vault Without Purchase implementation will follow in a separate PR
 - Rename `PaymentMethodsFragment` to `FeaturesFragment`

https://github.com/paypal/paypal-android/assets/58225613/5adc10c6-22fa-473b-b3aa-2852dd0ed135

 ### Checklist

- @sshropshire


CardPayments/src/main/java/com/paypal/android/cardpayments/VaultListener.kt
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Listener to receive vault results.
+ */
+interface VaultListener {
+
+    /**
+     * Called after a successful vault occurs.
+     */
+    @MainThread
+    fun onVaultSuccess(result: VaultResult)
```
scannillo: Are we planning to add more cases, like if the attaching of the payment source to the setup token fails? Or if the user cancels the 3DS screen?

(looking at Jose's [doc](https://hackmd.io/2SUQKJh5TyGY_IozR4QUMQ?view#Kotlin))

CardPayments/src/main/java/com/paypal/android/cardpayments/VaultRequest.kt
```diff
@@ -0,0 +1,16 @@
+package com.paypal.android.cardpayments
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+/**
+ * @property card   Card used for payment
+ * @property returnUrl Url to return to app after SCA challenge finishes
```
scannillo: Nitpick - specify this is url scheme formatting

CardPayments/src/main/java/com/paypal/android/cardpayments/VaultListener.kt
sshropshire: Yup! We hit a snag though, vault without purchase is more involved than we anticipated. At the moment, most of the request/response classes are placeholders. The bulk of this PR is UI / UX cleanup to get the vault feature flow into the demo app.

KunJeongPark: Why do we need the place holders?

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -72,6 +76,18 @@ class CardClient internal constructor(
         }
     }
 
+    /**
+     * Initiate vaulting for a Card payment method.
+     *
+     * NOTE: 3DS Challenge will occur when required.
+     */
+    fun vault(vaultRequest: VaultRequest) {
```
scannillo: Since we have merchants using the SDK now, can we make the vault w/o purchase flow (for now) internal only until it's fully functional?

Similar question for the vault w/ purchase flow. Is that ready to be public? I know we still need to add to our `docs/` directory for that flow.

CardPayments/src/main/java/com/paypal/android/cardpayments/VaultListener.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import androidx.annotation.MainThread
+
+/**
+ * Listener to receive vault results.
+ */
+interface VaultListener {
+
+    /**
+     * Called after a successful vault occurs.
```
jaxdesmarais:
```suggestion
     * Called after vaulting is successful.
```

CardPayments/src/main/java/com/paypal/android/cardpayments/VaultRequest.kt
```diff
@@ -0,0 +1,16 @@
...
+ * @property customerId Optional customer Id to associate when vaulting
```
jaxdesmarais:
```suggestion
 * @property customerId Optional customer Id to associate with vaulting
```

CardPayments/src/main/java/com/paypal/android/cardpayments/api/DataVaultPaymentMethodTokensAPI.kt
```diff
@@ -0,0 +1,18 @@
+package com.paypal.android.cardpayments.api
+
+import com.paypal.android.cardpayments.VaultRequest
+import com.paypal.android.cardpayments.VaultResult
+import com.paypal.android.corepayments.CoreConfig
+import com.paypal.android.corepayments.RestClient
+
+// TODO: remove suppress statement when implementation is done
```
jaxdesmarais: Agree with @scannillo statement, we should keep this feature internal as a whole until it's ready

Demo/src/main/java/com/paypal/android/api/model/PaymentToken.kt
```diff
@@ -0,0 +1,8 @@
+package com.paypal.android.api.model
+
+data class PaymentToken(
+    val id: String,
```
jaxdesmarais: General Q, is this ID similar to the payment method token ID on BT? And if so can we call it payment token ID here? I know it's the demo app but may reduce confusion.

Demo/src/main/java/com/paypal/android/ui/card/CardFragment.kt
```diff
@@ -82,34 +81,78 @@ class CardFragment : Fragment() {
...
             setContent {
                 MaterialTheme {
                     Surface(modifier = Modifier.fillMaxSize()) {
+                        val feature = args.feature
                         val uiState by viewModel.uiState.collectAsStateWithLifecycle()
-                        CardView(uiState = uiState, onFormSubmit = { approveOrder() })
+                        CardView(
+                            feature = feature,
+                            uiState = uiState,
+                            onFormSubmit = { onFormSubmit() })
```
jaxdesmarais: nit based on other things in this file, but should the ending `)` be on a new line?

sshropshire: This is a good call. I'll fix after deleting all the unnecessary files in the mob review.

jaxdesmarais: I think the new line end parentheses still needs to get updated here

jaxdesmarais (approved):

KunJeongPark (approved):

PR #185, opened 2023-08-31
Show Test Cards as Modal Popup

sshropshire:
### Summary of changes

 - Refactor Demo app to show Test Cards UI as modal popup inline with Approve Order and Vault without Purchase flows

https://github.com/paypal/paypal-android/assets/58225613/1e11aa15-05e5-4b79-8c8d-1ff559824f3d

 ### Checklist

- @sshropshire

### DIFF
--- a/Demo/src/main/java/com/paypal/android/utils/BundleParcelableExt.kt
+++ b/Demo/src/main/java/com/paypal/android/utils/BundleParcelableExt.kt
@@ -1,0 +1,11 @@
+package com.paypal.android.utils
+
+import android.os.Build.VERSION.SDK_INT
+import android.os.Bundle
+import android.os.Parcelable
+
+// Ref: https://stackoverflow.com/a/73311814
+inline fun <reified T : Parcelable> Bundle.parcelable(key: String): T? = when {
+    SDK_INT >= 33 -> getParcelable(key, T::class.java)
+    else -> @Suppress("DEPRECATION") getParcelable(key) as? T
+}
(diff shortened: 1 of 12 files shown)


Demo/src/main/java/com/paypal/android/utils/BundleParcelableExt.kt
```diff
@@ -0,0 +1,12 @@
+package com.paypal.android.utils
+
+import android.os.Build
+import android.os.Build.VERSION.SDK_INT
+import android.os.Bundle
+import android.os.Parcelable
+
+// Ref: https://stackoverflow.com/a/73311814
+inline fun <reified T : Parcelable> Bundle.parcelable(key: String): T? = when {
```
jaxdesmarais, line 9: Any idea why the OG of this was deprecated? Just curious more than anything else.

jaxdesmarais (approved):

Demo/src/main/java/com/paypal/android/utils/BundleParcelableExt.kt
sshropshire: Yeah i'm not sure honestly. My guess would be improved type safety since the class needs to be declared explicitly rather than having to cast the value at runtime.

jaxdesmarais: Gotcha - that makes sense

scannillo (approved):

PR #190, opened 2023-09-12
Cleanup before GA Release

sshropshire:
### Summary of changes

* This PR Contains updates to reduce the surface area of the SDK before GA release
* Overall there are more changes in this PR than are noted in the CHANGELOG
  * I included changes that I felt were the most "merchant facing"
  * I excluded changes that I felt were internal to the SDK
  * I'd be up for adding more detail if needed

#### Breaking Changes
* CardPayments
  * Make `Amount` class internal
  * Make `ApproveOrderMetadata` class internal
  * Make `AuthenticationResult` class internal
  * Make `Payee` class internal
  * Make `PaymentSource` class internal
  * Make `PurchaseUnit` class internal
  * Make `ThreeDSecureResult` class internal
  * Make `Environment` enum associated values internal
  * Remove `OrderRequest` class
* CorePayments
  * Remove `open` modifier on `PayPalSDKError`

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


jaxdesmarais (approved):
🥳 love reducing access modifiers where we can - it'll make the merchant story so much easier!

CorePayments/src/main/java/com/paypal/android/corepayments/PayPalSDKError.kt
```diff
@@ -1,9 +1,11 @@
 package com.paypal.android.corepayments
 
+import androidx.annotation.RestrictTo
+
 /**
  * Class used to describe PayPal errors when they occur.
  */
```
sarahkoop, line 7: Should these docs be suppressed?

sarahkoop (approved):

CorePayments/src/main/java/com/paypal/android/corepayments/PayPalSDKError.kt
sshropshire: We actually still need to generate docs for PPCP using dokka, so for now it should be fine. I'm hoping that these get rendered as class level docs. It may be difficult to fully suppress docs for the library-private `constructor` because of the Kotlin syntax.

PR #199, opened 2023-10-13
Nxo auth config

kelay858:

### Summary of changes

 - Creating the auth config for android for merchant passed email

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/PayPalNativePayments/src/main/java/com/paypal/android/paypalnativepayments/PayPalNativeCheckoutRequest.kt
+++ b/PayPalNativePayments/src/main/java/com/paypal/android/paypalnativepayments/PayPalNativeCheckoutRequest.kt
@@ -7,5 +7,20 @@ data class PayPalNativeCheckoutRequest(
     /**
      * The order ID associated with the request.
      */
-    val orderId: String
+    val orderId: String,
+
+    /**
+     * Optional: The setup configuration for authentication
+     */
+    val authConfig: AuthConfig? = null
+)
+
+/**
+ * Used for setting up authentication configurations
+ */
+data class AuthConfig(
+    /**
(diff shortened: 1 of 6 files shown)


PayPalNativePayments/src/main/java/com/paypal/android/paypalnativepayments/PayPalNativeCheckoutRequest.kt
```diff
@@ -7,5 +7,20 @@ data class PayPalNativeCheckoutRequest(
     /**
      * The order ID associated with the request.
      */
-    val orderId: String
+    val orderId: String,
+
+    /**
+     * Optional: The setup configuration for authentication
+     */
+    val authConfig: AuthConfig? = null
```
sshropshire, line 15:
```suggestion
    val email: String? = null
```
nit: would it be cleaner to just have a `val email: String?` property ?

scannillo: ➕1 here

tdchow: I believe Kevin separated the `AuthConfig ` data class in the Checkout repo because there are a bunch of other properties on the Config object. I can update this to add the email field directly to request object here.

tdchow:
I don't have write access to Kevin's fork. We can wait until he's back from PTO to push the changes. 

If we end up releasing before he gets back, I'll open a separate PR for these changes. Thanks!

jaxdesmarais:
Closing as duplicate of #216

PR #201, opened 2023-10-16
[GitHub Actions] Create Workflow to Deploy Demo App to Google Play

sshropshire:
### Summary of changes

 - Use the [Triple-T/gradle-play-publisher ](https://github.com/Triple-T/gradle-play-publisher)3rd-party plugin to enable push button deployments of the Demo app to Google Play
 - Create GitHub Action to deploy to Google Play manually

 ### Checklist

- @sshropshire


Demo/build.gradle
```diff
@@ -16,12 +17,11 @@ android {
         applicationId "com.paypal.android"
         minSdkVersion 21
         targetSdkVersion 33
-        versionCode 1
+        versionCode 2
```
sarahkoop: Should there be a step in this workflow to bump this? It will fail if it's not a higher number than previous version right?

scannillo:
Is there a site similar to TestFlight where we can go to see that a build was properly published? Also how do we prevent the build from hitting the actual Play Store?

Demo/build.gradle
sshropshire: Yup! I was hoping to do a couple YOLO commits to verify that GitHub Actions before adding version code bump automation. I've been able to deploy from the command line on my local workstation so far.

sshropshire:
> Is there a site similar to TestFlight where we can go to see that a build was properly published? Also how do we prevent the build from hitting the actual Play Store?

Yeah! The Google Play Console will be the place. I'm gonna add a docs guide internally once everything's wrapped up.

Demo/build.gradle
```diff
@@ -61,6 +61,31 @@ android {
...
         }
     }
+
+    // https://developer.android.com/build/build-variants#groovy
+    signingConfigs {
+        release {
+            storeFile file(System.getenv('DEMO_KEYSTORE_FILE') ?: 'debug.keystore')
+            storePassword System.getenv('DEMO_KEYSTORE_PASSWORD') ?: 'android'
+            keyAlias System.getenv('DEMO_KEY_ALIAS') ?: 'androiddebugkey'
+            keyPassword System.getenv('DEMO_KEY_PASSWORD') ?: 'android'
```
scannillo, line 71: Do these variables need to be set locally in order to do a command-line deploy?

sarahkoop (approved):

Demo/build.gradle
sshropshire: They do need to be set yes, but they'll be set in the CI environment. I did some manual deploys early on but it should all be automated going forward.

Demo/build.gradle
sshropshire: Ended up updating the script to do this automatically 👍.

FraudProtection/build.gradle
```diff
@@ -9,8 +9,6 @@ android {
     defaultConfig {
         minSdkVersion modules.androidMinSdkVersion
         targetSdkVersion modules.androidTargetVersion
-        versionCode modules.sdkVersionCode
-        versionName modules.sdkVersionName
```
sshropshire, line 13: Removed these properties form library modules since they no longer apply ([reference](https://developer.android.com/build/releases/past-releases/agp-4-1-0-release-notes)).

scannillo (approved):

PR #204, opened 2023-11-07
Make `DateString` parsing more DRY

sshropshire:
### Summary of changes

 - This PR cleans up some duplicate logic related to Expiration Date auto formatting
 - The `DateVisualTransformation` code can still use some improvement, but it's best to defer a more in-depth refactor for now since we don't actually ship an Expiration Date UI component in the SDK
 - The main goal is to clean up this piece of code before migrating away from Android Fragments in the Demo app

 ### Checklist

- @sshropshire

### DIFF
--- a/Demo/src/main/java/com/paypal/android/ui/card/DateString.kt
+++ b/Demo/src/main/java/com/paypal/android/ui/card/DateString.kt
@@ -70,12 +70,9 @@ data class DateString(private val rawDate: String) {
             formattedMonth = dateStringComponents[0]
             if (dateStringComponents.size > 1) {
                 val rawYear = dateStringComponents[1]
-                formattedYear = if (rawYear.length == 2) {
-                    // pad with 20 to assume 2000's
-                    "20$rawYear"
-                } else {
-                    rawYear
-                }
+
+                // assume date in 2000's and pad with "20" (if necessary)
+                formattedYear = if (rawYear.length == 2) "20$rawYear" else rawYear
             }
         }
 
--- a/Demo/src/main/java/com/paypal/android/ui/card/DateString.kt
+++ b/Demo/src/main/java/com/paypal/android/ui/card/DateString.kt
@@ -1,5 +1,6 @@
 package com.paypal.android.ui.card
 
+import okhttp3.internal.format
 import java.lang.Integer.min
 
 data class DateString(private val rawDate: String) {
@@ -11,6 +12,8 @@ data class DateString(private val rawDate: String) {
     private val didAddSlash: Boolean
 
     val formatted: String
+    val formattedMonth: String
+    val formattedYear: String
 
     companion object {
         private const val maxMonthLength = 2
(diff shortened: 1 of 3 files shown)


Demo/src/main/java/com/paypal/android/ui/card/DateString.kt
```diff
@@ -58,6 +60,27 @@ data class DateString(private val rawDate: String) {
...
         this.didAddSlash = addSlash
         this.formatted = formatted
+
+        // TODO: handle invalid date string
+        var formattedMonth = ""
+        var formattedYear = ""
+
+        val dateStringComponents = formatted.split("/")
+        if (dateStringComponents.isNotEmpty()) {
+            formattedMonth = dateStringComponents[0]
```
jaxdesmarais, line 70: Take it or leave it: do we care about handling months similar to year where if you enter a single digit month like `1` for January padding it with a leading `0`? Not sure if the API handles this or if we'd want to.

Demo/src/main/java/com/paypal/android/ui/card/DateString.kt
```diff
@@ -58,6 +60,27 @@ data class DateString(private val rawDate: String) {
...
+            if (dateStringComponents.size > 1) {
+                val rawYear = dateStringComponents[1]
+                formattedYear = if (rawYear.length == 2) {
+                    // pad with 20 to assume 2000's
+                    "20$rawYear"
+                } else {
+                    rawYear
+                }
```
jaxdesmarais: Also take it or leave it: seems like we could single line this which is close to the ternary operator in Swift. Bummer though it's not quite as clean as the Swift counterpart ([Kotlin docs here](https://kotlinlang.org/docs/control-flow.html#if-expression)).

jaxdesmarais (approved):

jwarmkessel (approved):
LGTM: Description of this PR to do a more comprehensive refactor of the code later.

Demo/src/main/java/com/paypal/android/ui/card/DateString.kt
sshropshire: Yeah we do pad it in this class it's a couple lines up. It's just weird because of the way Compose works we're using a class called a `VisualTransformation` to turn `"125"` into `"01/25"` when we're rendering the text, but the actual value of the text field in this case will be `"125"`.

sshropshire: Here's a link to [the padding part](https://github.com/paypal/paypal-android/blob/736e42b473c41379e3c6ea9aeed2412f7ae34132/Demo/src/main/java/com/paypal/android/ui/card/DateString.kt#L47).

Demo/src/main/java/com/paypal/android/ui/card/DateString.kt
sshropshire: Yeah that's a good call. I went ahead and flattened it out here: [1b5a4ae](https://github.com/paypal/paypal-android/pull/204/commits/1b5a4aeee93f860b832568904e79a9c7aff9536f)

Demo/src/main/java/com/paypal/android/ui/card/DateString.kt
jaxdesmarais: Oh dope! Didn't see it trapped in that non-expanded part.

PR #206, opened 2023-11-08
Reverse Publish Demo Workflow Order of Operations

sshropshire:
### Summary of changes

 - The bump of the version code number was not being updated in the release artifact
 - It appears to be a [general issue](https://github.com/actions/checkout/issues/439) with the GitHub Checkout action
 - Downstream jobs don't receive the recently pushed commit with the bumped version number
 - We can work around this by pushing the version code changes after a release to Google Play (similar to how we do for releases to maven central). 

 ### Checklist

- @sshropshire


jaxdesmarais (approved):
👻

scannillo (approved):

PR #218, opened 2023-11-21
Migrate from JetPack Navigation to Compose Navigation

sshropshire:
### Summary of changes

 - Migrating to Compose Navigation allows us to move the Demo app to 100% usage of Compose UI
 - Every demoable feature's associated `Fragment` instance has been deleted
 - Demo App `compileSdkVersion` and `targetSdkVersion` is now API Level 34
 - The SDK target API versions are unchanged, but we should consider bumping these up in a future release
 - `DemoActivity` has been renamed to `MainActivity`

### Screenshots

https://github.com/paypal/paypal-android/assets/58225613/656a541b-d6e9-49c6-a695-acbaef826d66

 ### Checklist

- @sshropshire


Demo/src/main/java/com/paypal/android/MainActivity.kt
```diff
@@ -0,0 +1,30 @@
...
+        super.onCreate(savedInstanceState)
+        setContent {
+            DemoApp()
+        }
+    }
+
+    override fun onNewIntent(newIntent: Intent) {
+        super.onNewIntent(newIntent)
+        intent = newIntent
+        Log.e("DEMO ACTIVITY", "ON NEW INTENT")
```
sarahkoop: Do we need this log? Is it helpful?

sarahkoop (approved):

Demo/src/main/java/com/paypal/android/MainActivity.kt
sshropshire: That's a good call, I'll remove this log.

jaxdesmarais (approved):

PR #220, opened 2023-12-04
Paypal Web Vault Without Purchase

sshropshire:
This PR adds PayPal Vault Without Purchase to the Demo app and the `PayPalWebPaments` SDK feature module.

### Summary of changes

- PayPalWebPayments
  - Add `PayPalWebCheckoutVaultExperienceContext` data class
  - Add `PayPalWebCheckoutVaultListener` interface
  - Add `PayPalWebCheckoutVaultResult` data class
  - Add `experienceContext` property to `PayPalWebCheckoutClient`
  - Add `vaultListener` property to `PayPalWebCheckoutClient`
  - Add `vault()` method to `PayPalWebCheckoutClient`

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


Demo/src/main/java/com/paypal/android/api/model/CardPaymentToken.kt
```diff
@@ -1,6 +1,6 @@
 package com.paypal.android.api.model
 
-data class PaymentToken(
+data class CardPaymentToken(
```
sshropshire, line 3: Ended up creating two token types in the Demo app: `CardPaymentToken` and `PayPalPaymentToken`. They are slightly different. Could have subclassed here, but it's more flexible to keep our data hierarchy flat in the Demo app to make change easier in the future.

Demo/src/main/java/com/paypal/android/api/model/CardSetupToken.kt
```diff
@@ -0,0 +1,7 @@
+package com.paypal.android.api.model
+
+data class CardSetupToken(
```
sshropshire, line 3: Same here, we may want to revisit in the future to keep code more DRY, but separate `CardSetupToken` and `PayPalSetupToken` types for now.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -68,6 +88,17 @@ class PayPalWebCheckoutClient internal constructor(
             }
         }
 
+    var vaultListener: PayPalWebCheckoutVaultListener? = null
```
sshropshire: Added a separate listener for PayPal Vault Without Purchase callbacks.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -96,11 +127,28 @@ class PayPalWebCheckoutClient internal constructor(
 
     internal fun handleBrowserSwitchResult() {
         browserSwitchResult = browserSwitchClient.deliverResult(activity)
-        listener?.also {
-            browserSwitchResult?.also { result ->
-                when (result.status) {
-                    BrowserSwitchStatus.SUCCESS -> deliverSuccess()
-                    BrowserSwitchStatus.CANCELED -> deliverCancellation()
+        if (listener == null && vaultListener == null) {
```
sshropshire: Preserving functionality here, but the original logic may need to be re-visited. We should also move to [manual browser switching](https://github.com/braintree/braintree_android/blob/main/v4.9.0%2B_MIGRATION_GUIDE.md#manual-browser-switching-for-browser-based-flows) in a next major version to align with `braintree_android` and prevent logic like this from being necessary.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -143,4 +219,12 @@ class PayPalWebCheckoutClient internal constructor(
         analyticsService.sendAnalyticsEvent("paypal-web-payments:succeeded", orderId)
         listener?.onPayPalWebSuccess(result)
     }
+
+    fun vault(activity: AppCompatActivity, setupTokenId: String, approveVaultHref: String) {
```
sshropshire: The name `vault()` doesn't have to be final here if we can find a better name for starting the vault flow. Also wondering if `setupTokenId` and `approveVaultHref` should be placed in a `PayPalWebCheckoutVaultRequest` parameter object.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutVaultExperienceContext.kt
```diff
@@ -0,0 +1,16 @@
+package com.paypal.android.paypalwebpayments
+
+data class PayPalWebCheckoutVaultExperienceContext internal constructor(
```
sshropshire: Adding this `PayPalWebCheckoutVaultExperienceContext` type to make constructing the Create Setup Token call easier for merchants. Having the SDK control the format of the return and cancel URLs can also help us enforce an internal deep linking schema for parsing browser switch results in a predictable manner.

Demo/src/main/java/com/paypal/android/ui/paypalwebvault/PayPalWebVaultView.kt
```diff
@@ -0,0 +1,137 @@
...
+            .padding(8.dp)
+            .verticalScroll(scrollState)
+    ) {
+        CreateSetupTokenForm(
+            isLoading = uiState.isCreateSetupTokenLoading,
+            customerId = uiState.vaultCustomerId,
+            onCustomerIdValueChange = { value -> viewModel.vaultCustomerId = value },
+            onSubmit = {
+                context.getActivity()?.let { activity ->
+                    viewModel.createSetupToken(activity)
```
sarahkoop: I know it's required by our API currently, but passing activity to the viewmodel is an anti-pattern I think - would there be a logical way to setup the demo app that doesn't require passing activity to viewmodel?

Demo/src/main/java/com/paypal/android/usecase/CreatePayPalPaymentTokenUseCase.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.json.JSONObject
+import javax.inject.Inject
+
+class CreatePayPalPaymentTokenUseCase @Inject constructor(
+    private val sdkSampleServerAPI: SDKSampleServerAPI
+) {
+
+    suspend operator fun invoke(setupToken: PayPalSetupToken): PayPalPaymentToken {
+        // language=JSON
+        val request = """
```
sarahkoop, line 17: Maybe a utils class for all of these JSON objects? Might be helpful for merchants to see the difference side by side (for Card vs PayPal)

Demo/src/main/java/com/paypal/android/usecase/CreatePayPalSetupTokenUseCase.kt
```diff
@@ -0,0 +1,59 @@
...
+        val customerJSON = responseJSON.getJSONObject("customer")
+
+        val linksJSON = responseJSON.optJSONArray("links") ?: JSONArray()
+        var approveVaultHref: String? = null
+        for (i in 0 until linksJSON.length()) {
+            val link = linksJSON.getJSONObject(i)
+            if (link.getString("rel") == "approve") {
+                approveVaultHref = link.getString("href")
+                break
+            }
```
sarahkoop: Want to make sure I understand - all this JSON building/parsing is something a merchant will have to do? It seems pretty complex - can we provide a helper for this?

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
sarahkoop: Would it be better to just use `listener?.let { /* invoke listener */ }` whenever we invoke the listeners throughout? Is there a chance with this logic that they've set the `listener` but not the `vaultListener` and would get an NPE if we try to invoke the `vaultLlistener` later on?

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutVaultExperienceContext.kt
sarahkoop: Can we add a doc string for this?

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutVaultListener.kt
```diff
@@ -0,0 +1,26 @@
+package com.paypal.android.paypalwebpayments
+
+import com.paypal.android.corepayments.PayPalSDKError
+
+/**
+ * Implement this callback to receive results from [PayPalWebCheckoutClient].
+ */
+interface PayPalWebCheckoutVaultListener {
```
sarahkoop:
```suggestion
interface PayPalWebVaultListener {
```

Does this need the `Checkout` in the name? Seems like all the methods are just for `Vault`.

Same comment for all of the `PayPalWebCheckoutVault` prefixed classes - seems pretty verbose can we just go with `PayPalWebVault`?

Demo/src/main/java/com/paypal/android/usecase/CreatePayPalSetupTokenUseCase.kt
sshropshire: A lot of this logic will be server side. The sample merchant server forwards the request to the SetupTokenAPI with a merchant client ID and client secret. TBH I've avoided using a serialization format for these requests mostly because we end up creating duplicate types that already exist in MXO.

The UseCase pattern gives us a nice unit of encapsulation for these requests, so it doesn't feel overly complex at the moment, but I'm open to re-strategizing that portion of the Demo app architecture.

Demo/src/main/java/com/paypal/android/ui/paypalwebvault/PayPalWebVaultView.kt
sshropshire: Oh wow yeah that's a good call! `PayPalWebCheckoutClient` takes a hard reference 🫨. I hadn't realized that until now. I think we may need to start version 2 sooner than later. At the moment `PayPalWebCheckoutClient` requires an activity reference at launch. We need to introduce a `PayPalWebLauncher` type to this project to keep `Activity` references out of the view model.

Demo/src/main/java/com/paypal/android/usecase/CreatePayPalPaymentTokenUseCase.kt
sshropshire: Yeah I agree it is very loosely-typed right now, mostly because we don't ship the Demo app we made a tradeoff for flexibility early on. But now that the app is getting large it makes sense to refactor a bit.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
sshropshire: Yeah it seems like the `listener?.also` pattern was there to skip parsing when no listener exists. I'm willing to bet though that this code doesn't work as intended. `deliverResult()` should only happen once, but the unit test mocks the behavior so that it does something different. We may need to refactor this because if I'm thinking it through correctly, the way [it's implemented currently](https://github.com/paypal/paypal-android/blob/fe8d82a0174e4bf49aaa2af5b8acd91bf428d459/PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt#L98) won't have a chance to defer delivery of a browser switch result.

When `browserSwitchClient.deliverResult(activity)` is called twice, the value at `browserSwitchResult` will become `null`. So even if a listener is set later on, there won't be a browser switch result to deliver because the pending result will have been cleared internally.

PPCP should use the `parseResult()` method as well. It'd be awesome if we could use an Activity launcher too because then browser switch could theoretically be launched from a Jetpack Composable without a direct activity reference.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutVaultListener.kt
sshropshire: Yeah I want to do this too, unfortunately the naming schema includes checkout in `v1`. I'm pretty confident we'll need to start moving towards `v2`. I've had that concern for a while, but this module makes it clear a clear decision.

Demo/src/main/java/com/paypal/android/ui/DemoApp.kt
```diff
@@ -98,6 +99,9 @@ fun DemoApp() {
                 composable(DemoAppDestinations.PAYPAL_WEB) {
                     PayPalWebView()
                 }
+                composable(DemoAppDestinations.PAYPAL_WEB_VAULT) {
+                    PayPalWebVaultView()
```
jaxdesmarais, line 103: General question - is there a reason we want a separate view for vaulting PayPal? iirc we have a toggle on iOS for vaulting PayPal vs a separate view.

sshropshire: For now it's simplicity. Each composable also typically has an associated `<FEATURE_NAME>ViewModel` type. We could make the code more DRY, but it would require us to define a common `interface` between the two view models.

Demo/src/main/java/com/paypal/android/api/model/CardPaymentToken.kt
KunJeongPark: Do you mean the payment source they return?

Demo/src/main/java/com/paypal/android/ui/DemoApp.kt
KunJeongPark: Same on iOS unless we are talking about something different.

Demo/src/main/java/com/paypal/android/api/model/CardPaymentToken.kt
sshropshire: Yeah CardPaymentToken has `cardLast4` and a few other properties that don't apply to `PayPalPaymentToken`.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -134,6 +182,34 @@ class PayPalWebCheckoutClient internal constructor(
...
+        if (deepLinkUrlResult != null && requestMetadata != null) {
+            val isFailure = deepLinkUrlResult.path?.contains(VAULT_CANCEL_URL_PATH) ?: false
+            if (isFailure) {
+                vaultListener?.onPayPalWebVaultFailure(PayPalWebCheckoutError.malformedResultError)
+            } else {
+                val approvalTokenId =
+                    deepLinkUrlResult.getQueryParameter(DEEP_LINK_PARAM_APPROVAL_TOKEN_ID)
+                val approvalSessionId =
+                    deepLinkUrlResult.getQueryParameter(DEEP_LINK_PARAM_APPROVAL_SESSION_ID)
+                val result = PayPalWebCheckoutVaultResult(approvalTokenId, approvalSessionId)
```
KunJeongPark: Oh we need to return the approvalSessionId? What is that for and is it meaningful to merchant?

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/BrowserSwitchHelper.kt
```diff
@@ -40,4 +44,16 @@ internal class BrowserSwitchHelper(private val urlScheme: String) {
...
             .metadata(metadata)
     }
+
+    fun configurePayPalVaultApproveSwitchOptions(
+        setupTokenId: String?,
+        approveOrderHref: String
+    ): BrowserSwitchOptions {
+        val metadata = JSONObject()
+            .put(METADATA_KEY_SETUP_TOKEN_ID, setupTokenId)
+        return BrowserSwitchOptions()
```
KunJeongPark: What is this metadata for Steven? Is this necessary, do we need to pass in the setupToken in there?
Is this needed specifically for Android? I am getting vaultID without it. 

This was not in the AC for iOS ticket. In iOS I am just taking the string from url and launching a webView.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
sarahkoop: I'd vote for a request object since that seems to be the consistent pattern across the SDK

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
sshropshire: I'm not 100% sure. I honestly just parse these values out because they're there.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/BrowserSwitchHelper.kt
sshropshire: This part is to fit into the way it's currently implemented on Android. We may not need it here either, it's subject to change. I doubt it's needed for iOS either because there's no concept of Browser Switching there.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
KunJeongPark: Without that value, the results from the vault are same for card and PayPal unless
it is decided that we return more values for card vaulting.

sshropshire: Yeah for this PR we have two vault result types, `PayPalWebCheckoutVaultResult` and `CardVaultResult`. Are y'all using a single result type on iOS?

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsService.kt
```diff
@@ -43,7 +43,7 @@ class AnalyticsService internal constructor(
     fun sendAnalyticsEvent(name: String, orderId: String?) {
         // TODO: send analytics event using WorkManager (supports coroutines) to avoid lint error
         // thrown because we don't use the Deferred result
-        scope.async {
+        scope.launch {
```
sshropshire, line 46: `launch()` is more appropriate here–`async()` typically has a balancing call to `await()`. In the long-term, once we migrate to WorkManager this will no longer be required.

Demo/src/main/java/com/paypal/android/ui/paypalwebvault/PayPalWebVaultView.kt
sshropshire: So unfortunately we won't be able to remove the hard reference in this PR, but we have discussed scoping out the work to remove this and start the `v2` beta.

Demo/src/main/java/com/paypal/android/usecase/CreatePayPalPaymentTokenUseCase.kt
sshropshire: I tried using `kotlinx.serialization` here and I ultimately decided to backtrack. There are some limitations that I'd like to have a larger discussion about with everyone, and see if we can land on a set of JSON serialization best practices.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -40,36 +42,29 @@ class PayPalWebCheckoutClient internal constructor(
...
+        )
     )
 
-    private val exceptionHandler = CoreCoroutineExceptionHandler {
-        deliverFailure(it)
-    }
-
-    private var browserSwitchResult: BrowserSwitchResult? = null
-
-    private var orderId: String? = null
```
sshropshire, line 55: I decided to remove these internal state properties. We should avoid having mutable state in the clients whenever possible. It also makes the tests much simpler to write. We don't have these on BT so I don't think it'll be problematic.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -78,69 +73,67 @@ class PayPalWebCheckoutClient internal constructor(
...
-            } catch (e: PayPalSDKError) {
-                deliverFailure(APIClientError.clientIDNotFoundError(e.code, e.correlationId))
-            }
+        analyticsService.sendAnalyticsEvent("paypal-web-payments:started", request.orderId)
+        payPalWebLauncher.launchPayPalWebCheckout(activity, request)?.let { launchError ->
+            notifyWebCheckoutFailure(launchError)
         }
     }
 
+    fun vault(activity: AppCompatActivity, request: PayPalWebVaultRequest) {
```
sshropshire: Still working on docs, but I wanted to get the fundamental structure going before adding that last piece of polish.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutVaultExperienceContext.kt
sshropshire: Same for docs here. Not sure if we'll end up keeping this. I'd like to see what iOS does to collaborate on the best API.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutVaultListener.kt
sshropshire: I may actually be down to go for it here. The refactor was interesting, we're browser switching in two different ways in the same PayPal module. `PayPalWebVaultListener` will set us up nicely before addressing existing `v1` naming concerns in `v2`.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebLauncher.kt
```diff
@@ -0,0 +1,162 @@
...
+import com.paypal.android.corepayments.CoreConfig
+import com.paypal.android.corepayments.Environment
+import com.paypal.android.corepayments.PayPalSDKError
+import com.paypal.android.paypalwebpayments.PayPalWebCheckoutClient.Companion.DEEP_LINK_PARAM_APPROVAL_SESSION_ID
+import com.paypal.android.paypalwebpayments.PayPalWebCheckoutClient.Companion.DEEP_LINK_PARAM_APPROVAL_TOKEN_ID
+import com.paypal.android.paypalwebpayments.PayPalWebCheckoutClient.Companion.VAULT_HOST
+import com.paypal.android.paypalwebpayments.errors.PayPalWebCheckoutError
+import org.json.JSONObject
+
+internal class PayPalWebLauncher(
```
sshropshire, line 16: This extracts browser switching behavior from `PayPalWebCheckoutClient` and combines it with `BrowserSwitchHelper` logic to encapsulate all browser switching logic in one place. In `v2` we'll be able to make this class public and adjust the API so that it works in tandem with `PayPalWebCheckoutClient`.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebLauncher.kt
```diff
@@ -0,0 +1,162 @@
...
+            error = APIClientError.clientIDNotFoundError(e.code, e.correlationId)
+        }
+        return error
+    }
+
+    fun launchPayPalWebVault(activity: FragmentActivity, vaultRequest: PayPalWebVaultRequest) {
+        val browserSwitchOptions = vaultRequest.run {
+            configurePayPalVaultApproveSwitchOptions(setupTokenId, approveVaultHref)
+        }
+        browserSwitchClient.start(activity, browserSwitchOptions)
```
sarahkoop: This method throws a `BrowserSwitchException`, should we catch that here and return a PayPalSDKError? Or update the method to throw that error as well? (Same comment for `launchPayPalWebCheckout`)

sshropshire: Yeah good call. I noticed that it seems like we never did catch this error on PPCP. It's tough because all exceptions are unchecked on the Kotlin side, so we don't get a compiler warning in this case.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
KunJeongPark: No I kept it different because card endpoint sends a lot of info and I thought there might be request to include more data in the future. But as of now, what we are returning is same except for the approval session id.

sshropshire: I ended up simplifying to `approvalSessionId` only on Android.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -78,69 +56,74 @@ class PayPalWebCheckoutClient internal constructor(
...
-                deliverFailure(PayPalWebCheckoutError.malformedResultError)
+    internal fun handleBrowserSwitchResult() {
+        payPalWebLauncher.deliverBrowserSwitchResult(activity)?.let { status ->
+            when (status) {
+                is PayPalWebStatus.CheckoutSuccess -> notifyWebCheckoutSuccess(status.result)
+                is PayPalWebStatus.CheckoutError -> notifyWebCheckoutFailure(status.error)
+                is PayPalWebStatus.CheckoutCanceled -> notifyWebCheckoutCancelation(status.orderId)
+                is PayPalWebStatus.VaultSuccess -> notifyVaultSuccess(status.result)
+                is PayPalWebStatus.VaultError -> notifyVaultFailure(status.error)
+                PayPalWebStatus.VaultCanceled -> notifyVaultCancelation()
```
sarahkoop, line 84:
```suggestion
               is PayPalWebStatus.VaultCanceled -> notifyVaultCancelation()
```
surprised this compiled - is the "is" not required? 🤔

sarahkoop (approved):
+1 on the next major version refactor notes, otherwise LGTM

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
sshropshire: Yeah in this case `PayPalWebStatus.VaultCanceled` is an `object` type so it's equivalent to `status === PayPalWebStatus.VaultCanceled` in the `when` statement.

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalWebViewModel.kt
```diff
@@ -117,26 +118,29 @@ class PayPalWebViewModel @Inject constructor(
...
+    }
+
     fun startWebCheckout(activity: AppCompatActivity) {
         isStartCheckoutLoading = true
         viewModelScope.launch {
-            try {
-                val clientId = sdkSampleServerAPI.fetchClientId()
+            fetchClientId()?.let { clientId ->
                 val coreConfig = CoreConfig(clientId)
                 payPalDataCollector = PayPalDataCollector(coreConfig)
```
KunJeongPark, line 138: We don't have this step in iOS demo app.

Demo/src/main/java/com/paypal/android/ui/paypalwebvault/PayPalWebVaultViewModel.kt
```diff
@@ -0,0 +1,143 @@
...
+    companion object {
+        const val URL_SCHEME = "com.paypal.android.demo"
+    }
+
+    private val _uiState = MutableStateFlow(PayPalWebVaultUiState())
+    val uiState = _uiState.asStateFlow()
+
+    private lateinit var paypalClient: PayPalWebCheckoutClient
+    private lateinit var payPalDataCollector: PayPalDataCollector
+
```
KunJeongPark, line 41: seems like you have PayPalDataCollector property with each viewModel. We should do that for iOS demo app as well.

KunJeongPark (approved):

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalWebViewModel.kt
sshropshire: Yeah merchants won't typically have a method like this either. It's only in the demo app to allow support multiple client IDs when we allow users to change the integration type (Direct, CP4, etc.).

Demo/src/main/java/com/paypal/android/ui/paypalwebvault/PayPalWebVaultViewModel.kt
sshropshire: Yeah for now we're sending the client metadata ID along with calls to Capture / Authorize from the Demo app.

PR #222, opened 2023-12-18
Material Theme

sshropshire:
### Summary of changes

 - Refactor Demo app UI to use Material themes
 - Refine design language across each feature demo
 - Add `StepHeader` composable to signal progress to the user as they complete steps
 - Add `UIConstants` object to help increase uniformity in the Demo app layout for each feature

 ### Checklist

### Screenshot

https://github.com/paypal/paypal-android/assets/58225613/e54991c0-4857-458f-80f2-89400c66314e

- @sshropshire


Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderView.kt
```diff
@@ -19,91 +17,143 @@ import androidx.compose.ui.Modifier
...
+        }
+        if (uiState.isApproveOrderSuccessful) {
+            Step3_CompleteOrder(uiState, viewModel)
+        }
+        Spacer(modifier = Modifier.size(contentPadding))
+    }
+}
+
+@Composable
+private fun Step1_CreateOrder(uiState: ApproveOrderUiState, viewModel: ApproveOrderViewModel) {
```
sarahkoop, line 69: I like the steps naming here very helpful 👍

sarahkoop:
Does this demo app get published to the play store now? Wondering if there are limitations on using the PP logo if it is

sshropshire:
> Does this demo app get published to the play store now? Wondering if there are limitations on using the PP logo if it is

Yeah that's a good call. For now they should just be internal builds, but I'm not sure if there's a way to prevent it 100%. But it would take a lot of manual effort to do so, but mostly want to make it distinguishable on the launch screen. The generic Android app icon can give hello world vibes I feel.

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalWebCheckoutResultView.kt
```diff
@@ -1,28 +1,19 @@
...
-        Column(modifier = Modifier.padding(8.dp)) {
-            Text(
-                text = stringResource(id = R.string.order_approved),
-                style = MaterialTheme.typography.titleLarge
-            )
-            PropertyView(name = "Order ID", value = orderId)
-            PropertyView(name = "Payer ID", value = payerId)
-        }
+    Column(
+        verticalArrangement = UIConstants.spacingMedium,
```
scannillo, line 13: 👏 For the UIConstants

Demo/src/main/java/com/paypal/android/uishared/components/StepHeader.kt
```diff
@@ -0,0 +1,77 @@
...
+        Box(
+            modifier = Modifier
+                .background(
+                    color = MaterialTheme.colorScheme.inverseSurface,
+                    shape = CircleShape
+                )
+                .size(40.dp)
+        ) {
+            Text(
+                text = "$stepNumber",
```
scannillo, line 35: Visually this Step # is also super helpful 👏

Demo/src/main/java/com/paypal/android/uishared/components/StepHeader.kt
```diff
@@ -0,0 +1,77 @@
...
+import androidx.compose.ui.text.PlatformTextStyle
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.paypal.android.utils.UIConstants
+
+@Composable
+fun StepHeader(
+    stepNumber: Int,
+    title: String,
+    modifier: Modifier = Modifier
```
scannillo: Do we need this modifier param? I don't see it used in this PR anywhere

sarahkoop (approved):

Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
```diff
@@ -99,23 +98,17 @@ class ApproveOrderViewModel @Inject constructor(
...
         viewModelScope.launch {
-            isCompleteOrderLoading = true
+            completeOrderState = ActionState.Loading
 
             val cmid = payPalDataCollector.collectDeviceData(context)
-            val orderId = createdOrder!!.id!!
-
-            completedOrder = completeOrderUseCase(orderId, intentOption, cmid)
-            isCompleteOrderLoading = false
+            val completedOrder = completeOrderUseCase(createdOrder!!.id!!, intentOption, cmid)
```
tdchow, line 105: Looks like this was already existing - would we want to prevent the demo app crash when `createOrder` or `id` is null?

Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
```diff
@@ -99,23 +98,17 @@ class ApproveOrderViewModel @Inject constructor(
...
     }
 
     private fun createCardRequest(uiState: ApproveOrderUiState, order: Order): CardRequest {
         val card = parseCard(uiState)
-        val sca = when (uiState.scaOption) {
-            "ALWAYS" -> SCA.SCA_ALWAYS
-            else -> SCA.SCA_WHEN_REQUIRED
-        }
-        return CardRequest(order.id!!, card, APP_RETURN_URL, sca)
+        return CardRequest(order.id!!, card, APP_RETURN_URL, uiState.scaOption)
```
tdchow, line 112: Same question about null order id here

Demo/src/main/java/com/paypal/android/uishared/components/StepHeader.kt
sshropshire: Yeah technically we aren't using it, but if we ever wanted to apply styles like padding, etc. from from the parent element we could do so. I saw it in a lot of the sample projects Google has, but you're right it doesn't have much use in our project at the moment. I'm down to take or leave.

Demo/src/main/java/com/paypal/android/uishared/components/StepHeader.kt
sshropshire: 🙌 ! Still trying to nail a UX pattern. The original thought was to accompany it with a Material3 progress bar at the top, but this should be a good start.

Demo/src/main/java/com/paypal/android/ui/paypalbuttons/PayPalButtonsView.kt
```diff
@@ -78,39 +84,74 @@ fun PayPalButtonsView(viewModel: PayPalButtonsViewModel = viewModel()) {
...
+            Slider(
+                value = value ?: 0.0f,
+                valueRange = 0f..CORNER_RADIUS_SLIDER_MAX,
+                steps = CORNER_RADIUS_SLIDER_MAX.toInt(),
+                colors = SliderDefaults.colors(
+                    inactiveTrackColor = MaterialTheme.colorScheme.onSurfaceVariant,
+                ),
+                onValueChange = { value ->
+                    // round to integer
+                    val roundedValue = value.roundToInt().toFloat()
```
tdchow: Is there a reason why we need to round the value to an integer? Would it be better to have this rounding logic live in the ViewModel where `viewModel.customCornerRadius` is set?

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalWebUiState.kt
```diff
@@ -5,16 +5,18 @@ import com.paypal.android.api.model.OrderIntent
...
-    val isStartCheckoutLoading: Boolean = false,
-    val createdOrder: Order? = null,
-    val completedOrder: Order? = null,
-    val payPalWebCheckoutResult: PayPalWebCheckoutResult? = null,
-    val payPalWebCheckoutError: PayPalSDKError? = null,
-    val isCheckoutCanceled: Boolean = false,
-    val isCompleteOrderLoading: Boolean = false,
+    val createOrderState: ActionState<Order, Exception> = ActionState.Ready,
+    val payPalWebCheckoutState: ActionState<PayPalWebCheckoutResult, Exception> = ActionState.Ready,
+    val completeOrderState: ActionState<Order, PayPalSDKError> = ActionState.Ready,
```
tdchow: The introduction of ActionStates really cleans up the number of properties in the UiState 🎉

Demo/src/main/java/com/paypal/android/uishared/components/StepHeader.kt
```diff
@@ -0,0 +1,77 @@
...
+        verticalAlignment = Alignment.CenterVertically,
+        modifier = modifier
+    ) {
+        Box(
+            modifier = Modifier
+                .background(
+                    color = MaterialTheme.colorScheme.inverseSurface,
+                    shape = CircleShape
+                )
+                .size(40.dp)
```
tdchow: Would it be worth adding the `40.dp` to the UIConstants?

tdchow (approved):
The compose views are looking great! Just had a few non blocking questions.

Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
sshropshire: Yeah great call. The best solution will most likely be to display an error in this case. It's tough because if it gets to that point it's highly probably that `createdOrder` is non-null. It makes me wonder if it's a code smell? I wonder if there's a design pattern we can leverage here to make it super clean.

Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
sshropshire: Yeah I'm leaning towards updating the `state` variable with an UnexpectedError state, almost along the lines of "This should not be null 😅" lol

Demo/src/main/java/com/paypal/android/ui/paypalbuttons/PayPalButtonsView.kt
sshropshire: :nod: good call yeah, that's a much better place to put the rounding logic.

Demo/src/main/java/com/paypal/android/uishared/components/StepHeader.kt
sshropshire: Totally. We can cut down on the magic numbers I agree. Still figuring out the overall semantics of the Compose UI in the Demo app. Updating this will be a good move in the right direction.

Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
tdchow: One option would be to fail fast when a merchant launches the SDK and required params or config is missing. This would allow us to have non-null types for these fields inside the SDK. No need to update this PR, but it might be someone looking into as a tech debt story or hack day project.

sshropshire: Ok cool. That does sound like a good idea. We can probably return an error result immediately from the UseCase to prevent the force unwraps.

Demo/src/main/java/com/paypal/android/uishared/components/StepHeader.kt
sshropshire: Went ahead and took it out 👍

Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
sshropshire: I'm gonna create a ticket for this tech debt story. We can fix this for all `UseCase` related code to make sure we're handling errors properly throughout the Demo app.

Demo/src/main/java/com/paypal/android/ui/paypalbuttons/PayPalButtonsView.kt
sshropshire: After some refactoring, ended up landing on an `IntSlider` to encapsulate the conversion logic (in [e3374da](https://github.com/paypal/paypal-android/pull/222/commits/e3374da17a5595f9a6b22288c5b30131d1740ec7)). Now the view model will only see `Int` types.

Demo/src/main/java/com/paypal/android/uishared/components/StepHeader.kt
sshropshire: Cleaned up remaining magic numbers (in [a52845c](https://github.com/paypal/paypal-android/pull/222/commits/a52845c83ae4356df787da376fb11eeb41a28d13)).

Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
sshropshire: Added a ticket DTPPCPSDK-1706 for addressing the tech debt.

PR #223, opened 2023-12-19
Add corner radius property and slider in the demo

warmkesselj:
### Summary of changes

 - Support custom corner radius on the PayPal button
 

 ### Checklist

 - [x] Added a changelog entry

### Authors
@scannillo 
@warmkesselj 

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # PayPal Android SDK Release Notes
 
+## UNRELEASED 
+* PaymentButtons
+  * Supporting custom corner radius on the PayPal Button
+
 ## 1.1.0 (2023-12-05)
 
 * PayPalNativeCheckout
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # PayPal Android SDK Release Notes
 
-## UNRELEASED 
+## unreleased 
 * PaymentButtons
   * Supporting custom corner radius on the PayPal Button
 
--- a/PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
+++ b/PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
@@ -72,6 +72,25 @@ abstract class PaymentButton<C : PaymentButtonColor> @JvmOverloads constructor(
             prefixTextView.visibility = prefixTextVisibility
         }
 
+    var customCornerRadius: Float? = null
+        set(value) {
+            field = value // Set the field value
+
+            // Determine the corner treatment based on the value
+            val cornerTreatment = if (value == 0.0f) {
+                CutCornerTreatment()
+            } else {
+                RoundedCornerTreatment()
+            }
+
+            // Build the shapeAppearanceModel considering the possibility of a null customCornerRadius
+            shapeAppearanceModel = ShapeAppearanceModel.builder().apply {
+                value?.let {
+                    setAllCornerSizes(it)
+                }
+                setAllCorners(cornerTreatment)
+            }.build()
+        }
     /**
      * Updates the shape of the Payment Button with the provided [PaymentButtonShape]
      * and defaults to [ROUNDED] if one is not provided.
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # PayPal Android SDK Release Notes
 
+## UNRELEASED
```
sarahkoop:
```suggestion
## unreleased

```

Nit but I think the release CI requires that formatting

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
```diff
@@ -72,6 +72,27 @@ abstract class PaymentButton<C : PaymentButtonColor> @JvmOverloads constructor(
             prefixTextView.visibility = prefixTextVisibility
         }
 
+    /**
+     * Custom corner radius
```
sarahkoop: Can we add a note to what the default is if not set?

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
```diff
@@ -94,9 +115,18 @@ abstract class PaymentButton<C : PaymentButtonColor> @JvmOverloads constructor(
...
 
-            val cornerTreatment = when (field) {
-                PaymentButtonShape.ROUNDED, PaymentButtonShape.PILL -> RoundedCornerTreatment()
-                PaymentButtonShape.RECTANGLE -> CutCornerTreatment()
+            val cornerTreatment = if (customCornerRadius != null) {
+                if (customCornerRadius == 0.0f) {
+                    CutCornerTreatment()
+                } else {
+                    RoundedCornerTreatment()
+                }
```
sarahkoop: This seems duplicate of lines 82-86 - does it need to be set again here?

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
```diff
@@ -72,6 +72,27 @@ abstract class PaymentButton<C : PaymentButtonColor> @JvmOverloads constructor(
...
         }
 
+    /**
+     * Custom corner radius
+     */
+    var customCornerRadius: Float? = null
+        set(value) {
+            field = value
+
+            val cornerTreatment = if (value == 0.0f) {
```
tdchow, line 86: It's probably an edge case that a merchant would set `customCornerRadius` as null, but would we want to have a null check here and not change the corner treatment?

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
tdchow: I think this is the set function for the `var shape: PaymentButtonShape`. So the `customCornerRadius` variable is checked to see if a value was set previously.

Is this logic intended behavior? If a merchant first sets `customCornerRadius` > 0 and then sets `shape` as `PaymentButtonShape.RECTANGLE`, the button will still display with rounded corners. Should we always style the button with the last property the merchant sets?

tdchow (approved):

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
sarahkoop: Sorry to clarify - what will the corner radius of the button be if the this is not set?

warmkesselj: No worries, this would be set to null.

sarahkoop (approved):

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
sshropshire: It'd be nice if we could make `var shape: PaymentButtonShape` a write-only, virtual property and make `var cornerRadius: Float` the source of truth. Kotlin doesn't seem to have explicit support for [write-only props](https://youtrack.jetbrains.com/issue/KT-6519) though. We may still be able to refactor our logic to eliminate nullability for both properties.

The duplication may be a sign of technical debt. It may make sense to the revisit the requirements for these buttons to make sure we're properly aligned with product.

PR #227, opened 2024-01-05
Parse 3DS Liability Shift in Card Approve Order Result

sshropshire:
### Summary of changes

 - Parse `liability_shift` parameter from `confirm-payment-source` 3DS contingency deep link response
 - Parse `error` parameter from `confirm-payment-source` 3DS contingency deep link response
 - Return `PayPalSDKError` when an error is detected in the 3DS contingency deep link response

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * CardPayments
   * Add `liabilityShift` property to `CardResult`
   * Parse `liabilityShift` from deep link return URL in `CardClient#approveOrder()` browser-switched flow
-  * Callback `PayPalSDKError` when CardClient#approveOrder()` 3DS verification fails
+  * Callback `PayPalSDKError` when `CardClient#approveOrder()` 3DS verification fails
 
 ## 1.3.0 (2024-01-09)
 * PaymentButtons
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # PayPal Android SDK Release Notes
 
+## unreleased
+
+* CardPayments
+  * Add `liabilityShift` property to `CardResult`
+  * Parse `liabilityShift` from deep link return URL in `CardClient#approveOrder()` browser-switched flow
+  * Callback `PayPalSDKError` when CardClient#approveOrder()` 3DS verification fails
+
 ## 1.3.0 (2024-01-09)
 * PaymentButtons
   * Add `PayPalCreditButtonColor.WHITE` and `.GOLD`
@@ -9,6 +16,7 @@
   * Bump native-checkout version to release `1.2.1`
 
 ## 1.2.0 (2024-01-04) 
+
 * PaymentButtons
   * Supporting custom corner radius on the PayPal Button
 * PayPalWebPayments
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 * CardPayments
   * Add `liabilityShift` property to `CardResult`
-  * Parse `liabilityShift` from deep link return URL in `CardClient#approveOrder()` browser-switched flow
   * Callback `PayPalSDKError` when `CardClient#approveOrder()` 3DS verification fails
 
 ## 1.3.0 (2024-01-09)
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
@@ -176,12 +176,7 @@ class CardClient internal constructor(
 
     @Throws(PayPalSDKError::class)
     private fun parseApproveOrderDeepLink(orderId: String, deepLinkUrl: Uri?): CardResult {
-        if (deepLinkUrl == null) {
-            throw CardError.threeDSVerificationError
-        }
-
-        val error = deepLinkUrl.getQueryParameter("error")
-        if (error != null) {
+        if (deepLinkUrl == null || deepLinkUrl.getQueryParameter("error") != null) {
             throw CardError.threeDSVerificationError
         }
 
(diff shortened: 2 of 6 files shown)


CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -160,26 +160,35 @@ class CardClient internal constructor(
...
-                    val result = CardResult(metadata.orderId, deepLinkUrl)
-                    notifyApproveOrderSuccess(result)
-                } catch (error: PayPalSDKError) {
-                    analyticsService.sendAnalyticsEvent(
-                        "card-payments:3ds:get-order-info:failed",
-                        metadata.orderId
-                    )
-                    throw error
-                }
+            try {
```
sshropshire, line 163: This internals of this method can work without coroutines.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -160,26 +160,35 @@ class CardClient internal constructor(
...
+                analyticsService.sendAnalyticsEvent(
+                    "card-payments:3ds:get-order-info:succeeded",
+                    metadata.orderId
+                )
```
scannillo: Looks like when we removed the getOrder API call (after we had to move from accessTokens to clientIDs) we never removed these analytic events. Could be another PR, just calling it out.

CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 # PayPal Android SDK Release Notes
 
+## unreleased
+
+* CardPayments
+  * Add `liabilityShift` property to `CardResult`
+  * Parse `liabilityShift` from deep link return URL in `CardClient#approveOrder()` browser-switched flow
+  * Callback `PayPalSDKError` when CardClient#approveOrder()` 3DS verification fails
```
scannillo:
```suggestion
  * Callback `PayPalSDKError` when `CardClient#approveOrder()` 3DS verification fails
```

scannillo (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 # PayPal Android SDK Release Notes
 
+## unreleased
+
+* CardPayments
+  * Add `liabilityShift` property to `CardResult`
+  * Parse `liabilityShift` from deep link return URL in `CardClient#approveOrder()` browser-switched flow
```
sarahkoop: This feels like an implementation detail - is it different from the line above (from a merchant perspective)?

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -160,26 +160,35 @@ class CardClient internal constructor(
...
+                notifyApproveOrderFailure(error)
             }
         }
     }
 
+    @Throws(PayPalSDKError::class)
+    private fun parseApproveOrderDeepLink(orderId: String, deepLinkUrl: Uri?): CardResult {
+        val error = deepLinkUrl?.getQueryParameter("error")
```
sarahkoop: Should we also return an error if the deepLinkUrl is null? Looks like currently that scenario will just return a CardResult with null params - not sure what the desired behavior is

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
sshropshire: Oh good call. I'll go ahead and remove 👍 .

CHANGELOG.md
sshropshire: That's true. I'll remove this too.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
sshropshire: Good catch I'll throw an error here when that happens. A user could technically clear their storage whenever they want.

sshropshire: Actually thinking it through this won't occur on a storage clear. Either way added a test case for it 👍 .

sarahkoop (approved):

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -160,26 +160,35 @@ class CardClient internal constructor(
...
+            throw CardError.threeDSVerificationError
+        }
+
+        val error = deepLinkUrl.getQueryParameter("error")
+        if (error != null) {
+            throw CardError.threeDSVerificationError
+        } else {
+            val liabilityShift = deepLinkUrl.getQueryParameter("liability_shift")
+            return CardResult(orderId, deepLinkUrl, liabilityShift)
+        }
```
KunJeongPark, line 181: Steven, in iOS, we also check for conditions for success and throw another error if it is not in the format specified. To me, it seems risky to assume success if error parameter is absent.
https://github.com/paypal/paypal-ios/blob/2896471c1533de72880afd18545a33a374f9cfc0/Sources/CardPayments/CardClient.swift#L140
What do you think?

sshropshire: I can add those checks in 👍 . I missed that part from the iOS PR.

sshropshire: Added here in [af7cbad](https://github.com/paypal/paypal-android/pull/227/commits/af7cbad6aede5a940ae13cc34ed4fd1bc06bb05f).

PR #240, opened 2024-02-16
Send Order ID for Analytics Events When Possible

sshropshire:
### Summary of changes

 - Add Order ID to analytics calls where possible

 ### Checklist

- @sshropshire

### DIFF
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
@@ -39,10 +39,11 @@ class CardClient internal constructor(
      */
     var cardVaultListener: CardVaultListener? = null
 
+    private var approveOrderId: String? = null
     private val lifeCycleObserver = CardLifeCycleObserver(this)
 
     private val approveOrderExceptionHandler = CoreCoroutineExceptionHandler { error ->
-        notifyApproveOrderFailure(error, null)
+        notifyApproveOrderFailure(error, approveOrderId)
     }
 
     private val vaultExceptionHandler = CoreCoroutineExceptionHandler { error ->
@@ -76,6 +77,7 @@ class CardClient internal constructor(
      * @param cardRequest [CardRequest] for requesting an order approval
      */
     fun approveOrder(activity: FragmentActivity, cardRequest: CardRequest) {
+        approveOrderId = cardRequest.orderId
         analyticsService.sendAnalyticsEvent("card-payments:3ds:started", cardRequest.orderId)
 
         CoroutineScope(dispatcher).launch(approveOrderExceptionHandler) {
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
@@ -42,14 +42,12 @@ class CardClient internal constructor(
     private val lifeCycleObserver = CardLifeCycleObserver(this)
 
     private val approveOrderExceptionHandler = CoreCoroutineExceptionHandler { error ->
-        notifyApproveOrderFailure(error)
+        notifyApproveOrderFailure(error, null)
     }
 
     private val vaultExceptionHandler = CoreCoroutineExceptionHandler { error ->
         cardVaultListener?.onVaultFailure(error)
     }
-
-    private var orderId: String? = null
 
     /**
      *  CardClient constructor
@@ -78,8 +76,7 @@ class CardClient internal constructor(
      * @param cardRequest [CardRequest] for requesting an order approval
      */
     fun approveOrder(activity: FragmentActivity, cardRequest: CardRequest) {
-        orderId = cardRequest.orderId
-        analyticsService.sendAnalyticsEvent("card-payments:3ds:started", orderId)
+        analyticsService.sendAnalyticsEvent("card-payments:3ds:started", cardRequest.orderId)
 
         CoroutineScope(dispatcher).launch(approveOrderExceptionHandler) {
             confirmPaymentSource(activity, cardRequest)
@@ -100,7 +97,7 @@ class CardClient internal constructor(
             } else {
                 analyticsService.sendAnalyticsEvent(
                     "card-payments:3ds:confirm-payment-source:challenge-required",
-                    orderId
+                    cardRequest.orderId
                 )
                 approveOrderListener?.onApproveOrderThreeDSecureWillLaunch()
 
@@ -153,7 +150,7 @@ class CardClient internal constructor(
             approveOrderListener?.onApproveOrderThreeDSecureDidFinish()
             when (browserSwitchResult.status) {
                 BrowserSwitchStatus.SUCCESS -> handleBrowserSwitchSuccess(browserSwitchResult)
-                BrowserSwitchStatus.CANCELED -> notifyApproveOrderCanceled()
+                BrowserSwitchStatus.CANCELED -> notifyApproveOrderCanceled(browserSwitchResult)
             }
         }
     }
@@ -169,7 +166,7 @@ class CardClient internal constructor(
                     "card-payments:3ds:get-order-info:failed",
                     metadata.orderId
                 )
-                notifyApproveOrderFailure(error)
+                notifyApproveOrderFailure(error, metadata.orderId)
             }
         }
     }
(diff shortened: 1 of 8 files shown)


CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -42,15 +42,13 @@ class CardClient internal constructor(
     private val lifeCycleObserver = CardLifeCycleObserver(this)
 
     private val approveOrderExceptionHandler = CoreCoroutineExceptionHandler { error ->
-        notifyApproveOrderFailure(error)
+        notifyApproveOrderFailure(error, null)
```
sshropshire: Order ID not available here.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -42,15 +42,13 @@ class CardClient internal constructor(
...
     }
 
     private val vaultExceptionHandler = CoreCoroutineExceptionHandler { error ->
         cardVaultListener?.onVaultFailure(error)
     }
 
-    private var orderId: String? = null
```
sshropshire, line 52: It feels like a code smell saving `orderId` in a member variable since they never get cleared. I've removed these internal state variables where possible.

PayPalNativePayments/src/main/java/com/paypal/android/paypalnativepayments/PayPalNativeCheckoutClient.kt
```diff
@@ -163,7 +160,7 @@ class PayPalNativeCheckoutClient internal constructor(
                 ShippingChangeType.OPTION_CHANGE -> {
                     analyticsService.sendAnalyticsEvent(
                         "paypal-native-payments:shipping-method-changed",
-                        orderId
+                        null
```
sshropshire: Order ID not available in MXO shipping callback.

PayPalNativePayments/src/main/java/com/paypal/android/paypalnativepayments/PayPalNativeCheckoutClient.kt
```diff
@@ -175,7 +172,7 @@ class PayPalNativeCheckoutClient internal constructor(
     }
 
     private fun notifyOnCancel() {
-        analyticsService.sendAnalyticsEvent("paypal-native-payments:canceled", orderId)
+        analyticsService.sendAnalyticsEvent("paypal-native-payments:canceled", null)
```
sshropshire: Order ID not available in cancel callback.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -87,43 +90,32 @@ class PayPalWebCheckoutClient internal constructor(
...
+        analyticsService.sendAnalyticsEvent("paypal-web-payments:browser-login:canceled", orderId)
         listener?.onPayPalWebCanceled()
     }
 
     private fun notifyVaultSuccess(result: PayPalWebVaultResult) {
-        analyticsService.sendAnalyticsEvent(
-            "paypal-web-payments:browser-login:canceled",
-            null
-        )
+        analyticsService.sendAnalyticsEvent("paypal-web-payments:browser-login:canceled")
```
sshropshire: There's no order ID available for Vault without Purchase Flows. At the same time this analytics event is definitely off. Have we defined analytics event names for Vault?

PayPalWebPayments/src/test/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClientUnitTest.kt
```diff
@@ -113,7 +113,7 @@ class PayPalWebCheckoutClientUnitTest {
         val error = PayPalSDKError(123, "fake-error-description")
         every {
             payPalWebLauncher.deliverBrowserSwitchResult(activity)
-        } returns PayPalWebStatus.CheckoutError(error)
+        } returns PayPalWebStatus.CheckoutError(error, null)
```
sshropshire, line 116: Small fix for a compilation error.

PayPalNativePayments/src/main/java/com/paypal/android/paypalnativepayments/PayPalNativeCheckoutClient.kt
scannillo: The `orderID` is what the data team uses to "stitch" events together. So for example if we wanted to calculate the total % of PayPalNative flows that were cancelled, they compare the # events from "paypal-native-payments:started" & "paypal-native-payments:canceled" that have the same `orderID`. I believe they discard any events that don't have an `orderID` linked.

scannillo: Also, I notice on iOS that we require an `orderID` at the `AnalyticsService` init level. We could do the same for Android?

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
scannillo: Yeh good call. Vault w/o purchase won't have the orderID. We could ask the data team if they want us to track the `setupTokenId` in the payload.

As for event names, I proposed some in this ticket (DTPPCPSDK-707), but I don't think they've been agreed upon yet. Maybe give those a look?

PayPalNativePayments/src/main/java/com/paypal/android/paypalnativepayments/PayPalNativeCheckoutClient.kt
jaxdesmarais: I _think_ the `payToken` in `ShippingChangeData` may be an order ID? But maybe @saperi22 or @tdchow can confirm.

PayPalNativePayments/src/main/java/com/paypal/android/paypalnativepayments/PayPalNativeCheckoutClient.kt
sshropshire: Oh wow wouldn't that cause a feature client to send the same initial order ID when it's used multiple times?

PayPalNativePayments/src/main/java/com/paypal/android/paypalnativepayments/PayPalNativeCheckoutClient.kt
sshropshire: Oh nice yeah i'll use it here until we get confirmation. We can fall back to the OG orderId param only when we need to.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
sshropshire: For now, keeping the `orderId` parameter (renamed to `approveOrderId`) until we talk to the data time and decide if it's absolutely necessary for every analytics call.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
sshropshire: 🙌 ! Updated to match the event names from the doc.

Only updating these because they seem to be wrong. We may want to do a more comprehensive Analytics audit in a separate PR to make sure everything's in line.

PayPalNativePayments/src/main/java/com/paypal/android/paypalnativepayments/PayPalNativeCheckoutClient.kt
scannillo: We're resetting it if the `start()` method is called again with a new request/orderID ([snippet](https://github.com/paypal/paypal-ios/blob/4e12e8b8e0e3cadd7aaa6e0fbc47126ffcc59f62/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift#L36))

sshropshire: Oh got it yeah that's definite architectural parity drift between Android & iOS. It seems more like an `AnalyticsClient` on iOS then since there are multiple instances associated with a given Feature Client over time.

There's still work to be done on the Android side for `AnalyticsService` as well, we still haven't batched FPTI events in the background like we do on BT.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -39,18 +39,17 @@ class CardClient internal constructor(
      */
     var cardVaultListener: CardVaultListener? = null
 
+    private var approveOrderId: String? = null
     private val lifeCycleObserver = CardLifeCycleObserver(this)
 
     private val approveOrderExceptionHandler = CoreCoroutineExceptionHandler { error ->
-        notifyApproveOrderFailure(error)
+        notifyApproveOrderFailure(error, approveOrderId)
     }
```
KunJeongPark, line 48: Oh I guess we need to do this in iOS as well, sending analytics with orderID on fail cases.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -39,18 +39,17 @@ class CardClient internal constructor(
      */
     var cardVaultListener: CardVaultListener? = null
 
+    private var approveOrderId: String? = null
     private val lifeCycleObserver = CardLifeCycleObserver(this)
```
KunJeongPark, line 43: why do you need this? orderID is in the CardRequest

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
sshropshire: Yeah I wish we could not have it tbh but this is in line with iOS.

sshropshire: We may re-visit one day to see if we really need it. For now it's easiest to maintain analytics parity.

jaxdesmarais (approved):

tdchow (approved):

PR #243, opened 2024-03-04
Card: Vault without Purchase 3DS

sshropshire:
### What

This PR adds Card Vault without Purchase 3DS authentication to the Demo app.

### Summary of changes

  * Add `CardClient#presentAuthChallenge()`
  * Add `returnUrl` property to `CardVaultRequest`
  * Add `authChallenge` property to `CardVaultResult`
  * Add `CardAuthChallenge` type

 ### Checklist

 - [x] Added a changelog entry

### Screenshot

https://github.com/paypal/paypal-android/assets/58225613/5aa4df70-02c6-4848-a5e3-484f083a159c

- @sshropshire

### DIFF
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardAuthLauncher.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardAuthLauncher.kt
@@ -1,6 +1,5 @@
 package com.paypal.android.cardpayments
 
-import android.net.Uri
 import androidx.fragment.app.FragmentActivity
 import com.braintreepayments.api.BrowserSwitchClient
 import com.braintreepayments.api.BrowserSwitchException
@@ -16,31 +15,39 @@ internal class CardAuthLauncher(
 
     companion object {
         private const val METADATA_KEY_REQUEST_TYPE = "request_type"
-        private const val METADATA_KEY_ORDER_ID = "order_id"
-        private const val METADATA_KEY_SETUP_TOKEN_ID = "setup_token_id"
-
-        private const val REQUEST_TYPE_CHECKOUT = "checkout"
+        private const val REQUEST_TYPE_APPROVE_ORDER = "approve_order"
         private const val REQUEST_TYPE_VAULT = "vault"
 
-        private const val URL_PARAM_APPROVAL_SESSION_ID = "approval_session_id"
+        private const val METADATA_KEY_ORDER_ID = "order_id"
+        private const val METADATA_KEY_SETUP_TOKEN_ID = "setup_token_id"
     }
 
     fun presentAuthChallenge(
         activity: FragmentActivity,
         authChallenge: CardAuthChallenge
     ): PayPalSDKError? {
-        // preform 3DS browser switch
-        val vaultRequest = authChallenge.request
-        val returnUrlScheme = vaultRequest.run { Uri.parse(returnUrl).scheme!! }
-        val metadata = JSONObject()
-            .put(METADATA_KEY_REQUEST_TYPE, REQUEST_TYPE_VAULT)
-            .put(METADATA_KEY_SETUP_TOKEN_ID, vaultRequest.setupTokenId)
+        val metadata = when (authChallenge) {
+            is CardAuthChallenge.ApproveOrder -> {
+                val request = authChallenge.request
+                JSONObject()
+                    .put(METADATA_KEY_REQUEST_TYPE, REQUEST_TYPE_APPROVE_ORDER)
+                    .put(METADATA_KEY_ORDER_ID, request.orderId)
+            }
 
-        val options = BrowserSwitchOptions()
+            is CardAuthChallenge.Vault -> {
+                val request = authChallenge.request
+                JSONObject()
+                    .put(METADATA_KEY_REQUEST_TYPE, REQUEST_TYPE_VAULT)
+                    .put(METADATA_KEY_SETUP_TOKEN_ID, request.setupTokenId)
+            }
+        }
+
+        // launch the 3DS flow
+        val browserSwitchOptions = BrowserSwitchOptions()
             .url(authChallenge.url)
-            .returnUrlScheme(returnUrlScheme)
+            .returnUrlScheme(authChallenge.returnUrlScheme)
             .metadata(metadata)
-        return launchBrowserSwitch(activity, options)
+        return launchBrowserSwitch(activity, browserSwitchOptions)
     }
 
     private fun launchBrowserSwitch(
@@ -63,8 +70,8 @@ internal class CardAuthLauncher(
             if (requestType == REQUEST_TYPE_VAULT) {
                 parseVaultResult(browserSwitchResult)
             } else {
-                // TODO: migrate approve order to use auth challenge launcher pattern internally
-//                parseWebCheckoutResult(browserSwitchResult)
+                // Assume REQUEST_TYPE_APPROVE_ORDER
+                parseApproveOrderResult(browserSwitchResult)
             }
         }
 
@@ -74,6 +81,21 @@ internal class CardAuthLauncher(
             BrowserSwitchStatus.CANCELED -> CardStatus.VaultCanceled
             else -> null
         }
+
+    private fun parseApproveOrderResult(browserSwitchResult: BrowserSwitchResult): CardStatus? {
+        val orderId = browserSwitchResult.requestMetadata?.getString(METADATA_KEY_ORDER_ID)
+        return if (orderId == null) {
+            CardStatus.ApproveOrderError(CardError.unknownError)
+        } else {
+            when (browserSwitchResult.status) {
+                BrowserSwitchStatus.SUCCESS ->
+                    parseApproveOrderSuccessResult(browserSwitchResult, orderId)
+
+                BrowserSwitchStatus.CANCELED -> CardStatus.ApproveOrderCanceled(orderId)
+                else -> null
+            }
+        }
+    }
 
     private fun parseVaultSuccessResult(browserSwitchResult: BrowserSwitchResult): CardStatus {
         val deepLinkUrl = browserSwitchResult.deepLinkUrl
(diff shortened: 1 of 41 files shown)


CardPayments/src/main/java/com/paypal/android/cardpayments/CardAuthLauncher.kt
```diff
@@ -0,0 +1,153 @@
...
+        } else {
+            // TODO: see if there's a way that we can require the merchant to make their
+            // return and cancel urls conform to a strict schema
+
+            // NOTE: this assumes that when the merchant created a setup token, they used a
+            // return_url with word "success" in it (or a cancel_url with the word "cancel" in it)
+            val setupTokenId =
+                browserSwitchResult.requestMetadata?.optString(METADATA_KEY_SETUP_TOKEN_ID)
+            val deepLinkUrlString = deepLinkUrl.toString()
+            val didSucceed = deepLinkUrlString.contains("success")
```
sshropshire, line 118: This was the best way I could figure out if 3DS was successful. If we keep it this way, we will need to call out to merchants that their URL needs to have "success" in it.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardAuthLauncher.kt
```diff
@@ -0,0 +1,153 @@
...
+            if (didSucceed) {
+                val result = CardVaultResult(setupTokenId!!, "SCA_COMPLETE")
+                CardStatus.VaultSuccess(result)
+            } else {
+                val didCancel = deepLinkUrlString.contains("cancel")
```
sshropshire, line 123: Same for "cancel" to determine if 3DS was canceled.

CardPayments/src/test/java/com/paypal/android/cardpayments/CardAuthLauncherUnitTest.kt
```diff
@@ -0,0 +1,274 @@
...
+
+        every { browserSwitchClient.start(any(), any()) } throws browserSwitchException
+
+        val returnUrl = "merchant.app://return.com/deep-link"
+        val cardRequest = CardRequest("fake-order-id", card, returnUrl)
+        val cardAuthChallenge = CardAuthChallenge.ApproveOrder(url, cardRequest)
+
+        sut = CardAuthLauncher(browserSwitchClient)
+        val error = sut.presentAuthChallenge(activity, authChallenge = cardAuthChallenge)
+        assertEquals("error message from browser switch", error?.errorDescription)
```
sshropshire, line 62: ^ These tests were extracted from `CardClientUnitTest`. Eventually the launcher will be separated from the client in the next major version.

CardPayments/src/test/java/com/paypal/android/cardpayments/CardAuthLauncherUnitTest.kt
```diff
@@ -0,0 +1,274 @@
...
+        sut = CardAuthLauncher(browserSwitchClient)
+        val error = sut.presentAuthChallenge(activity, authChallenge = vaultAuthRequest)
+        assertNull(error)
+
+        val browserSwitchOptions = slot.captured
+        val metadata = browserSwitchOptions.metadata
+        assertEquals("vault", metadata?.getString("request_type"))
+        assertEquals("fake-setup-token-id", metadata?.getString("setup_token_id"))
+        assertEquals("merchant.app", browserSwitchOptions.returnUrlScheme)
+        assertEquals(Uri.parse("https://fake.com/destination"), browserSwitchOptions.url)
```
sshropshire, line 104: ^There are tests for both vault and approve order flows.

Demo/src/main/java/com/paypal/android/ui/paypalbuttons/PayPalButtonColorOptionList.kt
```diff
@@ -4,7 +4,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.res.stringResource
 import com.paypal.android.R
 import com.paypal.android.paymentbuttons.PayPalButtonColor
-import com.paypal.android.ui.OptionList
+import com.paypal.android.uishared.components.OptionList
```
sshropshire, line 7: Moving OptionList to the more appropriate `.uishared.components` package.

Demo/src/main/java/com/paypal/android/ui/vaultcard/VaultCardViewModel.kt
```diff
@@ -155,4 +183,24 @@ class VaultCardViewModel @Inject constructor(
             securityCode = uiState.cardSecurityCode
         )
     }
+
+    fun presentAuthChallenge(activity: FragmentActivity, authChallenge: CardAuthChallenge) {
+        authChallengeState = ActionState.Loading
+
+        // change listener behavior to handle auth result
```
sshropshire, line 191: ^ Re-defining the listener here since the success callback is used for both 3DS and non-3DS card vaulting flows.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -63,64 +60,56 @@ class CardClient internal constructor(
...
+        // TODO: deprecate this method and offer auth challenge integration pattern (similar to vault)
         orderId = cardRequest.orderId
         analyticsService.sendAnalyticsEvent("card-payments:3ds:started", orderId)
 
         CoroutineScope(dispatcher).launch(approveOrderExceptionHandler) {
-            confirmPaymentSource(activity, cardRequest)
-        }
-    }
-
-    private suspend fun confirmPaymentSource(activity: FragmentActivity, cardRequest: CardRequest) {
```
sshropshire, line 88: Had to inline this method to reduce the overall method count for this class to please the linter 🙌

CardPayments/src/main/java/com/paypal/android/cardpayments/CardAuthLauncher.kt
```diff
@@ -0,0 +1,153 @@
...
+
+    private fun parseVaultSuccessResult(browserSwitchResult: BrowserSwitchResult): CardStatus {
+        val deepLinkUrl = browserSwitchResult.deepLinkUrl
+        val requestMetadata = browserSwitchResult.requestMetadata
+
+        return if (deepLinkUrl == null || requestMetadata == null) {
+            CardStatus.VaultError(CardError.unknownError)
+        } else {
+            // TODO: see if there's a way that we can require the merchant to make their
+            // return and cancel urls conform to a strict schema
```
tdchow, line 111: Would this input validation have to happen on the web side where the return URL is defined?

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
tdchow: Would it be better to add an exception to the linter rule here if having this function separated increased readability? Or would splitting out some of this internal logic into a separate class make sense?

CardPayments/src/main/java/com/paypal/android/cardpayments/CardVaultRequest.kt
```diff
@@ -10,9 +10,12 @@ import kotlinx.parcelize.Parcelize
  *
  * @property setupTokenId id for the setup token to update.
  * @property card card payment source to attach to the setup token.
+ * @property returnUrl return url for deep linking back into the merchant app after an auth challenge
  */
 @Parcelize
 data class CardVaultRequest(
     val setupTokenId: String,
     val card: Card,
+    // NOTE: This needs to be null by default to prevent a breaking change
```
tdchow: This might be a good callout as a PR comment, do we want to keep it in as a code comment?

CardPayments/src/main/java/com/paypal/android/cardpayments/CardVaultResult.kt
```diff
@@ -1,18 +1,19 @@
...
  */
-@Parcelize
+// NEXT MAJOR VERSION: make `CardVaultResult` constructor private
 data class CardVaultResult(
     val setupTokenId: String,
-    val status: String
-) : Parcelable
+    val status: String,
+    // NOTE: This technically needs to null by default to prevent a breaking change
+    val authChallenge: CardAuthChallenge? = null
```
tdchow, line 17: If the SDK is creating the instances of `CardVaultResult`, would having `authChallenge` as non null be a breaking change. It might just add a new property to the result object that merchants could reference in the new SDK version.

tdchow (approved):
Changes look good! just had a few non blocking comments.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardAuthLauncher.kt
sshropshire: It might have to happen on the SDK side when we receive and parse the deep link response.

Ideally the `UpdateSetupToken` graphQL endpoint would allow us to specify a 3DS URL and we could entirely encapsulate URL formatting.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardVaultRequest.kt
sshropshire: That's a good call I can take it out.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardVaultResult.kt
sshropshire: Yeah it's only breaking since the constructor is public by default, though technically we could treat it as non-breaking since we don't expect merchants to ever instantiate these internal types.

Next major version we should mark constructors as internal where we can, or at least restrict them with `LIBRARY_GROUP` annotations to keep them available for merchant unit tests.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
sshropshire: I thought about this at the time. I'm honestly torn. The linter has been somewhat strict, but it does a good job of making us rethink the arch sometimes. This made me feel like maybe the `notify` methods at the bottom of this class could go somewhere else, like in a composable `CardAnalyticsClient` that uses an `AnalyticsService` reference internally.

saperi22 (approved):

CardPayments/src/main/java/com/paypal/android/cardpayments/CardVaultResult.kt
tdchow: Gotcha, if we treat it as a non-breaking change, could we make `CardAuthChallenge` non-null? This would prevent the merchant from having to check nullability.

This is maybe a larger question, if we make breaking changes on classes that should be treated as internal, do we need treat them as a major version breaking change?

sshropshire: Oh actually looking back at it my comment is incorrect, my bad. Semantically this value should be `null` when 3DS is not required. It's the default argument that's required to prevent a strict breaking change.

tdchow: Ah gotcha, all good then!

PR #245, opened 2024-03-11
Button analytics

saperi22:
### Summary of changes

 - DTBTSDK-3474
 - Adding analytics to buttons.

 ### Checklist

 - [x] Added a changelog entry

- @saperi22

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
   * Undeprecate `PayPalCreditButtonColor.BLACK` and `.DARK_BLUE`
   * Undeprecate `PayPalButtonColor.BLUE`, `.BLACK`, and `.SILVER`
   * Added analytics events
-    * `paypal-button:initialized` and `paypal-button:tapped`
+    * `payment-button:initialized` and `payment-button:tapped`
 * CardPayments
   * Add `liabilityShift` property to `CardResult`
   * Callback `PayPalSDKError` when `CardClient#approveOrder()` 3DS verification fails
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 * PaymentButtons
   * Undeprecate `PayPalCreditButtonColor.BLACK` and `.DARK_BLUE`
   * Undeprecate `PayPalButtonColor.BLUE`, `.BLACK`, and `.SILVER`
+  * Added analytics events
+    * `paypal-button:initialized` and `paypal-button:tapped`
 * CardPayments
   * Add `liabilityShift` property to `CardResult`
   * Callback `PayPalSDKError` when `CardClient#approveOrder()` 3DS verification fails
--- a/CorePayments/src/main/java/com/paypal/android/corepayments/TrackingEventsAPI.kt
+++ b/CorePayments/src/main/java/com/paypal/android/corepayments/TrackingEventsAPI.kt
@@ -46,6 +46,7 @@ internal class TrackingEventsAPI constructor(
             .put(KEY_MERCHANT_APP_VERSION, merchantAppVersion)
             .put(KEY_PLATFORM, "Android")
             .put(KEY_ORDER_ID, event.orderId)
+            .put(KEY_BUTTON_TYPE, event.buttonType)
             .put(KEY_TIMESTAMP, event.timestamp.toString())
             .put(KEY_TENANT_NAME, "PayPal")
 
@@ -71,6 +72,7 @@ internal class TrackingEventsAPI constructor(
         const val KEY_IS_SIMULATOR = "is_simulator"
         const val KEY_MERCHANT_APP_VERSION = "mapv"
         const val KEY_ORDER_ID = "order_id"
+        const val KEY_BUTTON_TYPE = "button_type"
         const val KEY_PLATFORM = "platform"
         const val KEY_TIMESTAMP = "t"
         const val KEY_TENANT_NAME = "tenant_name"
--- a/CorePayments/src/main/java/com/paypal/android/corepayments/TrackingEventsAPI.kt
+++ b/CorePayments/src/main/java/com/paypal/android/corepayments/TrackingEventsAPI.kt
@@ -45,8 +45,8 @@ internal class TrackingEventsAPI constructor(
             .put(KEY_IS_SIMULATOR, isSimulator)
             .put(KEY_MERCHANT_APP_VERSION, merchantAppVersion)
             .put(KEY_PLATFORM, "Android")
-            .put(KEY_ORDER_ID, event.orderId)
-            .put(KEY_BUTTON_TYPE, event.buttonType)
+            .putOpt(KEY_ORDER_ID, event.orderId)
+            .putOpt(KEY_BUTTON_TYPE, event.buttonType)
             .put(KEY_TIMESTAMP, event.timestamp.toString())
             .put(KEY_TENANT_NAME, "PayPal")
 
--- a/CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
+++ b/CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
@@ -5,5 +5,6 @@ internal data class AnalyticsEventData(
     val environment: String,
     val eventName: String,
     val timestamp: Long,
-    val orderId: String?
+    val orderId: String?,
+    val buttonType: String? = null
 )
--- a/PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PayLaterButton.kt
+++ b/PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PayLaterButton.kt
@@ -49,5 +49,9 @@ class PayLaterButton @JvmOverloads constructor(
 
     init {
         updateLabel(PayPalButtonLabel.PAY_LATER)
+        analyticsService.sendAnalyticsEvent(
+            "paypal-button:initialized",
+            PaymentButtonFundingType.PAY_LATER.toString().lowercase()
+        )
     }
 }
--- a/PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PayLaterButton.kt
+++ b/PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PayLaterButton.kt
@@ -51,7 +51,8 @@ class PayLaterButton @JvmOverloads constructor(
         updateLabel(PayPalButtonLabel.PAY_LATER)
         analyticsService.sendAnalyticsEvent(
             "paypal-button:initialized",
-            PaymentButtonFundingType.PAY_LATER.toString().lowercase()
+            orderId = null,
+            buttonType = PaymentButtonFundingType.PAY_LATER.toString().lowercase()
         )
     }
 }
--- a/PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PayLaterButton.kt
+++ b/PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PayLaterButton.kt
@@ -50,7 +50,7 @@ class PayLaterButton @JvmOverloads constructor(
     init {
         updateLabel(PayPalButtonLabel.PAY_LATER)
         analyticsService.sendAnalyticsEvent(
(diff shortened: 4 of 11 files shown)


PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PayLaterButton.kt
```diff
@@ -49,5 +49,10 @@ class PayLaterButton @JvmOverloads constructor(
 
     init {
         updateLabel(PayPalButtonLabel.PAY_LATER)
```
saperi22, line 67: As PayLaterButton extends PayPalButton, we'll have a duplicate initialized event logged.

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
```diff
@@ -5,5 +5,6 @@ internal data class AnalyticsEventData(
     val environment: String,
     val eventName: String,
     val timestamp: Long,
-    val orderId: String?
+    val orderId: String?,
+    val buttonType: String? = null
```
sshropshire, line 9: q: Can we put the button type into the event name? That way we wouldn't need an extra property.

saperi22: `AnalyticsEventData` is a local data holder that's eventually converted to event data https://github.com/paypal/paypal-android/pull/245/files/07dd656c767aed4d03e04c384ada2d099fb45eb0#diff-65c7e120a418bd43366952c069e36e07ede55b6ca30b1bfebfecbdafe3443db4R49.

I'm not sure there's an advantage to appending it to the event name as we'll have to parse it out (need to add more logic) and send it separately in the event data anyway. Please let me know if I'm missing something.

sshropshire: Got it ok. FPTI may have more features than I realize. For BT we would flatten everything to a string.

CorePayments/src/main/java/com/paypal/android/corepayments/TrackingEventsAPI.kt
```diff
@@ -46,6 +46,7 @@ internal class TrackingEventsAPI constructor(
             .put(KEY_MERCHANT_APP_VERSION, merchantAppVersion)
             .put(KEY_PLATFORM, "Android")
             .put(KEY_ORDER_ID, event.orderId)
+            .put(KEY_BUTTON_TYPE, event.buttonType)
```
sshropshire:
```suggestion
            .putOpt(KEY_BUTTON_TYPE, event.buttonType)
```
I'd go with [putOpt()](https://developer.android.com/reference/org/json/JSONObject#putOpt(java.lang.String,%20java.lang.Object)) here since the value can be `null`.

CorePayments/src/main/java/com/paypal/android/corepayments/TrackingEventsAPI.kt
```diff
@@ -71,6 +72,7 @@ internal class TrackingEventsAPI constructor(
         const val KEY_IS_SIMULATOR = "is_simulator"
         const val KEY_MERCHANT_APP_VERSION = "mapv"
         const val KEY_ORDER_ID = "order_id"
+        const val KEY_BUTTON_TYPE = "button_type"
```
sshropshire, line 75: We should update [TrackingEventsAPIUnitTest.kt](https://github.com/paypal/paypal-android/blob/main/CorePayments/src/test/java/com/paypal/android/corepayments/TrackingEventsAPIUnitTest.kt) to verify that this field is sent.

CorePayments/src/main/java/com/paypal/android/corepayments/TrackingEventsAPI.kt
saperi22: done!

saperi22: @sshropshire as `orderId` is also nullable, I made the change there too.

sshropshire (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 * PaymentButtons
   * Undeprecate `PayPalCreditButtonColor.BLACK` and `.DARK_BLUE`
   * Undeprecate `PayPalButtonColor.BLUE`, `.BLACK`, and `.SILVER`
+  * Added analytics events
```
sarahkoop, line 7: This is internal functionality not anything the merchant needs to be aware of right? I don't think we need a CHANGELOG if so

jaxdesmarais: I think we have tried to add them to the changelog just for our own knowledge of what version certain analytics were released in. Though I think we can be more specific here on what analytics we are now sending.

saperi22: I think iOS went with the changelog. I've updated my PR based on the discussions on iOS PR, to maintain parity.

@jaxdesmarais what kind of specificity are you wanting to see in the changelog?

scannillo: +1 to having a changelog entry so we can track what version we added analytics in for buttons module

CorePayments/src/main/java/com/paypal/android/corepayments/TrackingEventsAPI.kt
```diff
@@ -45,7 +45,8 @@ internal class TrackingEventsAPI constructor(
             .put(KEY_IS_SIMULATOR, isSimulator)
             .put(KEY_MERCHANT_APP_VERSION, merchantAppVersion)
             .put(KEY_PLATFORM, "Android")
-            .put(KEY_ORDER_ID, event.orderId)
+            .putOpt(KEY_ORDER_ID, event.orderId)
+            .putOpt(KEY_BUTTON_TYPE, event.buttonType)
```
scannillo, line 49: Did we update our FPTI tags doc (internal on Excel/Sharepoint) to include this new param as well as the 3 possible string values for each button type?

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
jaxdesmarais: I think generally our approach with FPTI has been to keep the event names streamlined and add any extra data as a tag. This makes it easier to calculate conversion of a feature as a whole (counting all button conversion as a whole) as well as drill into the data (via tags) for the difference in conversion of things like paypal vs pay later vs credit. This prevents us from needing to do more math by adding slightly different event names together to calculate a whole feature.

CHANGELOG.md
jaxdesmarais: I think the update you made in [this commit](https://github.com/paypal/paypal-android/pull/245/commits/e079d00d63827867e99a779cc1508e1500b852d5) is in line with the specificity that I was thinking! Thanks for making that change.

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PayLaterButton.kt
jaxdesmarais: 🤔 is there a way to avoid this? I think we want to avoid sending duplicate events.

jaxdesmarais: If all of the button classes extend `PayPalButton` could we instead have this initialization event just sent there and use the `fundingType` property passed into the `buttonType`?

saperi22: > If all of the button classes extend `PayPalButton` could we instead have this initialization event just sent there and use the `fundingType` property passed into the `buttonType`?

I had initially put the analytics block in the `PaymentButton` - which is the super class of all these views. The issue I ran into and learned about was the execution order in which Kotlin executes various `constructors`, `init` blocks and the class level variables.
See: [An in-depth look at Kotlin’s initializers](https://medium.com/keepsafe-engineering/an-in-depth-look-at-kotlins-initializers-a0420fcbf546#:~:text=Execution%20order&text=First%2C%20default%20constructor%20arguments%20are,%2C%20top%2Dto%2Dbottom)

I've spent some time on it and wasn't able to figure out a way to make it possible.
I wouldn't mind pairing on this to figure out if it's possible.

Simply put, `fundingType` isn't available until _after_ all the variables in the children are all executed, which happens _after_ all of the parents blocks execution.

I _could_ probably wait for 1s (or a lower trivial amount of time) for all the initializations to occur and then fire off the analytics event, but it felt a bit far-fetched for the use-case. Let me know if you think this approach would be better.

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
saperi22: @jaxdesmarais to be clear, you want the `button_type` to be a param/tag and not concatenate `button_type` with the event name, correct?

CorePayments/src/main/java/com/paypal/android/corepayments/TrackingEventsAPI.kt
saperi22: Thanks for the reminder, I just updated it.

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
jaxdesmarais: Yep, how you have it, was just explaining a bit more of the reasoning to @sshropshire of why we have moved to utilizing tags vs throwing everything in the event name like we did for previous version of BT since I had the context. 😄

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PayLaterButton.kt
jaxdesmarais: 🤔 that's super interesting and thanks for doing that research/laying it out in a way that's easy to understand! Is there anything specific that we have to have from extending `PayPalPalButton` or can we duplicate some code/extend `PaymentButton` (similar to what it looks like the `PayPalCreditButton` does). Duplicating events for 1 payment button will pretty greatly skew our data. What do you think?

sshropshire: I've always felt like we should have one button class. Karthik felt the same way when he was here. I wonder if the button analytics work is starting to reveal tech debt?

saperi22: > 🤔 that's super interesting and thanks for doing that research/laying it out in a way that's easy to understand! Is there anything specific that we have to have from extending `PayPalPalButton` or can we duplicate some code/extend `PaymentButton` (similar to what it looks like the `PayPalCreditButton` does). Duplicating events for 1 payment button will pretty greatly skew our data. What do you think?

Duplicating logic should be doable.

>I've always felt like we should have one button class. Karthik felt the same way when he was here. I wonder if the button analytics work is starting to reveal tech debt?

@sshropshire there was a ticket to convert everything to a single button class.
1. it is a breaking change.
2. we'll likely run into issues with Android framework liking its constructors to not have any extra arguments. I've tried taking a stab at it, but I didn't like the outcome.

I feel like it might be ok to consolidate all buttons into a single one in compose.

jaxdesmarais: > Duplicating logic should be doable.

This to me makes sense to tackle in the scope of this PR since currently we'd be sending duplicate events. What do you think?

saperi22: > > Duplicating logic should be doable.
> 
> This to me makes sense to tackle in the scope of this PR since currently we'd be sending duplicate events. What do you think?

made the change :) Please take another look.

jaxdesmarais (approved):
Thanks for addressing my questions as well as tacking the duplicate events (as well as calling out that issue)!

PR #252, opened 2024-04-16
Google Play store compliance fix

saperi22:
### Summary of changes

 - Update `magnessdk` to 5.5.1.
 - Update MXO to 1.3.2.

 ### Checklist

 - [x] Added a changelog entry

- @saperi22


PayPalNativePayments/src/main/java/com/paypal/android/paypalnativepayments/PayPalNativeCheckoutRequest.kt
```diff
@@ -13,5 +13,14 @@ data class PayPalNativeCheckoutRequest(
...
+    val userAuthenticationEmail: String? = null,
+
+    /**
+     * @property [hasUserLocationConsent] informs the SDK if your application has obtained
+     * consent from the user to collect location data in compliance with
+     * <a href="https://support.google.com/googleplay/android-developer/answer/10144311#personal-sensitive">
+     * Google Play Developer Program policies</a>
+     * This flag enables PayPal to collect necessary information required for Fraud Detection and Risk Management.
+     */
+    val hasUserLocationConsent: Boolean = false
```
tdchow, line 25: I think we'll want to eventually make this a required field. I'm not sure if there's a way to deprecate the constructor that does not pass in this boolean 🤔

saperi22: As we'd want a non-breaking change, we provide a default value.
For the next version, we'd want to remove the default value so the merchant developer has to pass in the value for `hasUserLocationConsent`.

Because the request object is the same, we don't have to change the constructor itself.

I'm not sure if I've 'answered' what you are asking. :D

tdchow: Yep! we'd definitely want to keep the default value for this minor version update. But I'm wondering if there's a way we can deprecate the constructor that does not include the `hasUserLocationConsent` just so merchants have a heads up before the property is removed.

tdchow (approved):

PayPalNativePayments/src/main/java/com/paypal/android/paypalnativepayments/PayPalNativeCheckoutRequest.kt
saperi22: >But I'm wondering if there's a way we can deprecate the constructor that does not include the hasUserLocationConsent just so merchants have a heads up before the property is removed.

I'm not sure I understand this statement.

>the constructor that does not include the hasUserLocationConsent

can you please point me to what you are referring to ^?

The constructor `startCheckout(request: PayPalNativeCheckoutRequest)` accepts a request object. I'm adding the boolean to the request object instead of the constructor itself.

tdchow: Oh i was referring to the `PayPalNativeCheckoutRequest` constructor, not the `startCheckout()` function.

I was wondering if there was a way to flag when a merchant creates a `PayPalNativeCheckoutRequest` instance without passing `hasUserLocationConsent`.

saperi22: I had a brain fart. :/

build.gradle
```diff
@@ -77,7 +77,7 @@ buildscript {
             "json"                        : "org.json:json:20220320",
 
             // PayPal
-            "nativeCheckout"              : "com.paypal.checkout:android-sdk:1.2.1",
+            "nativeCheckout"              : "com.paypal.checkout:android-sdk:1.3.2",
```
agedd, line 80: qq here - should this be version 1.3.0, that's the latest release i'm seeing [here](https://github.paypal.com/nativexo/android-native-checkout/releases/tag/external-v1.3.0)
```suggestion
            "nativeCheckout"              : "com.paypal.checkout:android-sdk:1.3.0",
```

CHANGELOG.md
```diff
@@ -15,10 +15,14 @@
...
 * FraudDetection
   * Fixes Google Play Store Rejection
-    * Bump Magnes version to 5.5.0
+    * Bump Magnes version to 5.5.1
     * Create `PayPalDataCollectorRequest`
     * Add `PayPalDataCollector#collectDeviceData(context, request)`
     * Deprecate `PayPalDataCollector#collectDeviceData(context, clientMetadataId, additionalData)`
+* PayPalNativePayments
+  * Fixes Google Play Store Rejection
+    * Bump Native Checkout version to 1.3.2
```
agedd, line 24: nit: same comment as below
```suggestion
    * Bump Native Checkout version to 1.3.0
```

build.gradle
saperi22: hmmm.. great catch.
I feel the latest version might not have been tagged in the releases.

@tdchow please correct me if I'm wrong, but I believe 1.3.2 is the latest that we had to consume, correct?

tdchow: Yep, it looks like the latest versions have not been tagged in the GitHub releases. The correct version to use is 1.3.2.

https://repo1.maven.org/maven2/com/paypal/checkout/android-sdk/1.3.2/

agedd (approved):

KunJeongPark (approved):

PR #256, opened 2024-04-29
Update demo app Clients from lateinit vars to nullable vars

tdchow:
### Summary of changes

- Convert Clients from a lateinit var to a nullable var in the demo app for card and PayPal web flows

There were crashes in the demo app where lateinit Clients were not getting set either due to a network error or if the activity was cleared without setting the client. These crashes were introduced when the removeObservers() function was added to the clients.

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
+++ b/Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
@@ -47,7 +47,7 @@ class ApproveOrderViewModel @Inject constructor(
     private val _uiState = MutableStateFlow(ApproveOrderUiState())
     val uiState = _uiState.asStateFlow()
 
-    private lateinit var cardClient: CardClient
+    private var cardClient: CardClient? = null
     private lateinit var payPalDataCollector: PayPalDataCollector
 
     fun createOrder() {
@@ -85,7 +85,7 @@ class ApproveOrderViewModel @Inject constructor(
                 payPalDataCollector = PayPalDataCollector(coreConfig)
 
                 cardClient = CardClient(activity, coreConfig)
-                cardClient.approveOrderListener = object : ApproveOrderListener {
+                cardClient?.approveOrderListener = object : ApproveOrderListener {
                     override fun onApproveOrderSuccess(result: CardResult) {
                         approveOrderState = ActionState.Success(result)
                     }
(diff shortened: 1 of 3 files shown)


sarahkoop:
These errors are specific to our demo app integration right? I don't quite understand what in the `removeObservers` method would cause clients to be set to null - but wondering if this would impact merchant integrations too

tdchow:
> These errors are specific to our demo app integration right? I don't quite understand what in the `removeObservers` method would cause clients to be set to null - but wondering if this would impact merchant integrations too

Yep, these are crashes due to how the demo app's ViewModels declare the client instances as `lateinit`. If the ViewModel never sets the Client and the Activity is cleared, `kotlin.UninitializedPropertyAccessException: lateinit property cardClient has not been initialized` is thrown, when `removeObservers()` is called on an uninitialized client object.

If merchants followed this pattern in their integration, they'd also need to convert from a lateinit var to a nullable var.

sarahkoop:
> > These errors are specific to our demo app integration right? I don't quite understand what in the `removeObservers` method would cause clients to be set to null - but wondering if this would impact merchant integrations too
> 
> Yep, these are crashes due to how the demo app's ViewModels declare the client instances as `lateinit`. If the ViewModel never sets the Client and the Activity is cleared, `kotlin.UninitializedPropertyAccessException: lateinit property cardClient has not been initialized` is thrown, when `removeObservers()` is called on an uninitialized client object.
> 
> If merchants followed this pattern in their integration, they'd also need to convert from a lateinit var to a nullable var.

Should we add more docs to the `removeObservers` method to call this out? This only applies if merchants start using that method correct? (trying to confirm it's not a breaking change)

saperi22 (approved):

tdchow:
> > > These errors are specific to our demo app integration right? I don't quite understand what in the `removeObservers` method would cause clients to be set to null - but wondering if this would impact merchant integrations too
> > 
> > 
> > Yep, these are crashes due to how the demo app's ViewModels declare the client instances as `lateinit`. If the ViewModel never sets the Client and the Activity is cleared, `kotlin.UninitializedPropertyAccessException: lateinit property cardClient has not been initialized` is thrown, when `removeObservers()` is called on an uninitialized client object.
> > If merchants followed this pattern in their integration, they'd also need to convert from a lateinit var to a nullable var.
> 
> Should we add more docs to the `removeObservers` method to call this out? This only applies if merchants start using that method correct? (trying to confirm it's not a breaking change)

Yeah, I think we can call this out in the devdocs. Have we already made initial updates to the devdocs to add this function?

I wouldn't call this a breaking change since the merchant owns the instantiation and management of the client instances.

sarahkoop (approved):

Hwdp3524 (approved):

Hwdp3524 (approved):

PR #258, opened 2024-05-20
Provide ways to define PayLaterButton color from xml

saperi22:
### Summary of changes

Addresses a Live Issue reported by a merchant.
 - Fix issue with `PayLaterButton` not honoring `paypal_color` attribute in xml.
 - `PayLaterButton` used to extend from `PayPalButton`, but now it extends from `PaymentButton`. I'm providing a new attribute `paylater_color` to change the button color in xml.

 ### Checklist

 - [x] Added a changelog entry

- @saperi22 

### Screenshots

Screenshot 2024-05-23 at 9 36 08 AM
Screenshot 2024-05-23 at 9 36 14 AM

### DIFF
--- a/Demo/src/main/res/layout/pay_later_button_test_layout.xml
+++ b/Demo/src/main/res/layout/pay_later_button_test_layout.xml
@@ -1,7 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<!-- XML view for testing only. These aren't used anywhere else in the demo application. -->
-
 <androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
--- a/Demo/src/main/res/layout/pay_later_button_test_layout.xml
+++ b/Demo/src/main/res/layout/pay_later_button_test_layout.xml
@@ -1,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+
+    <com.paypal.android.paymentbuttons.PayLaterButton
+        android:id="@+id/button1"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_margin="8dp"
+        android:padding="8dp"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:paylater_color="white"
+        />
+
+    <com.paypal.android.paymentbuttons.PayLaterButton
+        android:id="@+id/button2"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/button1"
+        android:layout_margin="8dp"
+        android:padding="8dp"
+        app:paylater_color="gold"
+        />
+
+    <com.paypal.android.paymentbuttons.PayLaterButton
+        android:id="@+id/button3"
+        android:layout_width="wrap_content"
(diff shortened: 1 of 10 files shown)


sarahkoop:
@saperi22 Do you have more context on why the button changed from extending `PayPalButton` to extending `PaymentButton`?

saperi22:
@sarahkoop context here: https://github.com/paypal/paypal-android/pull/245/files#r1521937925
TL;DR a change made for button analytics.

Demo/src/main/res/layout/pay_later_button_test_layout.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!-- XML view for testing only. These aren't used anywhere else in the demo application. -->
```
tdchow: Should we add another page in the demo app showing all the different color variations? I'm not sure if our documentation has all of the variants. It might make it easier for merchants to decide on which color they should use in their app.

Demo/src/main/res/layout/pay_later_button_test_layout.xml
```diff
@@ -0,0 +1,66 @@
...
+    <com.paypal.android.paymentbuttons.PayLaterButton
+        android:id="@+id/button1"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_margin="8dp"
+        android:padding="8dp"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:paylater_color="white"
+        />
```
tdchow:
```suggestion
        app:paylater_color="white" />
```

Nit - I'm not sure if we have any custom styling for XML files, but I think the default formatting doesn't have the closing tag on a new line.

tdchow (approved):

Demo/src/main/res/layout/pay_later_button_test_layout.xml
saperi22: 🤷 I do it because when I want to delete the last line of text, I can do a `cmd+del` and that would still leave the closing tag intact.
I'm trying to move away from thinking too much about formatting, so if you feel strongly about the tag on the newline, I'll be happy to change it back.

tdchow: I see! That reasoning makes a lot of sense. I'm just worried about folks who will auto format the entire file. Should we make a modification to the format preference so Android Studio picks up this change?

Should we also follow this pattern for XML tags that have separate open and close tags?

saperi22: All good questions. I don't have a preference, tbh.
I've auto-formatted the whole file, and now all the tags are on the same line as the last one.

If there's a common setting (not on individual IDEs) that can take care of this formatting for us, I'm up for it.

tdchow: Screenshot 2024-05-22 at 9 37 01 AM

I think enabling this setting in Android Studio will have auto format add that extra new line. We could check in a format file, but everyone would need to manually apply the file in their IDE 🙁

saperi22: I've changed it to the regular expected formatting. We can look at this later. Not an important one :)

saperi22: Good to know about the setting though. :)

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PayLaterButton.kt
```diff
@@ -72,6 +75,14 @@ class PayLaterButton @JvmOverloads constructor(
         )
     }
 
+    private fun updateColorFrom(typedArray: TypedArray) {
```
sarahkoop, line 78: Not to expand the scope of this PR, but does this same bug exist for the `PayPalCreditButton` and `PayPalButton`?

saperi22: No, those didn't have changes.
The reason this broke is because of the change in inheritance.

sarahkoop (approved):

PR #262, opened 2024-06-27
Remove PayPalVaultRequest required URL param

scannillo:
### Reason for changes

DTMOBILES-681

- Currently, a merchant has to provide the SDK a URL to launch for the PP Vault w/o Purchase flow. This is not a great dev-ex since the SDK should provide more value and know which URL to launch.
- The JS SDK hard codes the URLs to launch for this experience, so the mobile SDKs can follow suit.

### Summary of changes

- Deprecate `PayPalWebVaultRequest(setupTokenId, approveVaultHref)`
- Add `PayPalWebVaultRequest(setupTokenId)`
- Add logic to SDK to construct vault w/o purchase URL itself

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebVaultRequest.kt
+++ b/PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebVaultRequest.kt
@@ -17,5 +17,5 @@ constructor(
      *
      * @property [setupTokenId] ID for the setup token associated with the vault approval
      */
-    constructor(setupTokenId: String): this(setupTokenId, null)
+    constructor(setupTokenId: String) : this(setupTokenId, null)
 }
--- a/PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebVaultRequest.kt
+++ b/PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebVaultRequest.kt
@@ -6,10 +6,11 @@ package com.paypal.android.paypalwebpayments
  * @property [setupTokenId] ID for the setup token associated with the vault approval
  * @property [approveVaultHref] URL for the approval web page
  */
-data class PayPalWebVaultRequest @Deprecated("Use PayPalWebVaultRequest(setupTokenId:) instead.")
+data class PayPalWebVaultRequest @Deprecated("Use PayPalWebVaultRequest(setupTokenId) instead.")
 constructor(
     val setupTokenId: String,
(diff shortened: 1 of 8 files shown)


PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebVaultRequest.kt
```diff
@@ -6,7 +6,16 @@ package com.paypal.android.paypalwebpayments
  * @property [setupTokenId] ID for the setup token associated with the vault approval
  * @property [approveVaultHref] URL for the approval web page
  */
-data class PayPalWebVaultRequest(
+data class PayPalWebVaultRequest @Deprecated("Use PayPalWebVaultRequest(setupTokenId:) instead.")
+constructor(
     val setupTokenId: String,
-    val approveVaultHref: String
-)
+    val approveVaultHref: String?? // NEXT_MAJOR_VERSION: - Remove this property
```
sshropshire:
```suggestion
    val approveVaultHref: String? // NEXT_MAJOR_VERSION: - Remove this property
```

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebVaultRequest.kt
```diff
@@ -6,7 +6,16 @@ package com.paypal.android.paypalwebpayments
  * @property [setupTokenId] ID for the setup token associated with the vault approval
  * @property [approveVaultHref] URL for the approval web page
  */
-data class PayPalWebVaultRequest(
+data class PayPalWebVaultRequest @Deprecated("Use PayPalWebVaultRequest(setupTokenId:) instead.")
```
sshropshire:
```suggestion
data class PayPalWebVaultRequest @Deprecated("Use PayPalWebVaultRequest(setupTokenId) instead.")
```
nit: Removing the colon. There's no set format that I can find for referencing the alternative constructor, but technically it'd be more like `PayPalWebVaultRequest(setupTokenId=)` in Kotlin.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebVaultRequest.kt
sshropshire: We may be able to mark the property as deprecated as well. I'm not sure if that's useful to merchants, but it couldn't hurt if it does work.

sshropshire (approved):

agedd (approved):
lgtm!! 🚀

KunJeongPark (approved):

PR #266, opened 2024-07-03
Revert 3DS URL Parsing for CardResult

sshropshire:
### Summary of changes

* Reference iOS PR: https://github.com/paypal/paypal-ios/pull/254
* CardPayments
  * Add `status` property to `CardResult`
  * Add `didAttemptThreeDSecureAuthentication` property to `CardResult`
* Breaking Changes
  * CardPayments
    * Remove `liabilityShift` property from `CardResult`

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # PayPal Android SDK Release Notes
 
+## unreleased
+
+* CardPayments
+  * Deprecate `CardResult.liabilityShift` property
+  * Add `CardResult.status` property
+  * Add `CardResult.didAttemptThreeDSecureAuthentication` property
+
 ## 1.5.0 (2024-07-10)
 * PayPalWebPayments
   * Deprecate `PayPalWebVaultRequest(setupTokenId, approveVaultHref)`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * CardPayments
   * Deprecate `CardResult.liabilityShift` property
+  * Deprecate `CardResult.deepLinkUrl` property
   * Add `CardResult.status` property
   * Add `CardResult.didAttemptThreeDSecureAuthentication` property
 
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
@@ -92,7 +92,11 @@ class CardClient internal constructor(
                 )
 
                 if (response.payerActionHref == null) {
-                    val result = CardResult(response.orderId)
+                    val result = CardResult(
+                        orderId = response.orderId,
+                        status = response.status?.name,
+                        didAttemptThreeDSecureAuthentication = false
+                    )
                     notifyApproveOrderSuccess(result)
                 } else {
                     analyticsService.sendAnalyticsEvent(
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardResult.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardResult.kt
@@ -1,7 +1,5 @@
 package com.paypal.android.cardpayments
 
-import android.net.Uri
-
 /**
  * A result returned by [CardClient] when an order was successfully approved with a [Card].
  *
@@ -14,14 +12,10 @@ import android.net.Uri
 data class CardResult(
     val orderId: String,
 
-    /**
-     * @suppress
-     */
-    val deepLinkUrl: Uri? = null,
-
     @Deprecated("Use didAttemptThreeDSecureAuthentication instead.")
     val liabilityShift: String? = null,
 
     var status: String? = null,
+
     val didAttemptThreeDSecureAuthentication: Boolean = false
 )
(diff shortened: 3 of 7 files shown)


CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -92,7 +92,11 @@ class CardClient internal constructor(
                 )
 
                 if (response.payerActionHref == null) {
-                    val result = CardResult(response.orderId)
+                    val result = CardResult(
+                        orderId = response.orderId,
+                        status = response.status?.name,
```
KunJeongPark, line 97: What is the .name Steven? I thought response.status was a String? Oh I'm thinking of CardResult, not response. Got it now.

KunJeongPark (approved):
Great job Steven!

tdchow (commented):
Would we need to wait until the next major version to introduce a breaking change? Could we mark `liabilityShift` as deprecated for this PR?

sshropshire:
> Would we need to wait until the next major version to introduce a breaking change? Could we mark `liabilityShift` as deprecated for this PR?

@tdchow It's also marked as breaking on iOS. We do have some things that we should break on Android as well. I'm not sure what the path to `v2` looks like for PPCP, but we will eventually need to get there.

sarahkoop:
> > Would we need to wait until the next major version to introduce a breaking change? Could we mark `liabilityShift` as deprecated for this PR?
> 
> @tdchow It's also marked as breaking on iOS. We do have some things that we should break on Android as well. I'm not sure what the path to `v2` looks like for PPCP, but we will eventually need to get there.

Can we separate the breaking changes part of this PR to a separate PR against a `v2` branch? I agree we should start planning for a major version, but I don't think we want removing liability shift to be the only breaking change in that version, and we are planning releases this sprint, so would be good to keep `main` in a release ready state

sshropshire:
@tdchow @sarahkoop I'll remove the breaking change. I see the difference with iOS now.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
sshropshire: Yeah the `name` property is built in to [Kotlin enums](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-enum/name.html).

CardPayments/src/main/java/com/paypal/android/cardpayments/CardResult.kt
```diff
@@ -1,19 +1,21 @@
...
+ * @property [didAttemptThreeDSecureAuthentication] 3DS verification was attempted.
+ * Use v2/checkout/orders/{orderId} in your server to get verification results.
  */
 data class CardResult(
     val orderId: String,
 
-    /**
-     * @suppress
-     */
-    val deepLinkUrl: Uri? = null,
```
sarahkoop, line 17: This is technically breaking too right - can we deprecate this instead?

sshropshire: We could yeah.

sshropshire: Updated: [933fe41](https://github.com/paypal/paypal-android/pull/266/commits/933fe41b6a67716209918f5bd870fb949ddd61f5)

CHANGELOG.md
```diff
@@ -1,9 +1,14 @@
 # PayPal Android SDK Release Notes
 
 ## unreleased
+
 * PayPalNativePayments (DEPRECATED)
   * **NOTE:** This module is being deprecated and will be removed in the future version of the SDK. Use `PayPalWebPayments` module instead.
   * Add deprecated warning message to all classes
+* CardPayments
+  * Deprecate `CardResult.liabilityShift` property
```
sarahkoop, line 6:
```suggestion
  * Deprecate `CardResult.liabilityShift` property
  * Deprecate `CardResult.deepLinkUrl` property
```

sarahkoop (approved):

CardPayments/src/main/java/com/paypal/android/cardpayments/CardResult.kt
```diff
@@ -7,13 +7,23 @@ import android.net.Uri
...
      * @suppress
      */
+    @Deprecated("Use status instead.")
     val deepLinkUrl: Uri? = null,
-    val liabilityShift: String? = null
+
+    @Deprecated("Use didAttemptThreeDSecureAuthentication instead.")
+    val liabilityShift: String? = null,
+
+    var status: String? = null,
```
jaxdesmarais: Basic q but is there a reason we use a `var` here and a `val` for all the other ones?

sshropshire: That's a good callout. It would make more sense if it was `val` to me. I don't think there's a good reason for it to be a mutable property.

sshropshire: Fix: [9271cab](https://github.com/paypal/paypal-android/pull/266/commits/9271cabeccc823ae12faac79e7447cc775e3cd61)

PR #268, opened 2024-07-19
CorePayments: EligibilityClient

sshropshire:
### Summary of changes

This PR adds the Eligibility Check feature to the SDK and Demo app.

* Add `EligibilityClient` to project
* Move `OrderIntent` from `:Demo` module to `:CorePayments` module to make it merchant facing
* Add `CheckEligibilityResultListener` interface
* Add `EligibilityClient`
* Add `EligibilityRequest`
* Add `EligibilityResult`

Equivalent iOS PR: https://github.com/paypal/paypal-ios/pull/286

 ### Checklist

 - [x] Added a changelog entry

### Screenshot

https://github.com/user-attachments/assets/bb69da35-9714-42a2-a211-5e03e46057e7

- @sshropshire


CorePayments/src/main/java/com/paypal/android/corepayments/apis/eligibility/SupportedCountryCurrencyType.kt
```diff
@@ -0,0 +1,5 @@
+package com.paypal.android.corepayments.apis.eligibility
+
+internal enum class SupportedCountryCurrencyType {
```
sshropshire: Random thought: Should we make this `enum` merchant facing to prevent unsupported currency codes from being used?

CorePayments/src/main/java/com/paypal/android/corepayments/apis/eligibility/Eligibility.kt
```diff
@@ -0,0 +1,9 @@
+package com.paypal.android.corepayments.apis.eligibility
+
+internal data class Eligibility(
+    val isVenmoEligible: Boolean,
+    val isPaypalEligible: Boolean,
+    val isPaypalCreditEligible: Boolean,
```
jaxdesmarais:
```suggestion
    val isPayPalEligible: Boolean,
    val isPayPalCreditEligible: Boolean,
```

CorePayments/src/main/java/com/paypal/android/corepayments/apis/eligibility/FundingEligibilityIntent.kt
```diff
@@ -0,0 +1,5 @@
+package com.paypal.android.corepayments.apis.eligibility
+
+internal enum class FundingEligibilityIntent {
+    SALE, CAPTURE, AUTHORIZE, ORDER, TOKENIZE, SUBSCRIPTION
```
jaxdesmarais: Update to match the iOS PR:
```suggestion
    CAPTURE, AUTHORIZE
```

CorePayments/src/main/java/com/paypal/android/corepayments/apis/eligibility/SupportedCountryCurrencyType.kt
jaxdesmarais: It looks like in the iOS PR we take this in as a string. How many supported currencies are there? I guess also how often do we expect them to change. If fairly often we may want to just have a public string for this like on iOS.

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/EligibilityClient.kt
```diff
@@ -0,0 +1,53 @@
...
+     */
+    fun check(
+        eligibilityRequest: EligibilityRequest,
+        callback: CheckEligibilityResultListener
+    ) {
+        CoroutineScope(dispatcher).launch {
+            try {
+                val response =
+                    eligibilityAPI.checkEligibility(applicationContext, eligibilityRequest)
+                callback.onCheckEligibilitySuccess(EligibilityResult(response.isVenmoEligible))
```
jaxdesmarais: Do we want to return all of the results here instead of just venmo?

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/EligibilityRequest.kt
```diff
@@ -0,0 +1,12 @@
+package com.paypal.android.corepayments.features.eligibility
+
+import com.paypal.android.corepayments.OrderIntent
+
+/**
+ * @property [intent] An order intent used in the request for eligibility.
+ * @property [currencyCode] A currency code used in the request for eligibility.
+ */
+data class EligibilityRequest(
+    val intent: OrderIntent,
```
jaxdesmarais, line 10: Do we want to call this `EligibilityIntent` to align with iOS?

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/EligibilityResult.kt
```diff
@@ -0,0 +1,6 @@
+package com.paypal.android.corepayments.features.eligibility
+
+/**
+ * Contains the result of a successful [EligibilityClient.check] response.
+ */
+data class EligibilityResult(val isVenmoEligible: Boolean)
```
jaxdesmarais: In the iOS PR it looks like we return all of the eligibilities (card, paypal, pay later, paypal credit) - not sure if we want to do the same here or update iOS to just return venmo

CorePayments/src/main/java/com/paypal/android/corepayments/apis/eligibility/EligibilityAPI.kt
```diff
@@ -0,0 +1,65 @@
...
+    private val resourceLoader: ResourceLoader
+) {
+    companion object {
+        const val VARIABLE_CLIENT_ID = "clientId"
+        const val VARIABLE_INTENT = "intent"
+        const val VARIABLE_CURRENCY = "currency"
+        const val VARIABLE_ENABLE_FUNDING = "enableFunding"
+    }
+
+    constructor(config: CoreConfig) :
```
tdchow: Could we combine this constructor with the main constructor by adding default values?

CorePayments/src/main/java/com/paypal/android/corepayments/apis/eligibility/SupportedCountryCurrencyType.kt
tdchow: This might be a good product question. How do we want to handle adding additional currencies? Are we okay requiring an SDK update for additional currency support, or can we enable it via a serverside change?

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/CheckEligibilityResultListener.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Implement this interface to receive results from calls to [EligibilityClient.check].
+ */
+interface CheckEligibilityResultListener {
+
+    /**
+     * Called when [EligibilityClient.check] succeeds.
+     *
+     * @param result The [EligibilityResult].
+     */
+    fun onCheckEligibilitySuccess(result: EligibilityResult)
```
tdchow: Not sure if PPCP has this pattern, but do we want to have a single function return a sealed class result?

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/EligibilityClient.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.paypal.android.corepayments.apis.eligibility.EligibilityAPI
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+
+/**
+ *  Use EligibilityClient to check eligibility for different payment methods.
+ */
+class EligibilityClient internal constructor(
```
tdchow, line 15: Can we add unit test coverage for this class?

tdchow (commented):
screen-20240723-111711.mp4 was added to the root directory. do we need this screen recording checked in?

CorePayments/src/main/java/com/paypal/android/corepayments/apis/eligibility/SupportedCountryCurrencyType.kt
sshropshire: Yeah I could see strengths in both String and Enum-based approaches. We may be able to let the GraphQL call fail when an unsupported currency string is provided. On the other hand, an enum may be helpful to let merchants discover a list of supported currencies through auto-completion instead of having to look them up on Google, etc.

sshropshire:
> screen-20240723-111711.mp4 was added to the root directory. do we need this screen recording checked in?

@tdchow I missed that earlier. I'll make sure to remove.

CorePayments/src/main/java/com/paypal/android/corepayments/apis/eligibility/EligibilityAPI.kt
sshropshire: Yeah def. I'll update.

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/CheckEligibilityResultListener.kt
sshropshire: Yeah that's one of the first breaking changes we'll have to make in `v2`. The current pattern in `v1` is outdated and should change. For now, this makes since to keep consistent with the other listeners.

This pattern made since initially, but it gets cumbersome when browser switching is involved e.g. [ApproveOrderListener.kt](https://github.com/paypal/paypal-android/blob/7285f90d939a7896fbfc868a2390deb8202c1059/CardPayments/src/main/java/com/paypal/android/cardpayments/ApproveOrderListener.kt#L9).

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/EligibilityClient.kt
sshropshire: Oh yeah I forgot to write the unit test for this one. I'll add it in another commit.

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/EligibilityClient.kt
sshropshire: Yeah I noticed iOS does this. I'm down to add it in. We should check to see if requesting eligibility for multiple payment methods affects latency though i.e. if it's quicker to ask for Venmo eligibility by itself rather than fetching eligibility for the other payment methods.

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/EligibilityRequest.kt
sshropshire: That's an interesting callout I wonder if iOS may want to rename to `OrderIntent`? The `intent` property has meaning outside of Eligibility–I believe it's tied to the Order domain.

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/EligibilityResult.kt
sshropshire: Yeah I'll update it to change. I wonder if there's a performance penalty for requesting eligibility for all at once. A lot of the performance characteristics of the call will depend on how the GraphQL layer resolves this information.

CorePayments/src/main/java/com/paypal/android/corepayments/apis/eligibility/SupportedCountryCurrencyType.kt
sshropshire: Went ahead and removed `SupportedCountryCurrencyType` here: [1cbfa70](https://github.com/paypal/paypal-android/pull/268/commits/1cbfa7018c4bb85767f29a77028016024271aaa8)

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/EligibilityClient.kt
sshropshire: Returning all other payment method eligibilities here: [f0b95d6](https://github.com/paypal/paypal-android/pull/268/commits/f0b95d69ba3f498f7303643466cccf13d3079b42)

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/EligibilityResult.kt
sshropshire: Addressed: [f0b95d6](https://github.com/paypal/paypal-android/pull/268/commits/f0b95d69ba3f498f7303643466cccf13d3079b42)

CorePayments/src/main/java/com/paypal/android/corepayments/apis/eligibility/EligibilityAPI.kt
sshropshire: Updated here: [be3a95b](https://github.com/paypal/paypal-android/pull/268/commits/be3a95beaa29eead1dcd84488b48966d4d94e9b6)

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/EligibilityClient.kt
sshropshire: Added unit tests here [3e27393](https://github.com/paypal/paypal-android/pull/268/commits/3e273930acd828d28aafeecb3a92fcda866467c5) and here [5ddb367](https://github.com/paypal/paypal-android/pull/268/commits/5ddb367e7624f4ef3378d58806438435e96f9b69).

tdchow (approved):

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/EligibilityRequest.kt
scannillo: Yeh the Intent is widespread. It lives with the orderID from creation all the way through to it's final approval. Here's how our Orders API public docs describe it https://github.com/braintree/braintree_ios/pull/1370

scannillo: IMO just Intent is enough

scannillo (approved):

CorePayments/src/main/java/com/paypal/android/corepayments/features/eligibility/EligibilityRequest.kt
jaxdesmarais: Are these the only intents (across any flow) that we will accept? If so agree we can go with generic `Intent` - I had assumed other requests would need different intents which was why it initially leaned `EligibilityIntent`. If that's not a concern lets do it!

sshropshire: Yeah I also like `Intent` but there's already an Android `Intent` class. It may be better for merchants if we go with `OrderIntent` on Android to avoid a class name collision.

CorePayments/src/main/java/com/paypal/android/corepayments/apis/eligibility/EligibilityAPI.kt
```diff
@@ -0,0 +1,61 @@
...
+        const val VARIABLE_CLIENT_ID = "clientId"
+        const val VARIABLE_INTENT = "intent"
+        const val VARIABLE_CURRENCY = "currency"
+        const val VARIABLE_ENABLE_FUNDING = "enableFunding"
+    }
+
+    @Throws(PayPalSDKError::class)
+    suspend fun checkEligibility(context: Context, request: EligibilityRequest): Eligibility {
+        val query = resourceLoader.loadRawResource(context, R.raw.graphql_query_funding_eligibility)
+        val enableFundingMethods = listOf(SupportedPaymentMethod.VENMO.name)
```
KunJeongPark, line 29: What is this?

sshropshire: For the most part it's equivalent to [this line](https://github.com/paypal/paypal-ios/blob/4335a57dab36d89b21257f617ffb02265aa0c712/Sources/CorePayments/Eligibility/EligibilityRequest.swift#L26) in `EligibilityRequest.swift`.

KunJeongPark: Doesn't it just return "VENMO"?

KunJeongPark: I'm just curious, why do we need this when the endpoint seems to return all eligibility for all the payment methods?

sshropshire: Yeah that's a good question actually I'm not sure. The GraphQL query does seem like it should take a dynamic array of requested payment method types.

PR #275, opened 2024-10-10
Remove PayPalNativePayments

sshropshire:
### Summary of changes

 - This PR removes the `:PayPalNativePayments` module
 - The PayPal Mobile Checkout library has been sunset and will not receive updates going forward

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


Demo/src/main/java/com/paypal/android/api/services/SDKSampleServerAPI.kt
```diff
@@ -59,12 +57,6 @@ class SDKSampleServerAPI {
         @GET("/client_id")
         suspend fun fetchClientId(): ClientId
 
-        @PATCH("/orders/{orderId}")
```
KunJeongPark, line 62: Oh I need to remove this too. I forgot that patch was just for NativeCheckout.

Demo/src/main/java/com/paypal/android/api/services/SDKSampleServerAPI.kt
```diff
@@ -136,7 +128,7 @@ class SDKSampleServerAPI {
         if (DEFAULT_ORDER_ID != null) {
             Order(DEFAULT_ORDER_ID, "CREATED")
         } else {
-            val body = JsonParser.parseString(orderRequest.toString()) as JsonObject
```
KunJeongPark: What is your reason for changing this?

KunJeongPark (approved):

Demo/src/main/java/com/paypal/android/api/services/SDKSampleServerAPI.kt
sshropshire: Yeah it couldn't hurt. There was an actual compilation issue on the Android side though. We were referencing a type that was owned by the Native XO sdk.

Demo/src/main/java/com/paypal/android/api/services/SDKSampleServerAPI.kt
sshropshire: Thanks for mentioning this. I may investigate it a bit further. What I think happened is the removal of mxo caused a ripple effect that downgraded the version of GSON linked by the Demo app somehow.

I'm going to upgrade the explicit version of GSON we reference in `build.gradle` to see if we can keep `JsonParser.parseString()` like it was before.

sshropshire: Was able to clean it up! [d46b91f](https://github.com/paypal/paypal-android/pull/275/commits/d46b91f683f486a82af38c086db57a15967280eb)

KunJeongPark: Great work! I didn't know that it does so much automatically.

sshropshire: Yeah it has to do with dependency resolution in [Gradle](https://docs.gradle.org/current/userguide/dependency_resolution.html#sub:resolution-strategy). When there are two versions of the same library linked to a project, Gradle will always pick the newest one.

PR #276, opened 2024-10-15
V2 Beta Auth Challenge Migration

sshropshire:
## Summary of changes

See [migration guide](https://github.com/paypal/paypal-android/blob/v2_beta__auth_challenge/v2_MIGRATION_GUIDE.md) for an in-depth description of the changes made.

### CardPayments

- Remove `CardClient(FragmentActivity, Config)` constructor
- Add `CardClient(Context, Config)` constructor
- Remove `CardClient.approveOrder(FragmentActivity, CardRequest)` method
- Add `CardClient.approveOrder(CardRequest)` method
- Remove `CardClient.presentAuthChallenge(FragmentActivity, CardAuthChallenge)` method
- Add `CardClient.presentAuthChallenge(ComponentActivity, CardAuthChallenge)` method
- Add `CardPresentAuthChallengeResult` type
- Add `CardClient.completeAuthChallenge(Intent, String)` method
- Add `CardStatus` type

### PayPalWebPayments

- Remove `PayPalWebCheckoutClient(FragmentActivity, CoreConfig, String)` constructor
- Add `PayPalWebCheckoutClient(Context, CoreConfig, String)` constructor
- Remove `PayPalWebCheckoutClient.start(PayPalWebCheckoutRequest)` method
- Add `PayPalWebCheckoutClient.start(ComponentActivity, PayPalWebCheckoutRequest)` method
- Remove `PayPalWebCheckoutClient.vault(PayPalWebVaultRequest)` method
- Add `PayPalWebCheckoutClient.vault(ComponentActivity, PayPalWebVaultRequest)` method
- Add `CardClient.completeAuthChallenge(Intent, String)` method
- Add `PayPalPresentAuthChallengeResult` type
- Add `PayPalWebStatus` type

 ## Checklist

 - [x] Added a changelog entry

- @sshropshire


Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
```diff
@@ -90,6 +95,22 @@ class ApproveOrderViewModel @Inject constructor(
                         approveOrderState = ActionState.Success(result)
                     }
 
+                    override fun onAuthorizationRequired(authChallenge: CardAuthChallenge) {
```
KunJeongPark, line 98: This is new for merchant. It puts the responsibility of implementing this and presenting AuthChallenge on merchants which seems very different from our existing pattern in Android and in iOS. Am I understanding this correctly?

sshropshire: Yup this is how BT `v5` now does it. It gives the merchant more control. We also no longer need callback methods like `override fun threeDSecureWillLaunch()`.

We'll also be getting rid of the listener pattern entirely in a separate PR.

Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderView.kt
```diff
@@ -43,6 +44,13 @@ fun ApproveOrderView(
         // continuously scroll to bottom of the list when scroll bounds change
         scrollState.animateScrollTo(scrollState.maxValue)
     }
+
+    val context = LocalContext.current
+    OnLifecycleOwnerResumeEffect {
+        val intent = context.getActivityOrNull()?.intent
+        intent?.let { viewModel.completeAuthChallenge(intent) }
```
KunJeongPark, line 51: This is taking place of previous CardLifeCycleObserver in the SDK, right?

sshropshire: Yes `CardLifeCycleObserver` used to do something similar behind the scenes. The `v2` integration requires the merchant to explicitly complete the auth challenge.

KunJeongPark: Let me summarize what I understand so far:
- There was change from passing in `AppCompatActivity` to `ComponentActivity` into functions like `approveOrder`
- Lifecycle Observer was removed from SDK
- Merchant application is now responsible for managing lifecycle and presenting authChallenge.

These changes were made for edge cases where there may be memory leaks in different merchant app implementations and undefined behavior in cases of process kill

sshropshire: Yup! That's it in summary. There will be follow up PRs to remove listeners as well. Ideally the SDK will be completely stateless in `v2`.

KunJeongPark: Nice work Steven! This is a lot!

KunJeongPark (approved):

Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderView.kt
sshropshire: 🙏 hopefully will be less going forward! This PR is a big tech debt payment. I'm glad we're able to get it merged.

CardPayments/src/main/java/com/paypal/android/cardpayments/ApproveOrderListener.kt
```diff
@@ -15,6 +15,9 @@ interface ApproveOrderListener {
     @MainThread
     fun onApproveOrderSuccess(result: CardResult)
```
KunJeongPark, line 16: That's interesting. I don't think it's possible to do this in iOS with protocols.
Do you feel like this is something merchant would want us to handle?

Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderView.kt
mikz88: Merchant_ID:🌐[7RZLE5LYGRGYE]: 
paypal.me/kimsoledad88

PR #281, opened 2024-11-05
Browser Switch Bugfix

sshropshire:
### Summary of changes

 - This PR udpates `:CardPayments` and `:PayPalPayments` to use the latest version of Browser Switch

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -150,6 +151,10 @@ class CardClient internal constructor(
 
             is CardStatus.ApproveOrderSuccess -> notifyApproveOrderSuccess(status.result)
             is CardStatus.ApproveOrderCanceled -> notifyApproveOrderCanceled(status.orderId)
+            is CardStatus.UnknownError -> {
+                Log.d("PayPalSDK", "An unknown error occurred: ${status.error.message}")
```
sshropshire, line 155: Currently, it's unclear which listener should receive this error message.

This should only occur when the `authState` variable sent to the client is tampered with. Once listeners are removed from the API, this will no longer be an issue.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -98,6 +99,9 @@ class PayPalWebCheckoutClient internal constructor(
             is PayPalWebStatus.VaultSuccess -> notifyVaultSuccess(status.result)
             is PayPalWebStatus.VaultError -> notifyVaultFailure(status.error)
             PayPalWebStatus.VaultCanceled -> notifyVaultCancelation()
+            is PayPalWebStatus.UnknownError -> {
+                Log.d("PayPalSDK", "An unknown error occurred: ${status.error.message}")
```
sshropshire, line 103: Same issue as seen [here in CardClient](https://github.com/paypal/paypal-android/pull/281/files#r1830010584).

KunJeongPark:
Steven, is this the fix for the GitHub inbound #282?

sshropshire:
It might fix #282 but technically it's unrelated. This PR migrates us to the [new browser switch](https://github.com/braintree/browser-switch-android/blob/main/v3_MIGRATION_GUIDE.md).

KunJeongPark (approved):

PR #294, opened 2024-11-19
Remove 3DS URL Parsing

sshropshire:
### Summary of changes

 - Remove unnecessary URL parsing from all Card 3DS routes to obtain feature parity with iOS
 - Clean up CHANGELOG after merging `v2_beta`

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

PR #296, opened 2024-11-21
[1.x] Remove 3DS URL Parsing

sshropshire:
### Summary of changes

 - Remove unnecessary URL parsing from all Card 3DS routes to obtain feature parity with iOS
 - We are back-porting this change to `v1`
 - We made the same change in v2: https://github.com/paypal/paypal-android/pull/294

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -43,17 +43,17 @@ jobs:
           java-version: '17'
           distribution: 'microsoft'
 
-      # Assemble artifacts from main branch
-      - name: Checkout Main Branch
+      # Assemble artifacts from Release branch
+      - name: Checkout Release Branch
         uses: actions/checkout@v4
         with:
-          ref: main
-      - name: Assemble Release AAR on Main Branch
+          ref: 1.x
+      - name: Assemble Release AAR on Release Branch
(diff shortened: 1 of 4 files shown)


.github/workflows/static_analysis.yml
```diff
@@ -43,17 +43,17 @@ jobs:
...
 
-      # Assemble artifacts from main branch
-      - name: Checkout Main Branch
+      # Assemble artifacts from Release branch
+      - name: Checkout Release Branch
         uses: actions/checkout@v4
         with:
-          ref: main
-      - name: Assemble Release AAR on Main Branch
+          ref: 1.x
```
sshropshire, line 50: This GitHub Action compares SDK size with previous versions. I had to update these to create a separate set of artifacts for `1.x` releases because `:PayPalNativePayments` is no longer being built on `main` (this causes the action to fail).

KunJeongPark: Steven, do you need to change it back when you do release for 2.0.0 beta versions?

KunJeongPark (approved):

.github/workflows/static_analysis.yml
sshropshire: Just to clarify this is only for `1.x` releases and should not affect `main` in any way.

PR #300, opened 2024-11-26
Migrate PayPalWebClient Analytics Logic Into PayPalWebAnalytics Class

sshropshire:
### Summary of changes

 - This PR creates a `PayPalWebAnalytics` class to encapsulate analytics notification logic
 - Externalizing Analytics code should make Analytics Events easier to audit in one place
 - Externalizing Analytics will also make future refactoring to migrate away from the listener pattern more clear

 ### Checklist

- @sshropshire

### DIFF
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardAnalytics.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardAnalytics.kt
@@ -3,9 +3,7 @@ package com.paypal.android.cardpayments
 import com.paypal.android.corepayments.analytics.AnalyticsService
 
 @Suppress("TooManyFunctions")
-class CardAnalytics(
-    private val analyticsService: AnalyticsService
-) {
+class CardAnalytics(private val analyticsService: AnalyticsService) {
 
     // region Approve Order
     fun notifyApproveOrderStarted(orderId: String) {
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardAnalytics.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardAnalytics.kt
@@ -3,7 +3,7 @@ package com.paypal.android.cardpayments
 import com.paypal.android.corepayments.analytics.AnalyticsService
 
 @Suppress("TooManyFunctions")
-class CardAnalytics(private val analyticsService: AnalyticsService) {
+internal class CardAnalytics(private val analyticsService: AnalyticsService) {
 
     // region Approve Order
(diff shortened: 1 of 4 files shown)


CardPayments/src/main/java/com/paypal/android/cardpayments/CardAnalytics.kt
```diff
@@ -3,9 +3,7 @@ package com.paypal.android.cardpayments
 import com.paypal.android.corepayments.analytics.AnalyticsService
 
 @Suppress("TooManyFunctions")
-class CardAnalytics(
-    private val analyticsService: AnalyticsService
-) {
+internal class CardAnalytics(private val analyticsService: AnalyticsService) {
```
sshropshire, line 6: Marking analytics related classes `internal`. I forgot to add this in the previous PR so adding it here now.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -5,7 +5,6 @@ import android.content.Intent
 import android.util.Log
 import androidx.activity.ComponentActivity
 import com.paypal.android.corepayments.CoreConfig
-import com.paypal.android.corepayments.PayPalSDKError
```
KunJeongPark, line 8: How come you don't need this now?

sshropshire: Looks like this import was only used in a function declaration e.g.

```kotlin
private fun notifyWebCheckoutFailure(error: PayPalSDKError, orderId: String?) {}
```

We removed those methods and the import got flagged by `detekt` as no longer needed.

KunJeongPark: Is it not needed because the launcher imports `PayPalWebCheckoutError`? I think I got it.

KunJeongPark (approved):

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
sshropshire: Yup since `PayPalWebCheckoutError` is no longer explicitly referenced in this file, we're able to skip importing it.

PR #302, opened 2024-12-02
Card: Remove try/catch from Approve Order Flow

sshropshire:
### Summary of changes

 - This PR removes `try/catch` from `CardClient.approveOrder()`
 - To do this, we've renamed `ConfirmPaymentSourceResponse` to `ConfirmPaymentSourceResult` and made it a sealed class with `.Success` and `.Failure` subclasses
 - Also in Kotlin `try/catch` is unchecked, meaning the compiler will not notify us when we call a method that `throws` outside of a try/catch block.
 - Using "Result" types cleans up the codebase by providing compile time support for error handling 

 ### Checklist

- @sshropshire

### DIFF
--- a/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
+++ b/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
@@ -41,12 +41,6 @@ class CardClient internal constructor(
     private var approveOrderId: String? = null
 
     // TODO: remove once try-catch is removed
-    private val approveOrderExceptionHandler = CoreCoroutineExceptionHandler { error ->
-        analytics.notifyApproveOrderUnknownError(null)
-        approveOrderListener?.onApproveOrderFailure(error)
-    }
-
-    // TODO: remove once try-catch is removed
     private val vaultExceptionHandler = CoreCoroutineExceptionHandler { error ->
         analytics.notifyVaultUnknownError(null)
         cardVaultListener?.onVaultFailure(error)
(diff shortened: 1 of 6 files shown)


CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -76,26 +71,28 @@ class CardClient internal constructor(
...
+        CoroutineScope(dispatcher).launch {
+            when (val response = checkoutOrdersAPI.confirmPaymentSource(cardRequest)) {
+                is ConfirmPaymentSourceResult.Success -> {
+                    if (response.payerActionHref == null) {
+                        analytics.notifyApproveOrderSucceeded(response.orderId)
+                        val result =
+                            response.run { CardResult(orderId = orderId, status = status?.name) }
+                        approveOrderListener?.onApproveOrderSuccess(result)
+                    } else {
+                        analytics.notifyApproveOrderAuthChallengeReceived(cardRequest.orderId)
```
KunJeongPark, line 83: Do you check formatting of the returnURL? In iOS, we have this code that checks formatting https://github.com/paypal/paypal-ios/blob/60f2f95eab44048211b5289e8fb5b87538daa751/Sources/CardPayments/CardClient.swift#L108

KunJeongPark:
It's cleaner but just double checking my understanding of the Kotlin code, you were catching all the exceptions in the try catch block yourself in the functions, right?

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
sshropshire: We don't actually on Android. Is this something we can omit?

sshropshire:
> It's cleaner but just double checking my understanding of the Kotlin code, you were catching all the exceptions in the try catch block yourself in the functions, right?

Yes the SDK has been catching exceptions internally. But in our codebase, most of the time we are the ones throwing the exception via the `throw` keyword.

In contrast to throwing exceptions, returning a custom `Result` type is seen as a best practice in Kotlin because it enables the compiler to ensure that all possible error paths are handled when coupled with a `when` statement (similar to `switch` in Swift).

[This article](https://elizarov.medium.com/kotlin-and-exceptions-8062f589d07) breaks it down nicely. Ultimately the author [recommends](https://elizarov.medium.com/kotlin-and-exceptions-8062f589d07#2273) using a `sealed class` in Kotlin to signal error. This will also be the pattern we present to merchants once we migrate from listeners to callbacks.

KunJeongPark (approved):

PR #303, opened 2024-12-03
Card: Remove try/catch from Vault Flow

sshropshire:
### Summary of changes

 - This PR removes `try/catch` from `CardClient.vault()`
 - To do this, we've refactored `UpdateSetupTokenResult` to make it a sealed class with `.Success` and `.Failure` subclasses
 - Using "Result" types cleans up the codebase by providing compile time support for error handling 
 - This is a follow up PR to #302 

 ### Checklist

- @sshropshire


KunJeongPark:
Are you going to do this for all the API calls? It's an interesting pattern. I'm thinking about this for iOS. 
Not sure if it works with async/await pattern in iOS. It'd be weird, to have do catch block and non throwing function that returns Result type. It takes away advantage of error propagation in iOS async/await.

But I see that's similar to what you are doing here. You have try/catch block that returns a result type. 

I think you had mentioned that you were going to remove listener pattern eventually.
So the merchant would receive the result type as well?

sshropshire:
@KunJeongPark yes for Kotlin it's considered a best practice. Honestly I'm not sure this will be needed for iOS though. Swift does have checked exceptions i.e. when you call code that `throws` you have to surround it with `do/catch` in order for compilation to succeed. On Kotlin the compiler doesn't even warn us when we call throwing code without exception handling, and that's the primary reason we need a pattern like this on Android to make sure we have compiler enforcement.

CorePayments/src/main/java/com/paypal/android/corepayments/ResourceLoader.kt
```diff
@@ -21,19 +21,22 @@ class ResourceLoader {
...
+                LoadRawResourceResult.Success(String(resAsBytes))
+            } catch (e: Resources.NotFoundException) {
+                val errorDescription = "Resource with id $resId not found."
+                val resourceNotFoundError = PayPalSDKError(0, errorDescription, reason = e)
+                LoadRawResourceResult.Failure(resourceNotFoundError)
+            } catch (e: IOException) {
+                val errorDescription = "Error loading resource with id $resId."
+                val ioError = PayPalSDKError(0, errorDescription, reason = e)
+                LoadRawResourceResult.Failure(ioError)
+            }
```
KunJeongPark, line 40: Could it throw other exceptions?

sshropshire: That's a good question. I took a thorough tour of the docs and here are all the methods called:

- [Context#getResources()](https://developer.android.com/reference/android/content/Context#getResources()) does not throw
- [Resources#openRawResource()](https://developer.android.com/reference/android/content/res/Resources#openRawResource(int,%20android.util.TypedValue)) `throws Resources.NotFoundException`
- [InputStream#available()](https://developer.android.com/reference/java/io/InputStream#available()) `throws IOException`
- [ByteArray()](https://docs.oracle.com/cd/E14571_01/apirefs.1111/e12063/oracle/dss/util/xdo/common/io/ByteArray.html#ByteArray(byte[])) does not throw
- [InputStream#read()](https://developer.android.com/reference/java/io/InputStream#read(byte[])) `throws IOException, NullPointerException`
- [InputStream#close()](https://developer.android.com/reference/java/io/InputStream#close()) `throws IOException`

According to the docs `InputStream#read()` will only throw if `resAsBytes` is `null`, and through Kotlin we have guarantees that it is non-null. All the other possible exceptions we've caught.

KunJeongPark (approved):

PR #304, opened 2024-12-03
Analytics: Fix SDK Version Bug

sshropshire:
### Summary of changes

 - This PR fixes a bug in our Gradle file that tries to inject an environment variable that doesn't exist
 - We should use `rootProject.version` going forward instead of `versionName`

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # PayPal Android SDK Release Notes
 
+## unreleased
+* CorePayments
+  * Fix issue that causes analytics version number to always be `null`
+
 ## 2.0.0-beta1 (2024-11-20)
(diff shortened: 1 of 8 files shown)


KunJeongPark:
Did you check Notebooks to see analytics output?

sshropshire:
> Did you check Notebooks to see analytics output?

@KunJeongPark I can, I haven't verified end-to-end yet.

KunJeongPark (approved):

sshropshire:
@KunJeongPark I was able to verify and also ended up adding another revision to send all analytics to the LIVE environment so they can be tracked. Great catch!

PR #305, opened 2024-12-05
Card: Remove Approve Order Listener Pattern

sshropshire:
### Summary of changes

 - This PR migrates the `CardClient` approve order flow away from the listener pattern.
 - Migration Guide updates will come in a later PR
 - Users must now balance calls to `CardClient.presentAuthChallenge()` with calls to `CardClient.finishApproveOrder()` to complete an order flow

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


CardPayments/src/main/java/com/paypal/android/cardpayments/CardApproveOrderCallback.kt
```diff
@@ -0,0 +1,13 @@
...
+import androidx.annotation.MainThread
+
+fun interface CardApproveOrderCallback {
+    /**
+     * Called when the order is approved.
+     *
+     * @param result [CardApproveOrderResult] result with details
+     */
+    @MainThread
+    fun onCardApproveOrderResult(result: CardApproveOrderResult)
```
KunJeongPark, line 12: That's interesting. I don't think it's possible to do this in iOS with protocols.
Do you feel like this is something merchant would want us to handle?

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -69,22 +68,27 @@ class CardClient internal constructor(
                 is ConfirmPaymentSourceResult.Success -> {
                     if (response.payerActionHref == null) {
```
KunJeongPark, line 69: this syntax looks really similar to completion handler. So callback in place of listener pattern is safer for memory management?

CardPayments/src/main/java/com/paypal/android/cardpayments/CardApproveOrderCallback.kt
sshropshire: I'm not sure if we need to deliver results on the main thread, but it is convenient when updating the UI after executing a transaction.

It's up to y'all if you want to do the same. I feel like the closest equivalent to this annotation in Swift would be `@MainActor`.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
sshropshire: It's similar. The main difference now is there's a callback for every async method. With the listener pattern, there is only one listener, and it needs to be set by the merchant before starting any of the payment flows.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt
```diff
@@ -69,22 +68,27 @@ class CardClient internal constructor(
...
+                        }
+                        callback.onCardApproveOrderResult(result)
                     } else {
                         analytics.notifyApproveOrderAuthChallengeReceived(cardRequest.orderId)
-                        approveOrderListener?.onApproveOrderThreeDSecureWillLaunch()
 
                         val url = Uri.parse(response.payerActionHref)
                         val authChallenge = CardAuthChallenge.ApproveOrder(url, cardRequest)
-                        approveOrderListener?.onApproveOrderAuthorizationRequired(authChallenge)
+                        val result = CardApproveOrderResult.AuthorizationRequired(authChallenge)
```
KunJeongPark, line 83: It's interesting that you have this intermediate state on the same result type.
So there is result type for API call and final result type for the merchant.

sshropshire: This method does have three possible results. Alternatively, I thought about bundling the `payer_action_required` status into the `Failure` case but this felt more natural.

Technically it isn't a failure. PPaaS just needs more information before the buyer can continue.

Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
```diff
@@ -89,47 +87,43 @@ class ApproveOrderViewModel @Inject constructor(
...
-                        Log.d(TAG, "3DS Success")
-                    }
+    private fun presentAuthChallenge(
+        activity: ComponentActivity,
+        authChallenge: CardAuthChallenge
+    ) {
+        cardClient?.presentAuthChallenge(activity, authChallenge)?.let { presentAuthResult ->
+            when (presentAuthResult) {
+                is CardPresentAuthChallengeResult.Success -> {
+                    authState = presentAuthResult.authState
```
KunJeongPark, line 121: I'm a bit lost on how this authState gets translated to success CardApproveOrderResult back to the merchant.

sshropshire: So `authState` is a JSON string that allows us to continue the flow once the merchant app comes back into the foreground after browser switching. We created this pattern in BT `v5` to allow Android merchants to recover from a process kill. In BT it's called a [pendingRequest](https://github.com/braintree/braintree_android/blob/main/v5_MIGRATION_GUIDE.md#paypal).

We can always come up with a new name, `authState` for me makes sense because it's similar to the `state` parameter used in [OAuth2](https://stackoverflow.com/questions/26132066/what-is-the-purpose-of-the-state-parameter-in-oauth-authorization-request).

sshropshire: Also this example should be a lot clearer in our revised sample merchant app we have planned.

KunJeongPark: I see that viewModel's completeAuthChallenge gets called in ApproveOrderView.

KunJeongPark (approved):

PR #307, opened 2024-12-11
PayPal: Remove Checkout Listener Pattern

sshropshire:
### Summary of changes

 - This PR migrates the `PayPalClient` checkout flow away from the listener pattern.
 - Migration Guide updates will come in a later PR
 - Users must now balance calls to `PayPalWebCheckoutClient.start()` with calls to `PayPalWebCheckoutClient.finishStart()` to complete a vault flow

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
   * Fix issue that causes analytics version number to always be `null`
 * Breaking Changes
   * CardPayments
+    * Remove `CardClient.approveOrderListener` property
+    * Remove `CardClient.cardVaultListener` property
```
sshropshire, line 9: I realized these should be added to `CardClient` as well.

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalWebViewModel.kt
```diff
@@ -173,6 +154,25 @@ class PayPalWebViewModel @Inject constructor(
     }
 
     fun handleBrowserSwitchResult(activity: ComponentActivity) {
-        authState?.let { paypalClient?.completeAuthChallenge(activity.intent, it) }
+        val result = authState?.let { paypalClient?.finishStart(activity.intent, it) }
```
KunJeongPark, line 157: I had to do a double take on this because our checkout main function is "start" 
But it goes with the pattern. finish[function]

sshropshire: Yeah honestly I wouldn't mind renaming it to either `checkout()` or `confirmPaymentSource()`. It might be nice to have `CardClient.confirmPaymentSource()` as well instead of `approveOrder()` to align with PPaaS.

sshropshire: Then it would be `finishCheckout()` or `finishConfirmPaymentSource()`.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebLauncher.kt
```diff
@@ -106,6 +107,25 @@ internal class PayPalWebLauncher(
...
+            is BrowserSwitchFinalResult.Failure -> {
+                // TODO: remove error codes and error description from project; the built in
+                // Throwable type already has a message property and error codes are only required
+                // for iOS Error protocol conformance
+                val message = "Browser switch failed"
+                val browserSwitchError = PayPalSDKError(0, message, reason = finalResult.error)
+                PayPalWebCheckoutFinishStartResult.Failure(browserSwitchError, null)
+            }
+
+            BrowserSwitchFinalResult.NoResult -> PayPalWebCheckoutFinishStartResult.NoResult
```
KunJeongPark, line 125: How does it return no result?

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebLauncher.kt
```diff
@@ -116,25 +136,31 @@ internal class PayPalWebLauncher(
...
+                val orderId = metadata.optString(METADATA_KEY_ORDER_ID)
+                if (orderId.isNullOrBlank() || payerId.isNullOrBlank()) {
+                    val malformedResultError = PayPalWebCheckoutError.malformedResultError
+                    PayPalWebCheckoutFinishStartResult.Failure(malformedResultError, orderId)
+                } else {
+                    PayPalWebCheckoutFinishStartResult.Success(orderId, payerId)
+                }
             }
+        } else {
+            PayPalWebCheckoutFinishStartResult.NoResult
```
KunJeongPark, line 163: I c. this is possible although just theoretically, right?

sshropshire: `NoResult` really means the SDK doesn't have enough information to determine the state of the browser switched flow. We don't receive a "close" event with Chrome Custom Tabs or anything, so when the merchant app comes back into the foreground we have to parse the deep link to determine if it was successful.

Some merchants may choose to support other browser-switch enabled payment methods in the same view. E.g. we have to make sure that we don't provide a false-postive result for PayPal success when the deep link indicates that the success event is actually for Card. So if the deep link is for another payment method other than PayPal, `PayPalWebCheckoutClient` returns `NoResult`.

KunJeongPark (approved):

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalWebViewModel.kt
KunJeongPark: I like `checkout(`) but maybe it's misleading because we provide intermediate steps in our SDK.
In `CardClient`, we have `approveOrder`, which is what we do. maybe that makes sense
`approveOrder` for both `CardClient` and `PayPalwebCheckout` but let's hold on changing that
so we can address it on different PRs. wdyt?

sshropshire: My bad I missed this. I agree we should hold off on re-naming until we're closer to GA.

PR #311, opened 2025-01-28
Remove onNewIntent Override in MainActivity

sshropshire:
### Summary of changes

 - We no longer need to override `onNewIntent` in the Demo app's `MainActivity`
 - This PR refactors each Composable to use `OnNewIntentEffect` internally

 ### Checklist

- @sshropshire

### DIFF
--- a/Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderView.kt
+++ b/Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderView.kt
@@ -27,6 +27,7 @@ import com.paypal.android.uishared.components.OrderView
 import com.paypal.android.uishared.components.StepHeader
 import com.paypal.android.uishared.state.CompletedActionState
 import com.paypal.android.utils.OnLifecycleOwnerResumeEffect
+import com.paypal.android.utils.OnNewIntentEffect
 import com.paypal.android.utils.UIConstants
 import com.paypal.android.utils.getActivityOrNull
 
@@ -48,7 +49,11 @@ fun ApproveOrderView(
     val context = LocalContext.current
     OnLifecycleOwnerResumeEffect {
         val intent = context.getActivityOrNull()?.intent
-        intent?.let { viewModel.completeAuthChallenge(intent) }
+        intent?.let { viewModel.completeAuthChallenge(it) }
+    }
+
+    OnNewIntentEffect { newIntent ->
+        viewModel.completeAuthChallenge(newIntent)
     }
 
     val contentPadding = UIConstants.paddingMedium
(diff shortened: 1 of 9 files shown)


Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderView.kt
```diff
@@ -48,7 +49,11 @@ fun ApproveOrderView(
     val context = LocalContext.current
     OnLifecycleOwnerResumeEffect {
         val intent = context.getActivityOrNull()?.intent
-        intent?.let { viewModel.completeAuthChallenge(intent) }
+        intent?.let { viewModel.completeAuthChallenge(it) }
```
KunJeongPark, line 52: Why do we need to check onResume as well as newIntent if we are assuming singeTop launch mode?
Is there a chance that this is called the second time in onResume?

sshropshire: This is a good callout. This is need for process kill, but I do need to make sure we're doing Activity state restoration and I'll debug with "Don't keep activities" to make sure it's being called.

Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
```diff
@@ -235,18 +235,25 @@ class ApproveOrderViewModel @Inject constructor(
                         OrderInfo(orderId, status, didAttemptThreeDSecureAuthentication)
                     }
                     approveOrderState = ActionState.Success(orderInfo)
+                    // discard authState
+                    authState = null
```
sshropshire, line 239: This is mentioned in the [v2_MIGRATION_GUIDE](https://github.com/paypal/paypal-android/blob/main/v2_MIGRATION_GUIDE.md#card-payments) but it was missing here.

The merchant needs to discard `authState` when they are done to prevent stale auth state from being parsed again to check for auth completion when a merchant app comes back into the foreground. Notice `authState` is not discarded when `.NoResult` is returned by the SDK.

KunJeongPark (approved):

Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
KunJeongPark: can u explain that a bit Steven? why authState is not discarded with .NoResult?

sshropshire: Yeah in the demo app, keeping `authState` when we can't determine if the user completed the flow allows the user to retry launching the browser-switched flow.

aaronh197 (approved):

PR #312, opened 2025-03-06
Fix NoResult Error

sshropshire:
### Summary of changes

 - Fix an issue with our PayPalWebCheckout demo that causes a `NoResult` event to be fired immediately after a `Success` event
 - Initially, a `Success` event is properly parsed as a result form `OnNewIntentEffect`
 - There is a bug in `PayPalWebViewModel` that causes the UI to reset the PayPal web launch button
 - To resolve this bug, we should expect `PayPalWebViewModel.completeAuthChallenge()` to no-op when no `authState` is present

 ### Checklist

- @sshropshire

### DIFF
--- a/Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalWebViewModel.kt
+++ b/Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalWebViewModel.kt
@@ -150,32 +150,35 @@ class PayPalWebViewModel @Inject constructor(
     }
 
     fun completeAuthChallenge(intent: Intent) {
-        val result = authState?.let { paypalClient?.finishStart(intent, it) }
-        when (result) {
-            is PayPalWebCheckoutFinishStartResult.Success -> {
-                payPalWebCheckoutState = ActionState.Success(result)
-                // discard authState
-                authState = null
-            }
+        authState?.let { authState ->
+            val result = paypalClient?.finishStart(intent, authState)
+            when (result) {
+                is PayPalWebCheckoutFinishStartResult.Success -> {
+                    payPalWebCheckoutState = ActionState.Success(result)
+                    discardAuthState()
+                }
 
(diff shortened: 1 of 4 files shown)


Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalWebViewModel.kt
```diff
@@ -149,33 +149,38 @@ class PayPalWebViewModel @Inject constructor(
         }
     }
 
+    private fun checkIfPayPalAuthFinished(intent: Intent): PayPalWebCheckoutFinishStartResult? =
+        authState?.let { paypalClient?.finishStart(intent, it) }
+
     fun completeAuthChallenge(intent: Intent) {
-        val result = authState?.let { paypalClient?.finishStart(intent, it) }
```
KunJeongPark, line 153: is this what triggered another completeAuthChallenge after Success result was parsed?

KunJeongPark (approved):

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalWebViewModel.kt
sshropshire: For the most part yes. The bug was really subtle, it boiled down to eliminating the `null` case in this switch statement in [PayPalWebViewModel.kt](https://github.com/paypal/paypal-android/blob/0ce520e2fdb70051cd8de42fe3e04441671f550d/Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalWebViewModel.kt#L175):

Screenshot 2025-03-13 at 1 08 14 PM

Both `onNewIntent` and `onResume` can trigger a call to `completeAuthChallenge()`. The expected behavior is there should be no affect when auth state is `null`, but the actual behavior is the UI gets reset even though it completed successfully. It also happens really fast–without breakpoints it just looks like it fails.

PR #314, opened 2025-04-02
Card Form in PayPalWebPayments

sshropshire:
### Summary of changes

 - This PR adds BCDC to the `:PayPalWebPayments` module
 - Corresponding iOS PR: https://github.com/paypal/paypal-ios/pull/321

 ### Checklist

 - [ ] Added a changelog entry

- @sshropshire


PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -61,6 +73,53 @@ class PayPalWebCheckoutClient internal constructor(
...
+                val updateConfigResult = request.run {
+                    updateClientConfigAPI.updateClientConfig(
+                        orderId = orderId,
+                        fundingSource = fundingSource
+                    )
+                }
+                if (updateConfigResult is UpdateClientConfigResult.Failure) {
+                    // notify failure
+                    callback.onPayPalWebStartResult(
+                        PayPalPresentAuthChallengeResult.Failure(updateConfigResult.error)
```
KunJeongPark, line 100: Have we decided to return failure upon failure on this GraphQL call?
I know JS SDK does not do that atm and Ravi had mentioned that this is probably what should be done.

sshropshire: I don't think we've made a decision on this. I'd be curious to know how failure affects the overall transaction before making the final call.

linetterobinson01-design (approved):

PR #325, opened 2025-05-23
Rename PaymentButton `shape` Property to `edges`

sshropshire:
### Summary of changes

 - This PR renames PaymentButton's `shape` to `edges` 

 ### Checklist

 - [x] Added a changelog entry

### <ins>Screenshot</ins>

- @sshropshire


PaymentButtons/src/main/res/values/dimens.xml
```diff
@@ -24,6 +26,6 @@
     <dimen name="paypal_payment_button_label_text_size_large">18sp</dimen>
 
     <!-- Make logo height 76% of button -->
-    <item name="paypal_payment_button_logo_weight" type="dimen" format="float">0.76</item>
-    <item name="paypal_payment_button_vertical_spacer_weight" type="dimen" format="float">0.12</item>
```
sshropshire, line 28: I thought the additional `type="dimen"` and `float="float"` attributes were necessary at one point, but I was able to verify using `<dimen />` by itself works fine.

Demo/src/main/java/com/paypal/android/ui/paypalbuttons/PayPalButtonsUiState.kt
```diff
@@ -3,13 +3,12 @@ package com.paypal.android.ui.paypalbuttons
...
+import com.paypal.android.paymentbuttons.PaymentButtonEdges
 
 data class PayPalButtonsUiState(
     val fundingType: ButtonFundingType = ButtonFundingType.PAYPAL,
     val payPalCreditButtonColor: PayPalCreditButtonColor = PayPalCreditButtonColor.BLUE,
     val payPalButtonColor: PayPalButtonColor = PayPalButtonColor.BLUE,
     val payPalButtonLabel: PayPalButtonLabel = PayPalButtonLabel.PAYPAL,
-    val paymentButtonShape: PaymentButtonShape = PaymentButtonShape.ROUNDED,
-    val customCornerRadius: Int? = null
+    val paymentButtonEdges: PaymentButtonEdges = PaymentButtonEdges.Soft,
```
kgangineni, line 13: `paymentButtonEdges` to `paymentButtonEdge` since it only holds 1

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButtonEdges.kt
```diff
@@ -0,0 +1,30 @@
+package com.paypal.android.paymentbuttons
+
+sealed class PaymentButtonEdges {
```
kgangineni, line 3: `PaymentButtonEdges` to `PaymentButtonEdge` same in other places also

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
```diff
@@ -323,6 +259,27 @@ abstract class PaymentButton<C : PaymentButtonColor> @JvmOverloads constructor(
         prefixTextView.setTextColor(textColor)
         suffixTextView.setTextColor(textColor)
     }
+
+    private fun applyEdgeStyling() {
+        val cornerSize: Float = when (val edges = edges) {
```
kgangineni, line 264: don't need scoped val here `when(edge){
}`

Demo/src/main/java/com/paypal/android/ui/paypalbuttons/PaymentButtonEdgesOptionList.kt
```diff
@@ -0,0 +1,104 @@
...
+            color = MaterialTheme.colorScheme.onSurfaceVariant,
+            style = MaterialTheme.typography.titleLarge,
+            modifier = Modifier
+                .padding(bottom = UIConstants.paddingSmall)
+                .fillMaxWidth()
+        )
+        IntSlider(
+            value = cornerRadius.toInt(),
+            valueRange = 0..CORNER_RADIUS_SLIDER_MAX,
+            steps = CORNER_RADIUS_SLIDER_MAX,
```
kgangineni: max in valueRange and steps same? means we can either select 0 or max?

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
sshropshire: Yeah I initially had `when (edges)` but oddly it doesn't compile without the `val edges = edges` assignment.

Demo/src/main/java/com/paypal/android/ui/paypalbuttons/PaymentButtonEdgesOptionList.kt
sshropshire: What this code does is set the number of steps equal to the max value. In this case the max is 100, so there should be 100 valid step values.

I can refactor this so the intent is more clear. Something like `valueRange.length` would be better.

Demo/src/main/java/com/paypal/android/ui/paypalbuttons/PayPalButtonsUiState.kt
sshropshire: The ticket calls for `edges` as the property name. We may need keep this for feature parity with iOS for now.

Demo/src/main/java/com/paypal/android/ui/paypalbuttons/PaymentButtonEdgesOptionList.kt
sshropshire: Updated here: [b13a2db](https://github.com/paypal/paypal-android/pull/325/commits/b13a2db217fac58d7de20f54f95ef84d4a695fab)

kgangineni (approved):
👍

PR #327, opened 2025-06-06
Basic working POC of app switch

kgangineni:
### Summary of changes

 - modified web checkout to include app switch

 ### Checklist

 - [ ] Added a changelog entry

-


PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebLauncher.kt
```diff
@@ -29,13 +36,27 @@ internal class PayPalWebLauncher(
...
     }
 
-    fun launchPayPalWebCheckout(
+    suspend fun launchPayPalWebCheckout(
         activity: ComponentActivity,
-        request: PayPalWebCheckoutRequest,
+        request: PayPalCheckoutRequest,
     ): PayPalPresentAuthChallengeResult {
+
+        val token = fetchClientToken()
```
KunJeongPark, line 44: Are we not distinguishing the app switch vs non app switch flows here?
The card funding option needs to go through the old checkout flow to present the guest checkout for example.
I think they are separate flows for CCO patch and we would only need the new LSAT for the app switch flow.

kgangineni: how to differentiate between both flows, without having order details?

KunJeongPark: request.appSwitchEnabled?
I think the extra call to get the LSAT is not needed if app switch option is not selected.
For web SDK, this is done on server side, but for us, I think it could be more latency.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebLauncher.kt
```diff
@@ -29,13 +36,27 @@ internal class PayPalWebLauncher(
...
+        println("PayPalWebCheckoutClient: Client token fetched: $token")
+        val patchCcoResponse = patchCCOWithAppSwitchEligibility(
+            context = activity,
+            token = token,
+            orderId = request.orderId,
+            tokenType = "ORDER_ID",
+            merchantOptInForAppSwitch = request.appSwitchEnabled
```
KunJeongPark, line 51: Looks like the request has app switch enabled boolean

PR #328, opened 2025-06-25
Get low scoped access token

kgangineni:
### Summary of changes

 - Get access token using client id
 - Use it in `PaypalWebCheckoutClient`
 - Token is not used right now, added todo to be used in getting app switch eligibility
 ### Checklist

 - [ ] Added a changelog entry

-


CorePayments/src/main/java/com/paypal/android/corepayments/api/FetchClientToken.kt
```diff
@@ -0,0 +1,44 @@
...
+        val apiRequest = APIRequest(
+            path = "v1/oauth2/token",
+            method = HttpMethod.POST,
+            body = requestBody,
+            headers = headers
+        )
+        val httpResponse = restClient.send(apiRequest)
+        if (httpResponse.isSuccessful) {
+            val jsonObject = JSONObject(httpResponse.body ?: "")
+            return jsonObject.getString("access_token")
```
sshropshire: Could we use `jsonObject.optString()` here instead? I believe the regular `getString()` method will throw when `access_token` doesn't exist.

CorePayments/src/main/java/com/paypal/android/corepayments/api/FetchClientToken.kt
```diff
@@ -0,0 +1,44 @@
...
+        } else {
+            throw httpResponse.run { buildError(status, body) }
```
sshropshire: Could we return a `FetchClientResult.Failure()` type here instead of throwing?

CorePayments/src/main/java/com/paypal/android/corepayments/api/FetchClientToken.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.paypal.android.corepayments.APIClientError
+import com.paypal.android.corepayments.APIRequest
+import com.paypal.android.corepayments.CoreConfig
+import com.paypal.android.corepayments.HttpMethod
+import com.paypal.android.corepayments.RestClient
+import com.paypal.android.corepayments.base64encoded
+import org.json.JSONObject
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class FetchClientToken(
```
sshropshire:
```suggestion
class AuthenticationSecureTokenServiceAPI(
```
This name would fit our naming convention if we want to be 100% consistent with the other `API` suffixed classes.

CorePayments/src/main/java/com/paypal/android/corepayments/api/FetchClientToken.kt
```diff
@@ -0,0 +1,44 @@
...
+    private val coreConfig: CoreConfig,
+    private val restClient: RestClient = RestClient(configuration = coreConfig),
+) {
+    suspend operator fun invoke(): String {
```
sshropshire:
```suggestion
    suspend fun createToken(): String {
```
We mostly use the functor pattern for UseCase type classes. I am interested in this pattern. If we decide to go this route, I'd like for all of the other `API` suffixed classes to migrate to this pattern so we can keep it consistent.

CorePayments/src/test/java/com/paypal/android/corepayments/api/FetchClientTokenUnitTest.kt
```diff
@@ -0,0 +1,214 @@
...
+        // Verify headers
+        val headers = capturedRequest.headers!!
+        assertEquals("application/x-www-form-urlencoded", headers["Content-Type"])
+        assert(headers["Authorization"]!!.startsWith("Basic "))
+
+        assertEquals("test-token", result)
+    }
+
+    @Test
+    fun `invoke() returns access token from successful response`() = runTest {
```
sshropshire: While it's unlikely, I think we should add a test for when `access_token` is missing. I think right now it'll throw a `JSONException`.

CorePayments/src/main/java/com/paypal/android/corepayments/api/FetchClientToken.kt
kgangineni: renaming to `getToken()` because same token is returned after it's created

CorePayments/src/main/java/com/paypal/android/corepayments/api/AuthenticationSecureTokenServiceAPI.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.paypal.android.corepayments.RestClient
+import com.paypal.android.corepayments.base64encoded
+import org.json.JSONObject
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class AuthenticationSecureTokenServiceAPI(
+    private val coreConfig: CoreConfig,
+    private val restClient: RestClient = RestClient(configuration = coreConfig),
+) {
+    suspend fun getClientToken(): GetClientTokenResult {
```
KunJeongPark: Are these called client tokens?

CorePayments/src/main/java/com/paypal/android/corepayments/api/FetchClientToken.kt
sshropshire: It isn't a new token every time? Also, I went with "create" in the naming prefix because the underlying HTTP method is a `POST`.

CorePayments/src/main/java/com/paypal/android/corepayments/api/AuthenticationSecureTokenServiceAPI.kt
KunJeongPark: @sshropshire Unrelated, but wee put the UpdateClientConfigAPI` class in `PayPalWebCheckout `module for this draft PR:`
https://github.com/paypal/paypal-android/pull/314 and iOS version also, but I think we should move it to `CorePayments` as well, don't you think? I think originally we thought that this would be only for PayPalWebCheckout module, but since we have make that call everywhere, should live in CorePayments, right?

KunJeongPark: just one field I think would be dynamic.

sshropshire: Yeah `UpdateClientConfigAPI` sounds like it should live in the core module, that makes sense to me.

kgangineni: Are these called client tokens?
> @KunJeongPark technically it's access token, may be go full descriptive? `getLowScopedAccessToken`

KunJeongPark: that sounds good to me. What do you think @sshropshire ?

sshropshire: I do like `getLowScopedAccessToken()` it feels like the most expressive name.

CorePayments/src/main/java/com/paypal/android/corepayments/api/AuthenticationSecureTokenServiceAPI.kt
```diff
@@ -0,0 +1,65 @@
...
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class AuthenticationSecureTokenServiceAPI(
+    private val coreConfig: CoreConfig,
+    private val restClient: RestClient = RestClient(configuration = coreConfig),
+) {
+    suspend fun getLowScopedAccessToken(): GetClientTokenResult {
+        val requestBody = "grant_type=client_credentials&response_type=token"
+
+        val headers = mutableMapOf(
+            "Authorization" to "Basic ${coreConfig.clientId.base64encoded()}",
```
sshropshire:
```suggestion
            "Authorization" to "Basic ${coreConfig.clientId.base64encoded()}:",
```
nit: does this work without the colon at the end? I remember we used to do that when we had LSATs in the past.

CorePayments/src/main/java/com/paypal/android/corepayments/APIClientError.kt
```diff
@@ -91,4 +92,16 @@ object APIClientError {
...
+        code: Int,
+        errorDescription: String?,
+        correlationId: String? = null,
+        reason: Throwable? = null
+    ): PayPalSDKError = PayPalSDKError(
+        code = code,
+        errorDescription = errorDescription.orEmpty(),
+        correlationId = correlationId,
+        reason = reason
+    )
```
sshropshire: 🧶 : do we need this? Why don't we call the `PayPalSDKError` constructor directly in this scenario.

CorePayments/src/main/java/com/paypal/android/corepayments/api/AuthenticationSecureTokenServiceAPI.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.paypal.android.corepayments.RestClient
+import com.paypal.android.corepayments.base64encoded
+import org.json.JSONObject
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class AuthenticationSecureTokenServiceAPI(
+    private val coreConfig: CoreConfig,
+    private val restClient: RestClient = RestClient(configuration = coreConfig),
+) {
+    suspend fun getLowScopedAccessToken(): GetClientTokenResult {
```
sshropshire:
```suggestion
    suspend fun createLowScopedAccessToken(): GetClientTokenResult {
```
nit: Since we're technically creating a token rather than retrieving one that already exists.

CorePayments/src/main/java/com/paypal/android/corepayments/api/GetClientTokenResult.kt
```diff
@@ -0,0 +1,21 @@
+package com.paypal.android.corepayments.api
+
+import androidx.annotation.RestrictTo
+import com.paypal.android.corepayments.PayPalSDKError
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+sealed class GetClientTokenResult {
```
sshropshire:
```suggestion
sealed class CreateLowScopedAccessTokenResult {
```

🧶 : to align result name with function method name.

sshropshire (approved):
Couple take or leave 🧶 's but overall LGTM

PR #331, opened 2025-07-02
Integration with App Switch Eligibility, Launching app switch

kgangineni:
### Summary of changes

1. Introduces Boolean `appSwitchWhenEligible`, allowing the merchants to opt-in app switch flow
2. Makes patchCCOWithAppSwitchEligibility call to verify eligibility
3. Introduces callback for Vault and Checkout flows to enable callback flow for app switch

 ### Checklist

 - [ ] Added a changelog entry

-


CorePayments/src/main/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibility.kt
```diff
@@ -0,0 +1,76 @@
...
+        orderId: String,
+        tokenType: TokenType,
+        merchantOptInForAppSwitch: Boolean,
+        paypalNativeAppInstalled: Boolean
+    ): APIResult<PatchCcoWithAppSwitchEligibilityResponse> {
+        return runCatching {
+            val token =
+                when (val tokenResult = authenticationSecureTokenServiceAPI.createLowScopedAccessToken()) {
+                    is APIResult.Success -> tokenResult.data
+                    is APIResult.Failure -> return APIResult.Failure(tokenResult.error)
```
sshropshire: 🧶 : Is there a way to clean up this return statement? I understand what it's doing, but it seems sort of like an escape hatch. It might read a lot cleaner if we make the output of `when` be the return value of the `runCatching` block, rather than assigning the result of `when(){}` to `token`.

CorePayments/src/main/java/com/paypal/android/corepayments/model/PatchCcoWithAppSwitchEligibilityRequest.kt
```diff
@@ -0,0 +1,69 @@
...
+            val query = when (val result = resourceLoader.loadRawResource(context, resId)) {
+                is LoadRawResourceResult.Success -> result.value
+
+                is LoadRawResourceResult.Failure -> throw APIClientError.graphQLRequestLoadError()
+            }
+        val variablesJson = createVariablesJson(variables)
+        return JSONObject()
+            .put("query", query)
+            .put("variables", variablesJson)
+    }
```
sshropshire: I like this pattern. What if we move `create()` into a `companion object`? It doesn't seem to rely on instance state. Then `PatchCcoWithAppSwitchEligibilityRequest` can be a plain data object.

CorePayments/src/main/java/com/paypal/android/corepayments/model/PatchCcoWithAppSwitchEligibilityResponse.kt
```diff
@@ -0,0 +1,58 @@
+package com.paypal.android.corepayments.model
+
+import com.paypal.android.corepayments.APIClientError
+import org.json.JSONObject
+
+data class PatchCcoWithAppSwitchEligibilityResponse(
```
sshropshire: q: Can we make these data objects be `internal`? Merchants shouldn't need to create these types.

CorePayments/src/main/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibility.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    constructor(coreConfig: CoreConfig) : this(
+        authenticationSecureTokenServiceAPI = AuthenticationSecureTokenServiceAPI(
+            coreConfig,
+            RestClient(coreConfig)
+        ),
+        graphQLClient = GraphQLClient(coreConfig)
+    )
+
+    suspend operator fun invoke(
```
sshropshire, line 33: 🧶: Should we use the [UseCase suffix naming convention](https://developer.android.com/topic/architecture/domain-layer#conventions) for this class?

CorePayments/src/test/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibilityUnitTest.kt
```diff
@@ -0,0 +1,302 @@
...
+        val headersSlot = slot<Map<String, String>>()
+
+        // When
+        sut.invoke(
+            context = context,
+            orderId = testOrderId,
+            tokenType = TokenType.ORDER_ID,
+            merchantOptInForAppSwitch = true,
+            paypalNativeAppInstalled = true
+        )
```
sshropshire:
```suggestion
        sut(
            context = context,
            orderId = testOrderId,
            tokenType = TokenType.ORDER_ID,
            merchantOptInForAppSwitch = true,
            paypalNativeAppInstalled = true
        )
```
🧶 : We shouldn't need to explicitly call `invoke` here, we can just use the `()` invocation operator.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -58,31 +58,49 @@ class PayPalWebCheckoutClient internal constructor(
...
-                        analytics.notify(
-                            CheckoutEvent.AUTH_CHALLENGE_PRESENTATION_FAILED,
-                            checkoutOrderId
-                        )
-                }
-                result
+            // if RedirectUrl not available fallback to web checkout
+            if (!launchUrl.isNullOrEmpty()) {
+                payPalWebLauncher.launchWithUrl(
+                    activity,
```
sshropshire, line 78:
```suggestion
                    activity.applicationContext,
```
+1

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -110,14 +128,43 @@ class PayPalWebCheckoutClient internal constructor(
      *
      * @param request [PayPalWebVaultRequest] for vaulting PayPal as a payment method
      */
-    fun vault(
+    suspend fun vault(
```
sshropshire, line 119: We may need a `CHANGELOG` entry for this breaking change.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -110,14 +128,43 @@ class PayPalWebCheckoutClient internal constructor(
...
+            )
+            // Get RedirectUrl
+            val launchUrl = when (patchCcoResult) {
+                is APIResult.Success -> patchCcoResult.data.launchUrl
+                is APIResult.Failure -> null
+            }
+            // if RedirectUrl not available fallback to web vault
+            if (!launchUrl.isNullOrEmpty()) {
+                payPalWebLauncher.launchWithUrl(
+                    activity,
```
sshropshire, line 155:
```suggestion
                    activity.applicationContext,
```

sshropshire: Oh wait my bad, I forgot that BrowserSwitch requires an activity reference. Please disregard.

CorePayments/src/main/java/com/paypal/android/corepayments/model/PatchCcoWithAppSwitchEligibilityRequest.kt
kgangineni: That was the initial implementation, but variables made sense to be in the data class, that's the instance state used

CorePayments/src/main/java/com/paypal/android/corepayments/model/PatchCcoWithAppSwitchEligibilityResponse.kt
kgangineni: response is used in `PaypalWebCheckoutClient` but made it restricted to Library group

CorePayments/src/main/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibility.kt
kgangineni: I'm kind of in between, if in demo app definitely useCase
CoPilot response: 
Use PatchCCOWithAppSwitchEligibility for a straightforward implementation.
Use PatchCCOWithAppSwitchEligibilityUseCase if it is in layered clean architecture

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
 # PayPal Android SDK Release Notes
 
+```kotlin
```
sshropshire: This should probably be in a migration guide rather than the CHANGELOG. We don't typically place code snippets in our release notes.

CorePayments/src/main/java/com/paypal/android/corepayments/model/PatchCcoWithAppSwitchEligibilityRequest.kt
```diff
@@ -0,0 +1,69 @@
...
+            .put("contextId", variables.contextId)
+            .put("token", variables.token)
+            .put("osType", OS_TYPE)
+            .put("merchantOptInForAppSwitch", variables.merchantOptInForAppSwitch)
+            .put("paypalNativeAppInstalled", variables.paypalNativeAppInstalled)
+    }
+
+    companion object {
+        const val INTEGRATION_ARTIFACT = "NATIVE_SDK" // todo: use Mobile SDK artifact after backend changes
+        const val INTEGRATION_CHANNEL = "PPCP_NATIVE_SDK"
```
KunJeongPark, line 51: would this translate to CCO field "channel" or is this something else?
This reminds me, we should request this right away to XOOS, to accommodate our integration artifact value

If so, I think our plan was to use "MOBILE_APP"

CorePayments/src/main/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibility.kt
sshropshire: Which do you think is best going forward?

KunJeongPark:
@kgangineni Are we incorporating client token usage for` PatchCCOWithAppSwitchEligibility` instead of LSAT in a later PR?

CorePayments/src/main/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibility.kt
```diff
@@ -0,0 +1,76 @@
...
+        orderId: String,
+        tokenType: TokenType,
+        merchantOptInForAppSwitch: Boolean,
+        paypalNativeAppInstalled: Boolean
+    ): APIResult<PatchCcoWithAppSwitchEligibilityResponse> {
+        return runCatching {
+            val token =
+                when (val tokenResult = authenticationSecureTokenServiceAPI.createLowScopedAccessToken()) {
+                    is APIResult.Success -> tokenResult.data
+                    is APIResult.Failure -> throw tokenResult.error
```
sshropshire: qq: why do we need `runCatching` here? It doesn't seem like we're using any Java APIs.

CorePayments/src/main/java/com/paypal/android/corepayments/common/Headers.kt
```diff
@@ -0,0 +1,7 @@
+package com.paypal.android.corepayments.common
+
+object Headers {
```
sshropshire:
```suggestion
internal object Headers {
```

CorePayments/src/main/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibility.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.paypal.android.corepayments.graphql.GraphQLResult
+import com.paypal.android.corepayments.model.APIResult
+import com.paypal.android.corepayments.model.ExperimentationContext
+import com.paypal.android.corepayments.model.ExternalResponse
+import com.paypal.android.corepayments.model.PatchCcoWithAppSwitchEligibilityRequest
+import com.paypal.android.corepayments.model.PatchCcoWithAppSwitchEligibilityResponse
+import com.paypal.android.corepayments.model.TokenType
+import com.paypal.android.corepayments.model.Variables
+
+class PatchCCOWithAppSwitchEligibility internal constructor(
```
sshropshire, line 20:
```suggestion
@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
class PatchCCOWithAppSwitchEligibility internal constructor(
```

CorePayments/src/main/java/com/paypal/android/corepayments/model/TokenType.kt
```diff
@@ -0,0 +1,8 @@
+package com.paypal.android.corepayments.model
+
+enum class TokenType {
```
sshropshire, line 6:
```suggestion
@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
enum class TokenType {
```

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -58,31 +58,49 @@ class PayPalWebCheckoutClient internal constructor(
...
+                context = activity,
+                orderId = request.orderId,
+                tokenType = TokenType.ORDER_ID,
+                merchantOptInForAppSwitch = request.appSwitchWhenEligible,
+                paypalNativeAppInstalled = true // TODO: implement native app installed check
+            )
+            // Get RedirectUrl
+            val launchUrl = when (patchCcoResult) {
+                is APIResult.Success -> patchCcoResult.data.launchUrl
+                is APIResult.Failure -> null
```
sshropshire:
```suggestion
                is APIResult.Failure -> buildFallbackURL()
```
🧶 : We could build the fallback URL here to make the implementation cleaner.

kgangineni:
> @kgangineni Are we incorporating client token usage for` PatchCCOWithAppSwitchEligibility` instead of LSAT in a later PR?
We will switch to client-token after we make decision, it will be big change, not sure if we will go with LSAT for initial release, then make changes to client-token for whole SDK

CorePayments/src/main/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibility.kt
kgangineni: I think better to go with PatchCCOWithAppSwitchEligibility pattern, inside sdk, we can call it like a function

CorePayments/src/main/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibility.kt
kgangineni: we're not using JAVA APIs, this is only to make sure to send result as `APIResult.Failure` rather than throwing exception

sshropshire: Yeah that makes sense. If we aren't using any methods that `throw` can we remove `runCatching`?

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
kgangineni: ended cleaning up more, moved logic of buildUrl to `PayPalWebCheckoutClient` from `PayPalWebLauncher`
let me know what u think

CorePayments/src/main/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibility.kt
kgangineni: not just throwing error programatically, more of safe catching for unknown errors that we didn't thought of, kind of last catch block as fallback

CorePayments/src/main/java/com/paypal/android/corepayments/model/PatchCcoWithAppSwitchEligibilityRequest.kt
kgangineni: `MOBILE_APP`? in Jira it's mentioned as MOBILE_SDK, guess whatever changes u made for updateConfig, we will go with same

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # PayPal Android SDK Release Notes
 
+* added overloaded functions with callbacks for `start` and `vault` methods in
```
sshropshire:
```suggestion
## unreleased
* added overloaded functions with callbacks for `start` and `vault` methods in
```

KunJeongPark:
> > @kgangineni Are we incorporating client token usage for` PatchCCOWithAppSwitchEligibility` instead of LSAT in a later PR?
> > We will switch to client-token after we make decision, it will be big change, not sure if we will go with LSAT for initial release, then make changes to client-token for whole SDK

I think the decision was to use client token for v3-beta and per our last discussion with Nirvan, the v3-beta was going to have both button changes and app switch. But we can discuss that some more with Aisha and Ed.
I think v3-beta is a good place to release these big changes, I'd love to hear everyone's thoughts.

CorePayments/src/main/java/com/paypal/android/corepayments/model/PatchCcoWithAppSwitchEligibilityRequest.kt
KunJeongPark: Yes, those values were changed in meeting with Nirvan and risk and compliance.
does that INTEGRATION_CHANNEL map to channel CCO field?
I didn't know that there was PPCP_NATIVE_SDK
Did you try running this with MOBILE_SDK and see if the graphQL fails or not?

kgangineni: there is `PPCP_NATIVE_SDK` in graphql schema, `MOBILE_SDK` is not there, Yes I tried running, it gives graphql validation error
I will raise PR to add it to schema

CorePayments/src/main/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibility.kt
sshropshire: I do think we should return an `APIResult` here. It simplifies the API for the caller, they will only need a single `when` statement instead of a combination `when` + `try/catch`.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -190,4 +229,69 @@ class PayPalWebCheckoutClient internal constructor(
...
+            when (patchCcoResult) {
+                is APIResult.Success -> patchCcoResult.data.launchUrl?.toUri() ?: fallbackUri
+                is APIResult.Failure -> fallbackUri
+            }
+        } else {
+            fallbackUri
+        }
+    }
+
+    private fun launchWithCallback(block: suspend () -> Unit) {
```
sshropshire: Does the block need to suspend here? I'd like to try removing it to simplify the code.

kgangineni: yes, purpose is to take suspend function and make it callback so it can be called with java

CorePayments/src/main/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibility.kt
kgangineni: removed `runCatching` & some of bloat in parsing logic, made it private function

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # PayPal Android SDK Release Notes
 
+## unreleased
+* added overloaded functions with callbacks for `start` and `vault` methods in
+  `PayPalWebCheckoutClient`
+* modified `PayPalWebCheckoutRequest` to include `appSwitchWhenEligible` property to control app
+  switch behavior
+* Breaking Changes
+  * made start and vault functions in `PayPalWebCheckoutClient` suspend functions
```
sshropshire:
```suggestion
* Add overloaded functions with callbacks for `start` and `vault` methods in
  `PayPalWebCheckoutClient`
* Modify `PayPalWebCheckoutRequest` to include `appSwitchWhenEligible` property to control app
  switch behavior
* Breaking Changes
  * Make start and vault functions in `PayPalWebCheckoutClient` suspend functions
```
Placing these in the [imperative mood](https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/git-commit-message-imperative-mood-past-tense-convention-standards-best-practice-subject) to align with git commit message style.

CorePayments/src/main/java/com/paypal/android/corepayments/graphql/GraphQLClient.kt
```diff
@@ -36,13 +32,18 @@ class GraphQLClient internal constructor(
...
-    suspend fun send(graphQLRequestBody: JSONObject, queryName: String? = null): GraphQLResult {
+    suspend fun send(
+        graphQLRequestBody: JSONObject,
+        queryName: String? = null,
+        headers: Map<String, String>? = null
+    ): GraphQLResult {
         val body = graphQLRequestBody.toString()
         val urlString = if (queryName != null) "$graphQLURL?$queryName" else graphQLURL
         val httpRequest = HttpRequest(URL(urlString), HttpMethod.POST, body, httpRequestHeaders)
+        headers?.forEach { (key, value) -> httpRequestHeaders.put(key, value) }
```
sshropshire:
```suggestion
        headers?.forEach { (key, value) -> httpRequestHeaders.put(key, value) }
        val httpRequest = HttpRequest(URL(urlString), HttpMethod.POST, body, httpRequestHeaders)
```
I feel like it makes sense to add headers before constructing the `HttpRequest`. This will prevent bugs in the future if the `HttpRequest` constructor one day makes its own copy of `httpRequestHeaders`.

CorePayments/src/main/java/com/paypal/android/corepayments/model/PatchCcoWithAppSwitchEligibilityRequest.kt
```diff
@@ -0,0 +1,69 @@
...
+) {
+    suspend fun create(
+        context: Context,
+        resourceLoader: ResourceLoader = ResourceLoader()
+    ): JSONObject {
+        @RawRes val resId = R.raw.graphql_query_patch_cco_app_switch_eligibility
+            val query = when (val result = resourceLoader.loadRawResource(context, resId)) {
+                is LoadRawResourceResult.Success -> result.value
+
+                is LoadRawResourceResult.Failure -> throw APIClientError.graphQLRequestLoadError()
```
sshropshire: Can we replace `throw` here with a result type?

CorePayments/src/main/java/com/paypal/android/corepayments/model/PatchCcoWithAppSwitchEligibilityRequest.kt
```diff
@@ -0,0 +1,69 @@
...
+            .put("tokenType", variables.tokenType)
+            .put("contextId", variables.contextId)
+            .put("token", variables.token)
+            .put("osType", OS_TYPE)
+            .put("merchantOptInForAppSwitch", variables.merchantOptInForAppSwitch)
+            .put("paypalNativeAppInstalled", variables.paypalNativeAppInstalled)
+    }
+
+    companion object {
+        const val INTEGRATION_ARTIFACT = "NATIVE_SDK" // todo: use Mobile SDK artifact after backend changes
```
sshropshire:
```suggestion
        const val INTEGRATION_ARTIFACT = "NATIVE_SDK" // TODO: use Mobile SDK artifact after backend changes
```

CorePayments/src/test/java/com/paypal/android/corepayments/api/AuthenticationSecureTokenServiceAPIUnitTest.kt
```diff
@@ -62,8 +63,8 @@ class AuthenticationSecureTokenServiceAPIUnitTest {
         assertEquals("application/x-www-form-urlencoded", headers["Content-Type"])
         assert(headers["Authorization"]!!.startsWith("Basic "))
```
sshropshire:
```suggestion
        assertTrue(headers["Authorization"]!!.startsWith("Basic "))
```
I noticed this is Java `assert` rather than a JUnit `assertTrue`.

CorePayments/src/test/java/com/paypal/android/corepayments/api/AuthenticationSecureTokenServiceAPIUnitTest.kt
```diff
@@ -62,8 +63,8 @@ class AuthenticationSecureTokenServiceAPIUnitTest {
...
 
-            assertTrue(result is CreateLowScopedAccessTokenResult.Success)
-            assertEquals("test-token", (result as CreateLowScopedAccessTokenResult.Success).token)
+            assertTrue(result is APIResult.Success)
```
sshropshire, line 66: We may need to auto format this file. The indentation seems to be off a bit on some lines.

CorePayments/src/test/java/com/paypal/android/corepayments/api/AuthenticationSecureTokenServiceAPIUnitTest.kt
```diff
@@ -97,7 +98,7 @@ class AuthenticationSecureTokenServiceAPIUnitTest {
         val result = sut.createLowScopedAccessToken()
 
         // Then
-        assertTrue(result is CreateLowScopedAccessTokenResult.Failure)
+        assertTrue(result is APIResult.Failure)
```
sshropshire, line 101: Should we also assert the contents of the error message?

CorePayments/src/test/java/com/paypal/android/corepayments/api/AuthenticationSecureTokenServiceAPIUnitTest.kt
```diff
@@ -220,11 +221,11 @@ class AuthenticationSecureTokenServiceAPIUnitTest {
         val authHeader = requestSlot.captured.headers!!["Authorization"]!!
         assert(authHeader.startsWith("Basic "))
```
sshropshire:
```suggestion
        assertTrue(authHeader.startsWith("Basic "))
```

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
sshropshire: Do we need this method then? It seems like it does the exact same thing as
```kotlin
CoroutineScope(mainDispatcher).launch {
}
```
There's a slight readability gain for `launchWithCallback`, but when I see `CoroutineScope.launch()` I know exactly what's happening.

kgangineni: I guess we can remove it for now, it's only used in this file

CorePayments/src/main/java/com/paypal/android/corepayments/model/PatchCcoWithAppSwitchEligibilityRequest.kt
kgangineni: returning nullable instead of throwing error

CorePayments/src/main/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibility.kt
```diff
@@ -0,0 +1,108 @@
...
+    suspend operator fun invoke(
+        context: Context,
+        orderId: String,
+        tokenType: TokenType,
+        merchantOptInForAppSwitch: Boolean,
+        paypalNativeAppInstalled: Boolean
+    ): APIResult<AppSwitchEligibility> {
+        val tokenResult = authenticationSecureTokenServiceAPI.createLowScopedAccessToken()
+        if (tokenResult is APIResult.Failure) {
+            return APIResult.Failure(tokenResult.error)
```
sshropshire, line 42:
```suggestion
            return tokenResult
```
nit: We could technically forward the result in this case.

CorePayments/src/main/java/com/paypal/android/corepayments/model/PatchCcoWithAppSwitchEligibilityRequest.kt
```diff
@@ -0,0 +1,68 @@
...
+    private fun createVariablesJson(variables: Variables): JSONObject {
+        val experimentationContext = JSONObject()
+            .put("integrationChannel", variables.experimentationContext.integrationChannel)
+
+        return JSONObject()
+            .put("experimentationContext", experimentationContext)
+            .put(
+                "integrationArtifact",
+                INTEGRATION_ARTIFACT
+            )
```
sshropshire, line 39:
```suggestion
            .put("integrationArtifact", INTEGRATION_ARTIFACT)
```
nit: to help with readability.

sshropshire (approved):
LGTM

CorePayments/src/main/java/com/paypal/android/corepayments/model/PatchCcoWithAppSwitchEligibilityRequest.kt
KunJeongPark: I am curious about the mapping of the fields here
I know integration artifact, that we need to raise the issue for this and you've already done that.
but not sure what` integration channel` is
that is different from xobuyernodeserv's UpdateClientConfig.
Is there an equivalent to channel for this endpoint?
Yes, I see that it belongs to the experimentationContext, separate from CCO values. 

I know that` xorouternodeweb` is looking for correct integration artifact value (MOBILE_SDK)
to determine whether to respect our CCO values. So I think that maybe related to
why we are not seeing these values persisted in session explorer for this flow.

CorePayments/src/main/res/raw/graphql_query_patch_cco_app_switch_eligibility.graphql
```diff
@@ -0,0 +1,34 @@
...
+        experimentationContext: $experimentationContext
+        merchantOptInForAppSwitch: $merchantOptInForAppSwitch
+        osType: $osType
+        token: $token
+        tokenType: $tokenType
+      }
+      patchCcoInput: {
+        token: $token
+        clientConfig: {
+          integrationArtifact: $integrationArtifact
```
KunJeongPark, line 23: I think here is where you would also set channel: MOBILE_APP, userExperienceFlow: INCONTEXT, productFlow: HERMES.
I think product flow, that's the defaulted value in CSPS, but I just made it explicit in iOS just to be safe

CorePayments/src/main/res/raw/graphql_query_patch_cco_app_switch_eligibility.graphql
```diff
@@ -0,0 +1,34 @@
...
+  $token: externalToken!
+  $tokenType: externalTokenType!
+  $integrationArtifact: externalIntegrationArtifactType!
+) {
+  external {
+    patchCcoWithAppSwitchEligibility(
+      appSwitchEligibilityInput: {
+        contextId: $contextId
+        experimentationContext: $experimentationContext
+        merchantOptInForAppSwitch: $merchantOptInForAppSwitch
```
KunJeongPark, line 15: I don't see any reference for experimentationContext or merchantOptInForAppSwitch
Are there docs for these?

PR #332, opened 2025-07-08
Make CCO Call for all PayPal Web Checkout Funding Sources

sshropshire:
### Summary of changes

 - This PR moves `UpdateClientConfigAPI` to `CorePayments` so that it is accessible from all modules.
 - This PR also updates CCO for all PayPalWebCheckout funding sources

 ### Checklist

- @sshropshire


CorePayments/src/main/java/com/paypal/android/corepayments/UpdateClientConfigAPI.kt
```diff
@@ -1,42 +1,42 @@
...
     private val coreConfig: CoreConfig,
     private val applicationContext: Context,
     private val graphQLClient: GraphQLClient,
     private val resourceLoader: ResourceLoader
 ) {
 
+    private object Defaults {
+        const val INTEGRATION_ARTIFACT = "MOBILE_SDK"
+        const val USER_EXPERIENCE_FLOW = "INCONTEXT"
+        const val PRODUCT_FLOW = "MOBILE_SDK"
```
KunJeongPark: We landed on HERMES for product flow

KunJeongPark:
Steven, I'm a bit confused, are you going to replace https://github.com/paypal/paypal-android/pull/314
with this PR and another one for the card button?

CorePayments/src/main/java/com/paypal/android/corepayments/UpdateClientConfigAPI.kt
KunJeongPark: I also added channel to be explicit. 
We had relied on defaults in XOR before but I wanted to be as explicit as possible
because they are respecting our CCO values sent in from us now. (I think on version soon to be released)

sshropshire:
> Steven, I'm a bit confused, are you going to replace #314 with this PR and another one for the card button?

This PR is going to merge into that one. We will still need a complete overview of the BCDC feature.

CorePayments/src/main/java/com/paypal/android/corepayments/UpdateClientConfigAPI.kt
sshropshire: Ok is Nirvan aware of this change? The ticket has `MOBILE_SDK` for the `PRODUCT_FLOW` field.

KunJeongPark: Yes, I think the original ticket was made a while ago. And Nirvan was involved in all of the discussions with those teams. Apologies, I didn't realize that the ticket was not updated.

We are going to do some sandbox testing once XOR changes are released (It'd be on experiment in Sandbox until we ok it) and I will see if we need to send in channel explicitly then.

KunJeongPark:
Are you including the paypal-context-id in the header?
In iOS, we are passing it in in the networkClient.fetch in UpdateClientConfigAPI.

sshropshire:
> Are you including the paypal-context-id in the header?

Yes we're also setting it here in [UpdateClientConfigAPI.kt](https://github.com/paypal/paypal-android/blob/090be286d0624f98e1a9ff905861123a0c68a79d/CorePayments/src/main/java/com/paypal/android/corepayments/UpdateClientConfigAPI.kt#L67)

sshropshire:
@KunJeongPark that's a good catch though, the header name is incorrect. I'll update.

sshropshire:
@KunJeongPark actually should it be `paypal-context-id`? I noticed the header is also named `paypal-client-context` on iOS.

KunJeongPark:
> @KunJeongPark actually should it be `paypal-context-id`? I noticed the header is also named `paypal-client-context` on iOS.

you are right, I misspoke

CorePayments/src/main/java/com/paypal/android/corepayments/UpdateClientConfigAPI.kt
```diff
@@ -1,42 +1,42 @@
...
+class UpdateClientConfigAPI(
     private val coreConfig: CoreConfig,
     private val applicationContext: Context,
     private val graphQLClient: GraphQLClient,
     private val resourceLoader: ResourceLoader
 ) {
 
+    private object Defaults {
+        const val INTEGRATION_ARTIFACT = "MOBILE_SDK"
+        const val USER_EXPERIENCE_FLOW = "INCONTEXT"
```
kgangineni, line 24: USER_EXPERIENCE_FLOW is this value not `NATIVE`? because flow is initiated by native app, even though it's completed in web 
Q: `HERMES` what does this mean?

sshropshire: I believe `HERMES` is an internal name I'm not 100% sure of its origin. All of the values in `Defaults` come from the JIRA ticket I was assigned. Nirvan may have further background on the actual meaning of each.

KunJeongPark: I think we took JS SDK's INCONTEXT value as web view and meant to distinguish native experiences like Apple Pay and Venmo as NATIVE.

KunJeongPark: HERMES is name for old checkout experience, I think but here it is just default value most flows have for product_flow.
From what I understand, product_flow refers to a specific flow within your SDK and I don't think it applies to our SDK
so we were advised to just go with the default.

PR #333, opened 2025-07-10
UI and create order changes required for app switch in Demo app

kgangineni:
### Summary of changes

1. changes required to add return_url, cancel_url in create order request body
2. send app_url when app switch is enabled in create order request body and setup token request body
3. UI changes to opt-in or opt-out of app switch before creating order


Demo/src/main/java/com/paypal/android/ui/paypalwebvault/PayPalWebVaultView.kt
```diff
@@ -28,7 +32,7 @@ import com.paypal.android.utils.UIConstants
 import com.paypal.android.utils.getActivityOrNull
 
 @Composable
-fun PayPalWebVaultView(viewModel: PayPalWebVaultViewModel = hiltViewModel()) {
+fun PayPalWebVaultView(viewModel: PayPalVaultViewModel = hiltViewModel()) {
```
sshropshire:
```suggestion
fun PayPalVaultView(viewModel: PayPalVaultViewModel = hiltViewModel()) {
```
I like shortening the prefix. We'll have to look into doing this in the SDK as well. I think it makes sense to simplify the naming scheme as much as possible.

Demo/src/main/java/com/paypal/android/usecase/CreateOrderUseCase.kt
```diff
@@ -42,6 +42,28 @@ class CreateOrderUseCase @Inject constructor(
...
+                    .put("cancel_url", CANCEL_URL)
+                    .put("native-app", nativeAppJSON)
+
+                val paypalJSON = JSONObject()
+                    .put("experience_context", experienceContextJSON)
+
+                val paymentSourceJSON = JSONObject()
+                    .put("paypal", paypalJSON)
+
+                orderRequest.put("payment_source", paymentSourceJSON)
```
sshropshire, line 64: It would be really cool to go full GSON. I think this portion of the JSON is one area where GSON support will be difficult. PPaaS schemas aren't the most GSON friendly.

Demo/src/main/java/com/paypal/android/usecase/CreateOrderUseCase.kt
```diff
@@ -42,6 +42,28 @@ class CreateOrderUseCase @Inject constructor(
...
             }
+            if (request.appSwitchWhenEligible) {
+
+                val nativeAppJSON = JSONObject()
+                    .put("app_url", APP_URL)
+
+                val experienceContextJSON = JSONObject()
+                    .put("return_url", SUCCESS_URL)
+                    .put("cancel_url", CANCEL_URL)
+                    .put("native-app", nativeAppJSON)
```
sshropshire:
```suggestion
                    .put("native_app", nativeAppJSON)
```
qq: Is this not an underscore as well?

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalCheckoutView.kt
```diff
@@ -58,23 +61,40 @@ fun PayPalWebView(
...
+            Step4_CompleteOrder(uiState, viewModel)
         }
         Spacer(modifier = Modifier.size(contentPadding))
     }
 }
 
 @Composable
-private fun Step1_CreateOrder(uiState: PayPalWebUiState, viewModel: PayPalWebViewModel) {
+fun Step1_WebOrNativeCheckoutHeader(viewModel: PayPalCheckoutViewModel) {
+    val uiState: PayPalUiState by viewModel.uiState.collectAsStateWithLifecycle()
```
sshropshire: Any reason we create another StateFlow here? I wonder if this could cause a double render. A lot of the demo app is inspired by this [NowInAndroid app](https://github.com/android/nowinandroid/tree/main). You can see here the `collectAsStateWithLifecycle()` methods are only called [at the top level.](https://github.com/android/nowinandroid/blob/main/feature/foryou/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/ForYouScreen.kt#L113)

Demo/src/main/java/com/paypal/android/usecase/CreateOrderUseCase.kt
kgangineni: it is underscore, corrected,

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalCheckoutView.kt
kgangineni: ya, it's not needed, corrected

sshropshire (approved):
🔥

KunJeongPark (approved):

PR #334, opened 2025-07-14
Check for Paypal Mobile app before app switch

kgangineni:
### Summary of changes
 - checks if paypal app is installed
 - PatchCCOWithAppSwitchEligibility only if paypal app is installed
 - go with web checkout flow when app is not installed

 ### Checklist

 - [ ] Added a changelog entry

-


Demo/src/main/java/com/paypal/android/usecase/CreatePayPalSetupTokenUseCase.kt
```diff
@@ -23,28 +19,14 @@ class CreatePayPalSetupTokenUseCase @Inject constructor(
                 if (appSwitchEnabled) {
 
                     val nativeAppJSON = JSONObject()
-                        .put("os_type", "ANDROID")
-                        .put("os_version", "35")
-
-                    val appSwitchPreferenceJSON = JSONObject()
-//                        .put("app_url", app_url)
-//                        .put("launch_paypal_app", true)
```
sshropshire: Is this file supposed to be modified in this PR? It doesn't seem to be related to app switching.

kgangineni: it should not be, reverted, 🙇

CorePayments/src/main/java/com/paypal/android/corepayments/api/AuthenticationSecureTokenServiceAPI.kt
```diff
@@ -9,19 +9,21 @@ import com.paypal.android.corepayments.PayPalSDKError
...
     private val restClient: RestClient = RestClient(configuration = coreConfig),
 ) {
-    suspend fun createLowScopedAccessToken(): CreateLowScopedAccessTokenResult {
+    suspend fun createLowScopedAccessToken(): APIResult<String> {
         val requestBody = "grant_type=client_credentials&response_type=token"
 
         val headers = mutableMapOf(
-            "Authorization" to "Basic ${coreConfig.clientId.base64encoded()}:",
-            "Content-Type" to "application/x-www-form-urlencoded"
+            Headers.AUTHORIZATION to "Basic ${coreConfig.clientId.base64encoded()}",
```
sshropshire: We are missing the trailing colon here. It's [technically required](https://datatracker.ietf.org/doc/html/rfc2617#section-2) in the spec. It's safer to include it now and be compliant to avoid the risk of a future outage.

CorePayments/src/main/java/com/paypal/android/corepayments/api/PatchCCOWithAppSwitchEligibility.kt
```diff
@@ -0,0 +1,76 @@
...
+    )
+
+    suspend operator fun invoke(
+        context: Context,
+        orderId: String,
+        tokenType: TokenType,
+        merchantOptInForAppSwitch: Boolean,
+        paypalNativeAppInstalled: Boolean
+    ): APIResult<PatchCcoWithAppSwitchEligibilityResponse> {
+        return runCatching {
```
sshropshire: We should be able to remove `runCatching` here since we no longer throw?

Demo/src/main/java/com/paypal/android/api/model/PaymentSource.kt
```diff
@@ -0,0 +1,39 @@
...
+import kotlinx.parcelize.Parcelize
+import org.json.JSONObject
+
+@Parcelize
+data class PaymentSource(
+    val paypal: PayPalDetails? = null
+) : Parcelable {
+    companion object {
+
+        internal fun fromJson(jsonObject: JSONObject): PaymentSource? {
```
sshropshire: I'm still not a huge fan of `fromJson()` methods because the method itself doesn't say much about where the JSON came from. Can we move this parsing logic somewhere external to this file? It would be nice if this file contained pure data objects without any `companion` logic.

Demo/src/main/java/com/paypal/android/api/services/SDKSampleServerAPI.kt
```diff
@@ -234,15 +236,7 @@ class SDKSampleServerAPI {
...
-
-        val responseJSON = JSONObject(response.string())
-        val customerJSON = responseJSON.getJSONObject("customer")
-
-        PayPalSetupToken(
-            id = responseJSON.getString("id"),
-            customerId = customerJSON.getString("id"),
-            status = responseJSON.getString("status")
-        )
+        Gson().fromJson(response.string(), PayPalSetupToken::class.java)
```
sshropshire: Can we revert `Gson` parsing in this PR and create a ticket to fully adopt `Gson` in the demo app?

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalCheckoutView.kt
```diff
@@ -58,23 +61,40 @@ fun PayPalWebView(
...
+            Step4_CompleteOrder(uiState, viewModel)
         }
         Spacer(modifier = Modifier.size(contentPadding))
     }
 }
 
 @Composable
-private fun Step1_CreateOrder(uiState: PayPalWebUiState, viewModel: PayPalWebViewModel) {
+fun Step1_WebOrNativeCheckoutHeader(viewModel: PayPalCheckoutViewModel) {
+    val uiState: PayPalUiState by viewModel.uiState.collectAsStateWithLifecycle()
```
sshropshire: I think this PR has some old changes in it. We just updated this UI code in the last PR that was merged.

CorePayments/src/main/java/com/paypal/android/corepayments/api/AuthenticationSecureTokenServiceAPI.kt
kgangineni: Updated base branch, these were all old changes

CorePayments/src/main/java/com/paypal/android/corepayments/common/DeviceInspector.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import android.content.Context
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class DeviceInspector(private val context: Context) {
+
+    fun isPayPalInstalled(): Boolean {
+        return isAppInstalled(PAYPAL_APP_PACKAGE)
+    }
```
sshropshire:
```suggestion
    val isPayPalInstalled: Boolean
      get() = isAppInstalled(PAYPAL_APP_PACKAGE)
```
nit: it seems like this could be a read-only property.

sshropshire (approved):
🌊

KunJeongPark (approved):

PR #335, opened 2025-07-14
Button Label Updates

sshropshire:
### Summary of changes

* PaymentButtons
  * Add `PayPalButtonLabel.NONE` enum case
* Breaking Changes
  * PaymentButtons
    * Remove `PayPalButtonLabel.PAY_LATER` enum case
    * Remove `PayPalButtonLabel.PAYPAL` enum case

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   * Add `PaymentButtonEdges` sealed class
   * Add `PaymentButton.edges` property
   * Add `PaymentButton:payment_button_edges` xml attribute.
+  * Add `PayPalButtonLabel.NONE` enum case
 * Breaking Changes
   * PaymentButtons
     * Remove `PaymentButtonSize` enum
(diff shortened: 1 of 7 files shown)


KunJeongPark:
@sshropshire Hey Steven, this is just an issue on Android, I think.
In iOS, we fixed the labels to button specific types in the init.
I'm just making sure there's not something equivalent in iOS to fix.
Yes, I think we are good in iOS. we have PaymentButtonLabel enum but specific buttons
are initialized with a specific label enum.

I think I do need to add "Credit" label to the PayPalCreditButton. I will revisit the button design to make sure.

KunJeongPark (approved):

sshropshire:
> this is just an issue on Android, I think. In iOS, we fixed the labels to button specific types in the init. I'm just making sure there's not something equivalent in iOS to fix.

@KunJeongPark that makes sense. We definitely were missing a `NONE` type for the button. It seems like iOS already has a `.none` type in place.

PR #347, opened 2025-09-03
integrate kotlinx-serialization in CardPayments

kgangineni:
### Summary of changes
- Add required dependency for kotlinx serialization
- serialize request body for`/confirm-payment-source` with kotlinx serialization
- 
 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/CardPayments/proguard-rules.pro
+++ b/CardPayments/proguard-rules.pro
@@ -19,3 +19,27 @@
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
+
+# Kotlin serialization rules
+-keepattributes *Annotation*, InnerClasses
+-dontnote kotlinx.serialization.AnnotationsKt # core serialization annotations
+
+# kotlinx-serialization-json specific. Add this if you have problems with serialization.
+-keepclassmembers class kotlinx.serialization.json.** {
+    *** Companion;
+}
+-keepclasseswithmembers class kotlinx.serialization.json.** {
+    kotlinx.serialization.KSerializer serializer(...);
+}
+
+# Change here com.yourcompany.yourpackage
+-keep,includedescriptorclasses class com.paypal.android.cardpayments.**$$serializer { *; }
+-keepclassmembers class com.paypal.android.cardpayments.** {
+    *** Companion;
+}
+-keepclasseswithmembers class com.paypal.android.cardpayments.** {
+    kotlinx.serialization.KSerializer serializer(...);
+}
+
+# Keep @Serializable classes
+-keep @kotlinx.serialization.Serializable class * { *; }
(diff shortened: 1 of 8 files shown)


CardPayments/proguard-rules.pro
```diff
@@ -18,4 +18,28 @@
 
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
-#-renamesourcefileattribute SourceFile
\ No newline at end of file
+#-renamesourcefileattribute SourceFile
+
+# Kotlin serialization rules
```
sshropshire: Can we make proguard changes in a separate epic? Focusing on `kotlinx.serialization` will simplify these PRs a ton.

CardPayments/src/main/java/com/paypal/android/cardpayments/CardRequestFactory.kt
```diff
@@ -1,54 +1,75 @@
...
 import com.paypal.android.corepayments.APIRequest
 import com.paypal.android.corepayments.HttpMethod
-import org.json.JSONObject
+import kotlinx.serialization.InternalSerializationApi
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
 
 internal class CardRequestFactory {
 
+    private val json = Json
```
sshropshire:
```suggestion
```
Can we remove this line?

sshropshire (approved):

CardPayments/src/main/java/com/paypal/android/cardpayments/CardRequestFactory.kt
```diff
@@ -1,54 +1,76 @@
...
 import com.paypal.android.corepayments.APIRequest
 import com.paypal.android.corepayments.HttpMethod
-import org.json.JSONObject
+import kotlinx.serialization.InternalSerializationApi
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
 
-internal class CardRequestFactory {
+internal class CardRequestFactory(
+    private val json: Json = Json
```
sshropshire: I'm still on the fence about this one. I guess I'm not sure what benefit it provides?

kgangineni: meaning injecting instead of just using `Json` that library provides, the idea is this can be configured, right now, there are no configurations, just using all defaults, but if we want to configure and use it as common, then the same object can be injected

sshropshire: My only concern is it may be confusing to someone new to the codebase. I'd honestly prefer to follow the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle now and remove it since we don't have an immediate need to configure JSON parsing.

KunJeongPark (approved):

PR #348, opened 2025-09-05
Migrate PaymentsJSON to kotlin serialization

kgangineni:
### Summary of changes

- Remove PaymentsJSON.kt
- Migrate to typed data classes for response
- simplify CardResponseParser.kt

 ### Checklist

 - [ ] Added a changelog entry

-


CardPayments/src/main/java/com/paypal/android/cardpayments/model/PurchaseUnit.kt
```diff
@@ -1,19 +1,20 @@
...
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
 
+@InternalSerializationApi
 /**
  * A purchase unit for an order. If an order takes multiple purchase units,
  * each one must contain a reference id.
  */
+@Serializable
 internal data class PurchaseUnit(
```
KunJeongPark, line 13: I know this is just refactoring, but I just noticed that we don't use this field in iOS
We have it in demo app, but not in the SDK code.
I think maybe it's remnant of get order we used to have?

KunJeongPark (approved):

CardPayments/src/main/java/com/paypal/android/cardpayments/CardResponseParser.kt
```diff
@@ -1,79 +1,80 @@
...
+                ConfirmPaymentSourceResult.Failure(APIClientError.noResponseData(correlationId))
+            }
 
-                        val description = "$message -> $errorDetails"
-                        APIClientError.httpURLConnectionError(status, description, correlationId)
-                    }
+            else -> {
+                try {
+                    val response =
+                        json.decodeFromString<ConfirmPaymentSourceResponse>(httpResponse.body!!)
```
sshropshire: We do call `httpResponse.body.isNullOrBlank()` in multiple places. Is there a way we can avoid having to force unwrap here?

PR #352, opened 2025-09-12
Migrate Demo app to Kotlin Serialization

kgangineni:
### Summary of changes
  - Add kotlinx.serialization plugin and dependencies
  - Create strongly-typed serializable models for API requests and responses
  - Implement KotlinSerializationConverterFactory for Retrofit
  - Update API services to use typed request/response models
  - Replace manual JSON parsing with automatic deserialization
  - Update all use cases to use the new serialization models 

 ### Checklist

 - [ ] Added a changelog entry

-


Demo/src/main/java/com/paypal/android/api/model/CardPaymentToken.kt
```diff
@@ -1,8 +1,13 @@
...
 
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
 data class CardPaymentToken(
     val id: String,
     val customerId: String,
+    @SerialName("last_digits")
     val cardLast4: String,
```
sshropshire:
```suggestion
    val lastDigits: String,
```
I think it's worth renaming this property to `lastDigits` as well for consistency.

Demo/src/main/java/com/paypal/android/api/model/Order.kt
```diff
@@ -2,12 +2,16 @@ package com.paypal.android.api.model
...
+import kotlinx.serialization.Serializable
 
 @Parcelize
+@Serializable
 data class Order(
     val id: String? = null,
     val intent: String? = null,
     val status: String? = null,
+    @SerialName("last_digits")
     val cardLast4: String? = null,
```
sshropshire:
```suggestion
    val lastDigits: String? = null,
```
Same here, totally worth cleaning up to align with the actual name of the underlying resource. I think `cardLast4` as a naming convention is historically a Braintree thing.

sshropshire (approved):
🌊

PR #355, opened 2025-09-18
Release workflows

kgangineni:
### Summary of changes
  - Modified release.yml to trigger automatically on merge to main branch
  - Removed manual workflow dispatch trigger and version input parameter
  - Simplified workflow to always use the version from codebase
  - Adds workflow to create release PR
  - Example release [PR](https://github.com/paypal/paypal-android/pull/356)

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/.github/workflows/create_release_pr.yml
+++ b/.github/workflows/create_release_pr.yml
@@ -42,7 +42,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           commit-message: "Prepare release ${{ github.event.inputs.versionName }}"
           branch: release/${{ github.event.inputs.versionName }}
-          delete-branch: false
+          delete-branch: true
           title: "Release ${{ github.event.inputs.versionName }}"
           body: |
             ## Release PR for version ${{ github.event.inputs.versionName }}
--- a/.github/workflows/create_release_pr.yml
+++ b/.github/workflows/create_release_pr.yml
@@ -1,0 +1,57 @@
+name: Create Release PR
+
+on:
+  workflow_dispatch:
+    inputs:
+      versionName:
+        description: 'Version to release (e.g., 2.1.0)'
+        required: true
+        type: string
+
+jobs:
+  create_release_pr:
+    name: Create Release PR
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Full git history to ensure branch creation works
+
+      - name: Set up Java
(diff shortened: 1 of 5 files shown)


.github/workflows/create_release_pr.yml
```diff
@@ -0,0 +1,57 @@
...
+          DATE=$(date +%Y-%m-%d)
+          ./gradlew -PversionParam=${{ github.event.inputs.versionName }} updateCHANGELOGVersion
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v5
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: "Prepare release ${{ github.event.inputs.versionName }}"
+          branch: release/${{ github.event.inputs.versionName }}
+          delete-branch: false
```
sshropshire:
```suggestion
          delete-branch: true
```
Would it make sense to delete the branch afterward? We may end up with a bunch of dangling refs. We do subsequently tag the branch before release too, so we shouldn't need the branch after it's been merged.

.github/workflows/release.yml
```diff
@@ -149,30 +165,26 @@ jobs:
...
+          VERSION="${{ needs.extract_version.outputs.version }}"
+          echo "Using current version from build.gradle: $VERSION"
+          
+          # CHANGELOG is already updated by the release PR process
+          # Just create the tag without modifying files
+          git tag $VERSION -a -m "Release $VERSION, automated from main branch"
+          git push origin $VERSION
       - name: Save changelog entries to a file
-        run: sed -e '1,/##/d' -e '/##/,$d' CHANGELOG.md > changelog_entries.md
+        run: sed -e '1,/##/d'e '/##/,$d' CHANGELOG.md > changelog_entries.md
```
sshropshire: Is the `-e` supposed to be removed here?

.github/workflows/release.yml
```diff
@@ -149,30 +165,26 @@ jobs:
...
-          git add .
-          git commit -m 'Next version ready'
-          git push origin HEAD ${{ github.event.inputs.version }}
+          # Use extracted version from build.gradle
+          VERSION="${{ needs.extract_version.outputs.version }}"
+          echo "Using current version from build.gradle: $VERSION"
+          
+          # CHANGELOG is already updated by the release PR process
+          # Just create the tag without modifying files
+          git tag $VERSION -a -m "Release $VERSION, automated from main branch"
```
sshropshire: Can we add back the `${{ github.actor }}` part? I feel like that's good information for the tag.

sshropshire (approved):

PR #392, opened 2025-10-17
Update Client Config

kgangineni:
### Summary of changes
 - Make GraphQL call `graphql?UpdateClientConfig` to update client config for web checkout

 ### Checklist

 - [x] Added a changelog entry

- @kgangineni


PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -63,6 +72,15 @@ class PayPalWebCheckoutClient internal constructor(
         checkoutOrderId = request.orderId
         analytics.notify(CheckoutEvent.STARTED, checkoutOrderId)
 
+        // Call updateCCO for all funding sources (fire and forget)
+        CoroutineScope(ioDispatcher).launch {
+            updateCCO(request)
```
sshropshire: We are supposed to await here [similar to iOS](https://github.com/paypal/paypal-ios/blob/64093dc761a42140b8aeabd82626e66632bc1c8d/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift#L54).

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* PayPalWebPayments
+  * updates client config in web checkout
```
sshropshire: Do we need to add this to the `CHANGELOG`? CCO isn't part of the SDK's API so merchants probably won't know what this means.

CorePayments/src/main/java/com/paypal/android/corepayments/UpdateClientConfigAPI.kt
```diff
@@ -0,0 +1,119 @@
...
+                    )
+                when (graphQLResponse) {
+                    is GraphQLResult.Success -> {
+                        val correlationId = graphQLResponse.correlationId
+                        graphQLResponse.response.data?.let {
+                            UpdateClientConfigResult.Success
+                        } ?: UpdateClientConfigResult.Failure(
+                            APIClientError.noResponseData(
+                                correlationId
+                            )
```
sshropshire:
```suggestion
                            APIClientError.noResponseData(correlationId)
```
Does this break the line limit? Single line would be more readable here.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -56,13 +70,15 @@ class PayPalWebCheckoutClient internal constructor(
      *
      * @param request [PayPalWebCheckoutRequest] for requesting an order approval
      */
-    fun start(
+    suspend fun start(
```
sshropshire:
```suggestion
    private suspend fun start(
```
It makes more sense to fully support coroutines in the next major version.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -18,9 +24,14 @@ import com.paypal.android.paypalwebpayments.analytics.VaultEvent
 class PayPalWebCheckoutClient internal constructor(
     private val analytics: PayPalWebAnalytics,
     private val payPalWebLauncher: PayPalWebLauncher,
-    private val sessionStore: PayPalWebCheckoutSessionStore = PayPalWebCheckoutSessionStore()
+    private val sessionStore: PayPalWebCheckoutSessionStore,
+    private val updateClientConfigAPI: UpdateClientConfigAPI,
+    private val ioDispatcher: CoroutineDispatcher
 ) {
 
+    // application scope to execute fire-and-forget tasks
```
sshropshire:
```suggestion
    // application scope to execute CCO updates
```
Since we're no longer using fire and forget.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -81,6 +97,24 @@ class PayPalWebCheckoutClient internal constructor(
...
+     * @param callback [PayPalWebStartCallback] to receive the result
+     */
+    fun start(
+        activity: Activity,
+        request: PayPalWebCheckoutRequest,
+        callback: PayPalWebStartCallback
+    ) {
+        applicationScope.launch {
+            val result = start(activity, request)
+            callback.onPayPalWebStartResult(result)
```
sshropshire: q: should we callback on the main thread?

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
sshropshire: Actually this is a breaking change. This method should be deprecated instead of converted to a `suspend` function.

sshropshire: I'm surprised the CI tool didn't pick this up.

CHANGELOG.md
kgangineni: Minor security improvements?

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
kgangineni: it did flagged updated API. Thought we will go with major
deprecated this and vault functions

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
kgangineni: good catch yes, even launching browser switch should happen in main thread
ioDispatcher is taken care in Http class

CHANGELOG.md
sshropshire: I would add in the new `start()` with callback method and mention that the other method has been deprecated.

sshropshire (approved):

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
sshropshire: I think these are still calling back on the background thread. I think we have to wrap the callback invocation with `withContext (Dispatchers.Main) {}`.

PR #409, opened 2026-01-16
Auth Tab, Close listener functionality 

kgangineni:
### Summary of changes
 - Introduces auth tab functionality
 - introduces new overloaded function 
 kotlin```fun start(
    activity: Activity,
    request: PayPalWebCheckoutRequest,
    activityResultLauncher: ActivityResultLauncher<Intent>,
    callback: PayPalWebStartCallback
) {
}``` 
to pass activityResultLauncher parameter
 - Auth tab is allowed on devices with chrome version 137+ (may be hard to create in emulator but able to do in pixel devices)
 - API works including callback in ActivityLauncher on all chrome versions, it might not have additional security provided auth tab in lower versions of chrome

 ### Checklist

 - [ ] Added a changelog entry

-


CardPayments/api/CardPayments.api
```diff
@@ -19,7 +19,7 @@ public final class com/paypal/android/cardpayments/Card : android/os/Parcelable
 	public final fun component6 ()Lcom/paypal/android/corepayments/Address;
 	public final fun copy (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Lcom/paypal/android/corepayments/Address;)Lcom/paypal/android/cardpayments/Card;
 	public static synthetic fun copy$default (Lcom/paypal/android/cardpayments/Card;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Lcom/paypal/android/corepayments/Address;ILjava/lang/Object;)Lcom/paypal/android/cardpayments/Card;
-	public fun describeContents ()I
+	public final fun describeContents ()I
```
sshropshire, line 22: In a strict sense, this is technically a breaking change. Is there a way to prevent this change?

CorePayments/src/main/java/com/paypal/android/corepayments/browserswitch/AuthTabClient.kt
```diff
@@ -0,0 +1,43 @@
...
+            returnUrlScheme != null -> {
+                authTabIntent.launch(
+                    activityResultLauncher,
+                    options.launchUri,
+                    returnUrlScheme
+                )
+            }
+
+            else -> {
+                throw IllegalArgumentException("Either appLinkUrl or returnUrlScheme must not be null")
```
sshropshire: Code smell I noticed: This error message will be the same when `appLinkUri?.path == null`. To me it would be better if `host` and `path` are required parameters to the `launchAuthTab` method.

Demo/build.gradle
```diff
@@ -40,7 +41,7 @@ android {
 
     composeOptions {
         // Ref: https://developer.android.com/jetpack/androidx/releases/compose-kotlin#pre-release_kotlin_compatibility
-        kotlinCompilerExtensionVersion "1.5.14"
+        kotlinCompilerExtensionVersion "1.5.15"
```
sshropshire, line 44: This is probably what caused the breaking changes.

sshropshire: Not sure though actually, pure speculation.

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalCheckoutView.kt
```diff
@@ -46,11 +49,17 @@ fun PayPalCheckoutView(
     val context = LocalContext.current
     OnLifecycleOwnerResumeEffect {
         val intent = context.getActivityOrNull()?.intent
-        intent?.let { viewModel.completeAuthChallenge(it) }
+        if (!uiState.useAuthTabLauncher) {
```
sshropshire: In a fallback scenario where AuthTab is unavailable and a CCT is launch, would `useAuthTabLauncher == true` prevent the deep link from being captured?

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalCheckoutView.kt
```diff
@@ -107,10 +116,29 @@ private fun Step1_CreateOrder(uiState: PayPalUiState, viewModel: PayPalCheckoutV
 @Composable
 private fun Step2_StartPayPalCheckout(uiState: PayPalUiState, viewModel: PayPalCheckoutViewModel) {
     val context = LocalContext.current
+
+    val activityResultLauncher = rememberLauncherForActivityResult(
```
sshropshire: This should probably be [hoisted](https://developer.android.com/develop/ui/compose/state-hoisting) to the top level. The activity result launcher is only registered in Step2. In a cold start scenario, the deep link won't be captured.

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalCheckoutViewModel.kt
```diff
@@ -103,34 +110,54 @@ class PayPalCheckoutViewModel @Inject constructor(
...
             APP_CUSTOM_URL_SCHEME
         )
 
-        paypalClient.start(activity, checkoutRequest) { startResult ->
-            when (startResult) {
-                is PayPalPresentAuthChallengeResult.Success -> {
-                    // do nothing; wait for user to authenticate PayPal checkout in Chrome Custom Tab
-                }
+        if (useAuthTabLauncher && activityResultLauncher != null) {
+            paypalClient.start(activity, checkoutRequest, activityResultLauncher) { startResult ->
```
sshropshire, line 144: I still feel like we don't need to pass the activity result launcher to `PayPalClient`. It would clean up the design a lot.

build.gradle
```diff
@@ -16,7 +16,7 @@ buildscript {
             "demoAppVersionCode"  : 4,
             "androidMinSdkVersion": 23,
             "androidTargetVersion": 35,
```
sshropshire:
```suggestion
            "androidTargetVersion": 36,
```

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalCheckoutViewModel.kt
kgangineni: Ya, we can definitely achieve it with composable module that's in draft PR, but need to check if we can avoid in non composable scenarios, will try to do on POC next

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalCheckoutView.kt
kgangineni: fixed by adding check for it, good catch

Demo/build.gradle
kgangineni: no I think this is because of kotlin version update, this is only for demo app, we don't do api check for demo app

PR #415, opened 2026-02-10
Fix issue of null in return url when fallback custom url scheme is used

kgangineni:
### Summary of changes
 - When appLinkUrl and deprecated param for urlScheme are not passed, then instead of constructing url using fallbackUrlScheme, null is passed in return url, this PR fixes issue

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalCheckoutViewModel.kt
+++ b/Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalCheckoutViewModel.kt
@@ -61,7 +61,7 @@ class PayPalCheckoutViewModel @Inject constructor(
             _uiState.update { it.copy(appSwitchWhenEligible = value) }
         }
 
-    var returnToAppStrategy: ReturnToAppStrategyOption
+    var returnToAppStrategyOption: ReturnToAppStrategyOption
         get() = _uiState.value.returnToAppStrategyOption
         set(value) {
             _uiState.update { it.copy(returnToAppStrategyOption = value) }
@@ -125,7 +125,7 @@ class PayPalCheckoutViewModel @Inject constructor(
             orderId,
             fundingSource,
             appSwitchWhenEligible,
-            returnToAppStrategy.toReturnToAppStrategy()
+            returnToAppStrategyOption.toReturnToAppStrategy()
         )
 
         paypalClient.start(activity, checkoutRequest) { startResult ->
--- a/Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalCheckoutViewModel.kt
+++ b/Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalCheckoutViewModel.kt
@@ -6,8 +6,6 @@ import android.util.Log
 import androidx.activity.ComponentActivity
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
-import com.paypal.android.DemoConstants.APP_CUSTOM_URL_SCHEME
-import com.paypal.android.DemoConstants.APP_URL
 import com.paypal.android.api.model.Order
 import com.paypal.android.api.model.OrderIntent
 import com.paypal.android.api.services.SDKSampleServerAPI
@@ -20,7 +18,7 @@ import com.paypal.android.paypalwebpayments.PayPalWebCheckoutClient
 import com.paypal.android.paypalwebpayments.PayPalWebCheckoutFinishStartResult
 import com.paypal.android.paypalwebpayments.PayPalWebCheckoutFundingSource
 import com.paypal.android.paypalwebpayments.PayPalWebCheckoutRequest
-import com.paypal.android.uishared.enums.DeepLinkStrategy
+import com.paypal.android.uishared.enums.ReturnToAppStrategyOption
 import com.paypal.android.uishared.state.ActionState
 import com.paypal.android.usecase.CompleteOrderUseCase
 import com.paypal.android.usecase.CreateOrderUseCase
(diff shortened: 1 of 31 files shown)


PayPalWebPayments/src/test/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClientUnitTest.kt
```diff
@@ -1930,4 +1930,247 @@ class PayPalWebCheckoutClientUnitTest {
...
+            appLinkUrl = testAppLinkUrl
+        )
+        sut.startAsync(activity, request)
+
+        verify(exactly = 1) {
+            payPalWebLauncher.launchWithUrl(
+                activity = activity,
+                uri = match { uri ->
+                    uri.getQueryParameter("redirect_uri") == testAppLinkUrl
+                },
```
sshropshire: q: Is this something merchants will be required to do? I'm not super familiar with what `match` does.

kgangineni: `match` is just inline argument capture, in this case, this test only verifies when `appLinkUrl` is provided then it is sent as redirect url

sshropshire: Oh that's the DSL of the test library. Got it.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -470,8 +471,13 @@ class PayPalWebCheckoutClient internal constructor(
         return result
     }
 
-    private val redirectUriPayPalCheckout
-        get() = urlScheme?.let { "$urlScheme://x-callback-url/paypal-sdk/paypal-checkout" }
+    private fun redirectUrlWithCustomScheme(scheme: String?): String? {
```
sshropshire: I'd throw a comment here explaining the bugfix. Hopefully in the next major version we can find a better API that keeps this from being possible.

sshropshire:
Hey @kgangineni can you share the steps to reproduce the original bug?

sshropshire (approved):

kgangineni:
> Hey @kgangineni can you share the steps to reproduce the original bug?

1. checkout develop branch
2. make code change to send appLinkUrl only when DeepLinkStrategy is APP_LINKS send null otherwise
3. select deep link strategy as CUSTOM_URL_SCHEME
4. notice app doesn't return properly after checkout

sshropshire:
@kgangineni oh we might need to make this reproducible from UI. Otherwise it'll be difficult to catch regressions.

kgangineni:
> @kgangineni oh we might need to make this reproducible from UI. Otherwise it'll be difficult to catch regressions.

in this PR I made required UI changes, but since it also has fix you will not be able to reproduce

Demo/src/main/java/com/paypal/android/ui/paypalweb/PayPalCheckoutViewModel.kt
```diff
@@ -62,10 +61,10 @@ class PayPalCheckoutViewModel @Inject constructor(
             _uiState.update { it.copy(appSwitchWhenEligible = value) }
         }
 
-    var deepLinkStrategy
-        get() = _uiState.value.deepLinkStrategy
+    var returnToAppStrategy: ReturnToAppStrategyOption
```
sshropshire:
```suggestion
    var returnToAppStrategyOption: ReturnToAppStrategyOption
```

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -562,4 +570,31 @@ class PayPalWebCheckoutClient internal constructor(
...
+    ): ReturnToAppStrategy? {
+        return requestStrategy ?: urlScheme?.let {
+            ReturnToAppStrategy.CustomUrlScheme(urlScheme)
+        }
+    }
+
+    private val noReturnToAppStrategyError = PayPalSDKError(
+        code = PayPalSDKErrorCode.CHECKOUT_ERROR.ordinal,
+        errorDescription = "ReturnToAppStrategy or urlScheme is required. "
+    )
```
sshropshire: We should move this to [PayPalWebCheckoutError.kt](https://github.com/paypal/paypal-android/blob/develop/PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/errors/PayPalWebCheckoutError.kt).

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -562,4 +570,31 @@ class PayPalWebCheckoutClient internal constructor(
...
+
+    /**
+     * builds return URL from return to app strategy.
+     */
+    private val ReturnToAppStrategy.returnUrl: String
+        get() = when (this) {
+            is ReturnToAppStrategy.AppLink -> this.appLinkUrl
+            is ReturnToAppStrategy.CustomUrlScheme ->
+                "${this.urlScheme}://x-callback-url/paypal-sdk/paypal-checkout"
+        }
```
sshropshire: Any reason this can't be in [ReturnToAppStrategy.kt](https://github.com/paypal/paypal-android/blob/fixCustomUrlScheme/CorePayments/src/main/java/com/paypal/android/corepayments/ReturnToAppStrategy.kt) ?

kgangineni: it can be moved there

sshropshire (approved):

PR #429, opened 2026-03-03
Add PayPal Android SDK integration guide

naolson-paypal:
## Summary
- Adds a comprehensive integration guide draft covering card payments, PayPal web checkout, card vaulting, PayPal vaulting, deep link configuration, and process death recovery
- Includes project setup instructions, SDK initialization, and go-live checklist
- Located at `Docs/IntegrationGuide/draft.md`

## Test plan
- [ ] Review guide content for technical accuracy
- [ ] Verify code samples compile against current SDK APIs
- [ ] Confirm deep link configuration instructions match current implementation

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Docs/IntegrationGuide.md
+++ b/Docs/IntegrationGuide.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-This guide shows you how to accept card payments and PayPal payments in your Android app using the PayPal Android SDK. You learn how to set up the SDK, process one-time payments, and vault payment methods for future use. Before you begin, review the Prerequisites section to make sure your environment is ready.
+This guide shows you how to accept card payments and PayPal payments in your Android app using the PayPal Android SDK. You learn how to set up the SDK, process one-time payments, and vault payment methods for future use. Before you begin, review the [Prerequisites](#prerequisites) section to make sure your environment is ready.
 
 **PayPal checkout flow:**
 
@@ -122,11 +122,47 @@ implementation 'com.paypal.android:card-payments:2.3.0-SNAPSHOT'
 
 ### Configure deep links
 
-The SDK opens a Chrome Custom Tab or the PayPal app for certain flows (PayPal web checkout, PayPal vaulting, 3D Secure card authentication). When the customer finishes, PayPal redirects them back to your app using a deep link. Configure your app to receive this redirect.
+The SDK opens a Chrome Custom Tab or the PayPal app for certain flows (PayPal web checkout, PayPal vaulting, 3D Secure card authentication). When the customer finishes, PayPal redirects them back to your app using a deep link. You'll need to configure your app to receive this redirect.
 
-The SDK supports two return URL strategies. Choose one or configure both as a fallback:
+The SDK uses `ReturnToAppStrategy` to determine how the customer returns to your app. Choose one of the two strategies:
 
-**Option 1: Custom URL Scheme**
+**Option 1: Android App Links (HTTPS) — Recommended**
+
+App Links use HTTPS URLs with domain verification (`android:autoVerify="true"`). This requires you to host a `/.well-known/assetlinks.json` file on your domain.
+
+```xml
+<activity
+    android:name=".MainActivity"
+    android:exported="true"
+    android:launchMode="singleTop">
+
+    <!-- Standard launcher intent-filter -->
+    <intent-filter>
+        <action android:name="android.intent.action.MAIN" />
+        <category android:name="android.intent.category.LAUNCHER" />
+    </intent-filter>
+
+    <intent-filter android:autoVerify="true">
+        <action android:name="android.intent.action.VIEW" />
+        <category android:name="android.intent.category.DEFAULT" />
+        <category android:name="android.intent.category.BROWSABLE" />
+        <data
+            android:host="your-app-domain.example.com"
+            android:pathPrefix="/"
+            android:scheme="https" />
+    </intent-filter>
+</activity>
+```
+
+Use this strategy in SDK requests:
+
+```kotlin
+val returnToAppStrategy = ReturnToAppStrategy.AppLink(
+    appLinkUrl = "https://your-app-domain.example.com"
+)
+```
+
+**Option 2: Custom URL Scheme**
 
 Register a custom URL scheme in your `AndroidManifest.xml`. Your activity needs to use `android:launchMode="singleTop"` so that the OS delivers the return intent to the existing activity instance rather than creating a new one.
 
@@ -152,25 +188,13 @@ Register a custom URL scheme in your `AndroidManifest.xml`. Your activity needs 
 </activity>
 ```
 
-Use this scheme as your `fallbackUrlScheme` in SDK requests: `"com.example.myapp"`.
+Use this strategy in SDK requests:
 
-**Option 2: Android App Links (HTTPS)**
-
-App Links use HTTPS URLs with domain verification (`android:autoVerify="true"`). This requires you to host a `/.well-known/assetlinks.json` file on your domain.
-
-```xml
-<intent-filter android:autoVerify="true">
-    <action android:name="android.intent.action.VIEW" />
-    <category android:name="android.intent.category.DEFAULT" />
-    <category android:name="android.intent.category.BROWSABLE" />
-    <data
-        android:host="your-app-domain.example.com"
-        android:pathPrefix="/"
-        android:scheme="https" />
-</intent-filter>
+```kotlin
+val returnToAppStrategy = ReturnToAppStrategy.CustomUrlScheme(
+    urlScheme = "com.example.myapp"
+)
 ```
-
-Use your HTTPS URL as the `appLinkUrl` in SDK requests: `"https://your-app-domain.example.com"`.
 
 > **Architecture warning:** Only the activity that declares these intent-filters can receive deep links back into the host application. If your app has multiple activities, route all SDK flows through the activity that declares the deep link intent-filters.
 
@@ -178,7 +202,7 @@ Use your HTTPS URL as the `appLinkUrl` in SDK requests: `"https://your-app-domai
 
 ## Initialize the SDK
 
-Create a `CoreConfig` object with your client ID and environment. Then create a client for each payment method you need.
+Create a `CoreConfig` object with your client ID and environment. Then create a client for each payment method you need. The Demo app creates clients inside ViewModel constructors so they are scoped to the screen lifecycle.
 
 ```kotlin
 // Create configuration (use Environment.LIVE for production)
@@ -239,7 +263,7 @@ The same `instanceState` / `restore()` pattern applies to `PayPalWebCheckoutClie
 
 The merchant-facing [Demo app](Demo/) included in the repository uses Jetpack Compose Navigation with a single entry-point activity. All SDK flows live within `MainActivity`, which contains a `NavHost`. Individual payment screens are composable destinations. Deep link returns arrive at `MainActivity` and are dispatched to the active screen.
 
-This single-entry-point pattern avoids routing ambiguity: the OS delivers deep link intents to `MainActivity`, which is the only activity that declares the deep link intent-filters.
+This single-entry-point pattern is recommended because it avoids routing ambiguity: the OS delivers deep link intents to `MainActivity`, which is the only activity that declares the intent-filters for deep linking.
 
 > **Note:** Your app's architecture may differ from the Demo app. The key requirement is that deep link intent-filters are declared on the activity that handles SDK return flows. The patterns shown in this guide are not prescriptive — adapt them to fit your app's architecture.
 
@@ -337,8 +361,9 @@ val card = Card(
 )
 
 // Build the request
-// returnUrl must match a URL scheme registered in your AndroidManifest.xml
-val returnUrl = "com.example.myapp://"    // or your HTTPS app link URL
+// returnUrl must match a deep link registered in your AndroidManifest.xml
+// Use your App Link URL or custom URL scheme (e.g., "com.example.myapp://")
+val returnUrl = "https://your-app-domain.example.com"
 val cardRequest = CardRequest(
     orderId = orderId,                    // order ID from your server
     card = card,
@@ -443,6 +468,54 @@ class MainActivity : ComponentActivity() {
 }
 ```
 
+**Jetpack Compose alternative**
+
+If your app uses Jetpack Compose, you can handle deep link returns with composable effects instead of Activity overrides. Place these two effects in the composable that manages your payment flow. Both are needed to cover the two ways Android can deliver the return intent.
+
+```kotlin
+@Composable
+fun CardPaymentScreen(cardClient: CardClient) {
+    val context = LocalContext.current
+    val lifecycleOwner = LocalLifecycleOwner.current
+
+    // Handle return when the activity resumes (e.g., returning from Chrome Custom Tab)
+    val lifecycleState by lifecycleOwner.lifecycle.currentStateFlow.collectAsState()
+    LaunchedEffect(lifecycleState) {
+        if (lifecycleState == Lifecycle.State.RESUMED) {
+            val activity = context as? ComponentActivity ?: return@LaunchedEffect
+            activity.intent?.let { intent -> handleFinishApproveOrder(cardClient, intent) }
+        }
+    }
+
+    // Handle return when a new intent is delivered to a singleTop activity
+    DisposableEffect(Unit) {
+        val activity = context as? ComponentActivity ?: return@DisposableEffect onDispose {}
+        val listener = Consumer<Intent> { newIntent ->
+            handleFinishApproveOrder(cardClient, newIntent)
+        }
+        activity.addOnNewIntentListener(listener)
+        onDispose { activity.removeOnNewIntentListener(listener) }
+    }
+
+    // ... your payment UI
+}
+
+private fun handleFinishApproveOrder(cardClient: CardClient, intent: Intent) {
+    cardClient.finishApproveOrder(intent)?.let { result ->
+        when (result) {
+            is CardFinishApproveOrderResult.Success ->
+                captureOrderOnServer(result.orderId)
+            is CardFinishApproveOrderResult.Failure ->
+                showError(result.error.errorDescription)
+            CardFinishApproveOrderResult.Canceled ->
+                resetToIdleState()
+            CardFinishApproveOrderResult.NoResult ->
+                resetToIdleState()
+        }
+    }
+}
+```
+
 > **Understanding `NoResult`:** `NoResult` means the SDK does not have enough information to determine the outcome — for example, if the customer returned to the app without completing authentication, or if the deep link intent was not associated with the current auth session. The payment may or may not have succeeded server-side. When you receive `NoResult`, query the [Orders v2 API](https://developer.paypal.com/docs/api/orders/v2/) from your server to check the current order status before showing an error or retrying.
 
 ### What's next
@@ -475,8 +548,8 @@ val card = Card(
     securityCode = "123"
 )
 
-// returnUrl must match a URL scheme registered in your AndroidManifest.xml
-val returnUrl = "com.example.myapp://"
+// returnUrl must match a deep link registered in your AndroidManifest.xml
+val returnUrl = "https://your-app-domain.example.com"
 val cardVaultRequest = CardVaultRequest(
     setupTokenId = setupTokenId,   // from your server
     card = card,
@@ -519,7 +592,7 @@ fun createPaymentToken(setupTokenId: String) {
 
 ### Deep link return
 
-Handle the deep link return using the same `onResume()` / `onNewIntent()` pattern used in the card payment flow. Call `finishVault()` instead of `finishApproveOrder()`.
+Handle the deep link return using the same `onResume()` / `onNewIntent()` pattern used in the card payment flow (or the Jetpack Compose alternative). Call `finishVault()` instead of `finishApproveOrder()`.
 
 ```kotlin
 private fun handleVaultReturnIntent(intent: Intent) {
@@ -590,8 +663,7 @@ val checkoutRequest = PayPalWebCheckoutRequest(
     orderId = orderId,                                           // from your server
     fundingSource = PayPalWebCheckoutFundingSource.PAYPAL,       // PAYPAL, PAYPAL_CREDIT, or PAY_LATER
     appSwitchWhenEligible = true,                                // switch to PayPal app if installed
-    appLinkUrl = "https://your-app-domain.example.com",          // HTTPS return URL (or null)
-    fallbackUrlScheme = "com.example.myapp"                      // custom scheme fallback (or null)
+    returnToAppStrategy = returnToAppStrategy                    // from Configure deep links section
 )
 
 paypalClient.start(activity, checkoutRequest) { startResult ->
@@ -632,7 +704,7 @@ fun capturePayPalOrder(orderId: String?) {
 
 ### Deep link return
 
-Override `onResume()` and `onNewIntent()` in your activity to handle the return. Call `finishStart()` to complete the PayPal checkout flow.
+Override `onResume()` and `onNewIntent()` in your activity to handle the return (or use the Jetpack Compose alternative shown in the card payment section). Call `finishStart()` to complete the PayPal checkout flow.
 
 ```kotlin
 private fun handlePayPalReturnIntent(intent: Intent) {
@@ -684,8 +756,7 @@ Your server creates a PayPal setup token and returns the setup token ID.
 val vaultRequest = PayPalWebVaultRequest(
     setupTokenId = setupTokenId,                                 // from your server
     appSwitchWhenEligible = true,                                // switch to PayPal app if installed
-    appLinkUrl = "https://your-app-domain.example.com",          // HTTPS return URL (or null)
-    fallbackUrlScheme = "com.example.myapp"                      // custom scheme fallback (or null)
+    returnToAppStrategy = returnToAppStrategy                    // from Configure deep links section
 )
 
 paypalClient.vault(activity, vaultRequest) { result ->
--- a/Docs/IntegrationGuide.md
+++ b/Docs/IntegrationGuide.md
@@ -4,7 +4,55 @@
 
 This guide shows you how to accept card payments and PayPal payments in your Android app using the PayPal Android SDK. You learn how to set up the SDK, process one-time payments, and vault payment methods for future use. Before you begin, review the Prerequisites section to make sure your environment is ready.
 
-<PlaceholderImage description="End-to-end flow diagram showing the three actors: your Android app (SDK), your server, and PayPal APIs. Arrows show: app collects payment details → app calls SDK → SDK contacts PayPal → PayPal returns result to SDK → SDK returns result to app → app calls your server to capture/authorize." />
+**PayPal checkout flow:**
+
+```mermaid
+sequenceDiagram
+    participant App as Your Android App
+    participant SDK as PayPal SDK
+    participant Browser as Chrome Custom Tab / PayPal App
+    participant Server as Your Server
+    participant PayPal as PayPal APIs
+
+    App->>Server: Create order
+    Server->>PayPal: Orders v2 API (create)
+    PayPal-->>Server: orderId
+    Server-->>App: orderId
+    App->>SDK: start(checkoutRequest)
+    SDK->>Browser: Launch checkout URL
+    Note over Browser,PayPal: Customer logs in & approves payment
+    PayPal-->>Browser: Redirect (deep link)
+    Browser-->>App: Return intent
+    App->>SDK: finishStart(intent)
+    SDK-->>App: orderId, payerId
+    App->>Server: Capture/authorize order
+    Server->>PayPal: Orders v2 API (capture)
+    PayPal-->>Server: Confirmation
+    Server-->>App: Order captured
+```
+
+**Card payment flow:**
+
+```mermaid
+sequenceDiagram
+    participant App as Your Android App
+    participant SDK as PayPal SDK
+    participant Server as Your Server
+    participant PayPal as PayPal APIs
+
+    App->>Server: Create order
+    Server->>PayPal: Orders v2 API (create)
+    PayPal-->>Server: orderId
+    Server-->>App: orderId
+    App->>SDK: approveOrder(cardRequest)
+    SDK->>PayPal: Confirm payment source (REST API)
+    PayPal-->>SDK: Approval result
+    SDK-->>App: Success or AuthorizationRequired
+    App->>Server: Capture/authorize order
+    Server->>PayPal: Orders v2 API (capture)
+    PayPal-->>Server: Confirmation
+    Server-->>App: Order captured
+```
 
 ---
 
@@ -23,7 +71,8 @@ Before you integrate the SDK, make sure you have the following:
 - A server-side integration that can create PayPal orders and capture or authorize them (see [Orders v2 API](https://developer.paypal.com/docs/api/orders/v2/))
 - For card vaulting or PayPal vaulting: a server that can create setup tokens and payment tokens (see [PayPal vaulting documentation](https://developer.paypal.com/docs/))
 - Android Studio with a project targeting Android SDK 23 or higher
-- Kotlin or Java project (the SDK supports both languages)
+- Kotlin or Java project (the SDK is written in Kotlin but is fully callable from Java)
+- If you haven't set up your server integration yet, see the [server-side setup guide](https://developer.paypal.com/docs/checkout/advanced/integrate/) before continuing
 
 ---
 
@@ -55,7 +104,7 @@ All modules are published to Maven Central under the `com.paypal.android` group.
 
 **Snapshot builds**
 
-If you'd like to test upcoming features before they ship, use snapshot builds. First, add the snapshot repository to your top-level `build.gradle`:
+To test upcoming features before they ship, use snapshot builds. First, add the snapshot repository to your top-level `build.gradle`:
 
 ```groovy
 repositories {
@@ -96,7 +145,7 @@ Register a custom URL scheme in your `AndroidManifest.xml`. Your activity needs 
     <!-- Custom URL scheme for returning to app -->
     <intent-filter>
         <action android:name="android.intent.action.VIEW"/>
-        <data android:scheme="com.example.your-app-domain"/>
+        <data android:scheme="com.example.myapp"/>
         <category android:name="android.intent.category.DEFAULT"/>
         <category android:name="android.intent.category.BROWSABLE"/>
     </intent-filter>
@@ -139,7 +188,7 @@ val coreConfig = CoreConfig(
 )
 
 // Create the card payments client
-val cardClient = CardClient(applicationContext, coreConfig)
+val cardClient = CardClient(context, coreConfig)
 
 // Create the PayPal web checkout / vault client
 val paypalClient = PayPalWebCheckoutClient(context, coreConfig)
@@ -188,7 +237,7 @@ The same `instanceState` / `restore()` pattern applies to `PayPalWebCheckoutClie
 
 ## Demo app architecture
 
-The [Demo app](Demo/) included in this repository uses Jetpack Compose Navigation with a single entry-point activity. All SDK flows live within `MainActivity`, which contains a `NavHost`. Individual payment screens are composable destinations. Deep link returns arrive at `MainActivity` and are dispatched to the active screen.
+The merchant-facing [Demo app](Demo/) included in the repository uses Jetpack Compose Navigation with a single entry-point activity. All SDK flows live within `MainActivity`, which contains a `NavHost`. Individual payment screens are composable destinations. Deep link returns arrive at `MainActivity` and are dispatched to the active screen.
 
 This single-entry-point pattern avoids routing ambiguity: the OS delivers deep link intents to `MainActivity`, which is the only activity that declares the deep link intent-filters.
 
@@ -213,7 +262,36 @@ class MainActivity : ComponentActivity() {
 
 Use this flow when a customer pays with a credit or debit card. The flow has three steps: your server creates an order, the SDK approves the order with the card, and your server captures or authorizes the order.
 
-<PlaceholderImage description="Sequence diagram for card payment: 1) App calls server to create order → server calls Orders v2 API and returns orderId. 2) App calls CardClient.approveOrder() → SDK contacts PayPal → if 3D Secure required, SDK opens Chrome Custom Tab → customer completes 3D Secure → deep link returns to app → app calls finishApproveOrder(). 3) App calls server to capture/authorize → server calls Orders v2 API." />
+```mermaid
+sequenceDiagram
+    participant App as Your App
+    participant Server as Your Server
+    participant SDK as CardClient
+    participant PayPal as PayPal
+    participant Browser as Chrome Custom Tab
+
+    App->>Server: Create order
+    Server->>PayPal: Orders v2 API
+    PayPal-->>Server: orderId
+    Server-->>App: orderId
+    App->>SDK: approveOrder(cardRequest)
+    SDK->>PayPal: Submit card data
+    alt No 3D Secure required
+        PayPal-->>SDK: Success
+        SDK-->>App: CardApproveOrderResult.Success
+    else 3D Secure required
+        PayPal-->>SDK: AuthorizationRequired
+        SDK-->>App: CardApproveOrderResult.AuthorizationRequired
+        App->>Browser: presentAuthChallenge()
+        Browser->>PayPal: Customer completes 3DS
+        PayPal-->>Browser: Redirect (deep link)
+        Browser-->>App: Return intent
+        App->>SDK: finishApproveOrder(intent)
+        SDK-->>App: CardFinishApproveOrderResult
+    end
+    App->>Server: Capture/authorize order
+    Server->>PayPal: Orders v2 API
+```
 
 ### What you'll do
 
@@ -343,6 +421,8 @@ class MainActivity : ComponentActivity() {
 }
 ```
 
+> **Understanding `NoResult`:** `NoResult` means the SDK does not have enough information to determine the outcome — for example, if the customer returned to the app without completing authentication, or if the deep link intent was not associated with the current auth session. The payment may or may not have succeeded server-side. When you receive `NoResult`, query the [Orders v2 API](https://developer.paypal.com/docs/api/orders/v2/) from your server to check the current order status before showing an error or retrying.
+
 ### What's next
 
 After capturing or authorizing the order on your server, show a confirmation to the customer. The card payment flow is complete.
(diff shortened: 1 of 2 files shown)


Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+
+Then reference the snapshot version:
+
+```groovy
+implementation 'com.paypal.android:card-payments:2.3.0-SNAPSHOT'
+```
+
+### Configure build settings
+
+Do not hardcode your client ID in source code. Store it in a `paypal.properties` file that you exclude from version control, then inject it at build time.
```
kgangineni: This is not necessary, it's upto merchant apps, on how to handle injecting clientId,
it can be build config, a properties file, or any other way 
Same below also, all the properties related info is not needed

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+def loadPropertiesFromFile(filePath) {
+    def result = new Properties()
+    try {
+        result.load(new FileInputStream(rootProject.file(filePath)))
+    } catch (e) { /* file not found; ignore */ }
+    return result
+}
+```
+
+Access the client ID at runtime using `BuildConfig.CLIENT_ID`.
```
kgangineni: upto this line, all info is regarding how to inject Ciient Id, which merchants apps can chose however they want, it can be flavors, properties file or settings

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+
+### Configure deep links
+
+PayPal web checkout and PayPal vault flows use a browser-switch — the SDK opens a Chrome Custom Tab or the PayPal app, and PayPal redirects the customer back to your app when they finish. Be sure to configure your app to receive this redirect.
```
kgangineni: browser-switch is the name of a dependency internal to our repo; this doesn't mean anything to outside our team

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+    android:exported="true"
+    android:launchMode="singleTop">
+
+    <!-- Standard launcher intent-filter -->
+    <intent-filter>
+        <action android:name="android.intent.action.MAIN" />
+        <category android:name="android.intent.category.LAUNCHER" />
+    </intent-filter>
+
+    <!-- Custom URL scheme for browser-switch return -->
```
kgangineni: better for this to be

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+
+Use your HTTPS URL as the `appLinkUrl` in SDK requests: `"https://your-app-domain.example.com"`.
+
+> **Architecture warning:** Only the activity that declares these intent-filters can receive browser-switch returns. Based on the Demo app pattern in `Demo/src/main/AndroidManifest.xml`, a single activity (`MainActivity`) owns both intent-filters. If your app has multiple activities, route all SDK flows through the activity that declares the deep link intent-filters.
+
+---
+
+## App architecture
+
+### Recommended architecture
```
kgangineni: Not sure if this section is necessary, or can rename the section to Demo app architecture rather than recommended architecture, because in reality, apps are much bigger than our demo app

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+Register two effects in your composable to handle the return. Both are needed because Android can deliver the intent in two ways depending on whether the activity was in the foreground.
+
+```kotlin
+import com.paypal.android.cardpayments.CardFinishApproveOrderResult
+
+// In your composable screen
+OnLifecycleOwnerResumeEffect {
+    // Called when the activity returns to the foreground (covers the common case)
+    val intent = context.getActivityOrNull()?.intent
+    intent?.let { viewModel.completeAuthChallenge(it) }
```
kgangineni: this is not wrong, but `completeAuthChallenge` can go anywhere doesn't necessarily be in `ViewModel`, I think claude here again just following what's in demo app, instead I will put it more like 
After returning to app, use SDK API to complete transation using 
```kotlin
 cardClient.finishApproveOrder(intent)?.let { result ->
        // finishApproveOrder returns null if the intent is not for this auth session
        when (result) {
            is CardFinishApproveOrderResult.Success -> {
                // Authentication succeeded; capture/authorize the order on your server
                val orderStatus = result.status  // nullable — check before use
// make server call to capture order in your server
            }

            is CardFinishApproveOrderResult.Failure ->
// Handle UI for failure scenarios

            CardFinishApproveOrderResult.Canceled ->
                // Handle UI for Cancelled scenarios

            CardFinishApproveOrderResult.NoResult -> {
                // This intent was not for this auth session
                // Re-enable the pay button so the customer can retry
                resetToIdleState()
            }
        }
    }
```

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,769 @@
...
+After the flow completes, capture or authorize the order on your server.
+
+```kotlin
+// After finishStart returns Success
+fun capturePayPalOrder(orderId: String?) {
+    if (orderId == null) {
+        showError("Order ID is missing")
+        return
+    }
+    yourServer.captureOrder(orderId)
```
kgangineni, line 811: `yourServer.captureOrder(orderId)` this format is nice, to tell merchant it needs to be performed in their server 👍

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+
+### Deep link return
+
+Register both effects in your composable. Call `finishStart()` in your ViewModel.
+
+```kotlin
+import com.paypal.android.paypalwebpayments.PayPalWebCheckoutFinishStartResult
+
+// In your composable
+OnLifecycleOwnerResumeEffect {
```
kgangineni: this is how our demo app does
`OnLifecycleOwnerResume`, `onNewIntentEffect`, but merchant apps can chose however they want, it's better to say it override `OnResume`, `OnNewIntent` functions in activity to handle return to app, same comment above also

kgangineni:
path for this file can be `Docs/IntegrationGuide.md` rather than `Docs/IntegrationGuide/draft.md`

Docs/IntegrationGuide/draft.md
naolson-paypal: @kgangineni so are you recommending that lines 77-106 are just removed? Or more?

Docs/IntegrationGuide/draft.md
naolson-paypal: Interesting feedback - I kind of assumed that Browser-Switch made sense outside of the repo context. How would you recommend we describe this?

Docs/IntegrationGuide/draft.md
naolson-paypal: Can you explain a bit more? I'm not sure I understand what your feedback is here

Docs/IntegrationGuide/draft.md
naolson-paypal: so you would recommend lines 167-212 be either removed or clearly presented as an example of the architecture in the Demo App?

Docs/IntegrationGuide/draft.md
kgangineni: remove this part "PayPal web checkout and PayPal vault flows use a browser-switch", rest all is good,

Docs/IntegrationGuide/draft.md
kgangineni: better for this to be 
`<!-- Custom URL scheme for returning to app -->`

Docs/IntegrationGuide/draft.md
kgangineni: Ya I will say remove it, all Sdk needs is `CoreConfig` how to create it is covered in other sections

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
+# PayPal Android SDK Integration Guide
+
+## Overview
+
+This guide shows you how to accept card payments and PayPal payments in your Android app using the PayPal Android SDK. You learn how to set up the SDK, initialize clients, process one-time payments, and vault payment methods for future use. Before you begin, review the Prerequisites section to make sure your environment is ready.
```
sshropshire:
```suggestion
This guide shows you how to accept card payments and PayPal payments in your Android app using the PayPal Android SDK. You learn how to set up the SDK, process one-time payments, and vault payment methods for future use. Before you begin, review the Prerequisites section to make sure your environment is ready.
```
I would remove "initialize clients" here because It isn't a standard term. "Set up the SDK" will be enough on its own.

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+
+### Add SDK dependencies
+
+Add the modules you need to your app-level `build.gradle` file. Use the current release version.
+
+```groovy
+dependencies {
+    // Required: add at least one payment module
+
+    // For card payments (card-present and card-not-present)
```
sshropshire:
```suggestion
    // For card payments
```
I'd remove the card present, card not present distinction here. I believe these terms are technically meant for physical point of sale terminals.

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+    implementation 'com.paypal.android:paypal-web-payments:2.3.0'
+
+    // Optional: collect device data to help reduce fraud risk
+    implementation 'com.paypal.android:fraud-protection:2.3.0'
+
+    // Optional: PayPal payment button UI components
+    implementation 'com.paypal.android:payment-buttons:2.3.0'
+
+    // Optional: Venmo payments
+    implementation 'com.paypal.android:venmo:2.3.0'
```
sshropshire:
```suggestion
```
This can be removed until we have official support for Venmo.

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+
+```groovy
+repositories {
+    maven {
+        url 'https://central.sonatype.com/repository/maven-snapshots/'
+    }
+}
+```
+
+Then reference the snapshot version:
```
sshropshire:
```suggestion
Then reference the latest snapshot version:
```

Docs/IntegrationGuide/draft.md
sshropshire: Yeah `CLIENT_ID` doesn't require the same protections as a `CLIENT_SECRET` does. They can place `CLIENT_ID` directly into their Android codebase without any issues.

Docs/IntegrationGuide/draft.md
sshropshire: Yeah we've migrated away from the Braintree `browser-switch` library. We use Chrome Custom Tabs directly in our SDK.

Docs/IntegrationGuide/draft.md
sshropshire:
```suggestion
    <!-- Custom URL scheme for returning to app -->
```

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+    <data
+        android:host="your-app-domain.example.com"
+        android:pathPrefix="/"
+        android:scheme="https" />
+</intent-filter>
+```
+
+Use your HTTPS URL as the `appLinkUrl` in SDK requests: `"https://your-app-domain.example.com"`.
+
+> **Architecture warning:** Only the activity that declares these intent-filters can receive browser-switch returns. Based on the Demo app pattern in `Demo/src/main/AndroidManifest.xml`, a single activity (`MainActivity`) owns both intent-filters. If your app has multiple activities, route all SDK flows through the activity that declares the deep link intent-filters.
```
sshropshire:
```suggestion
> **Architecture warning:** Only the activity that declares these intent-filters can receive deep links back into the host application. If your app has multiple activities, route all SDK flows through the activity that declares the deep link intent-filters.
```

Docs/IntegrationGuide/draft.md
sshropshire: We also have two Demo apps so we might need to clarify to merchants that the merchant facing demo app is the one they should be looking at for reference.

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+
+---
+
+## App architecture
+
+### Recommended architecture
+
+The Demo app uses Jetpack Compose Navigation with a single entry-point activity. All SDK flows live within `MainActivity`, which hosts a `NavHost`. Individual payment screens are composable destinations. Deep link returns arrive at `MainActivity` and are dispatched to the active ViewModel.
+
+This single-entry-point pattern is recommended because it avoids routing ambiguity: the OS delivers deep link intents to `MainActivity`, which is the only activity that declares the browser-switch intent-filters.
```
sshropshire:
```suggestion
This single-entry-point pattern is recommended because it avoids routing ambiguity: the OS delivers deep link intents to `MainActivity`, which is the only activity that declares the intent-filters for deep linking.
```

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+savedStateHandle.get<String>("card_client_state")?.let { savedState ->
+    cardClient.restore(savedState)
+}
+```
+
+The same `instanceState` / `restore()` pattern applies to `PayPalWebCheckoutClient`.
+
+---
+
+## Initialize the SDK
```
sshropshire: I feel like this section could me moved up, at least before Process Death is mentioned.

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+
+Create a `CoreConfig` object with your client ID and environment. Then create the module clients you need. The Demo app creates clients inside ViewModel constructors so they are scoped to the screen lifecycle.
```
sshropshire:
```suggestion
Create a `CoreConfig` object with your client ID and environment. Then create a client for each payment method you need. The Demo app creates clients inside ViewModel constructors so they are scoped to the screen lifecycle.
```

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+
+```kotlin
+import com.paypal.android.corepayments.CoreConfig
+import com.paypal.android.corepayments.Environment
+import com.paypal.android.cardpayments.CardClient
+import com.paypal.android.paypalwebpayments.PayPalWebCheckoutClient
+import com.paypal.android.fraudprotection.PayPalDataCollector
```
sshropshire: We should consider omitting the import statements. They will be automatically added by the Android Studio IDE. 
```suggestion
```

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+    expirationMonth = "01",               // 2-digit month (MM)
+    expirationYear = "2026",              // 4-digit year (YYYY)
+    securityCode = "123",                 // CVV/CVC
+    cardholderName = "Jane Doe",          // optional
+    billingAddress = null                 // optional
+)
+
+// Build the request
+// returnUrl must match a URL scheme registered in your AndroidManifest.xml
+val returnUrl = "com.example.myapp://"   // or your HTTPS app link URL
```
sshropshire:
```suggestion
val returnUrl = "com.example.myapp://"    // or your HTTPS app link URL
```

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+    sca = SCA.SCA_WHEN_REQUIRED           // default: only trigger 3D Secure when required
+)
+
+// Call approveOrder — result arrives asynchronously on the main thread
+cardClient.approveOrder(cardRequest) { result ->
+    when (result) {
+        is CardApproveOrderResult.Success -> {
+            // No 3D Secure challenge was needed; tell your server to capture/authorize
+            val orderStatus = result.status   // nullable String — check before use
+            val attempted3DS = result.didAttemptThreeDSecureAuthentication
```
sshropshire:
```suggestion
```
We may not need these property references in the example since they aren't used. The `orderId` property is the main thing that matters here for capture/authorize, and it is non-null.

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+
+When 3D Secure authentication is required, the SDK opens a Chrome Custom Tab. After the customer completes authentication, PayPal redirects them back to your app using the `returnUrl` you provided.
+
+Register two effects in your composable to handle the return. Both are needed because Android can deliver the intent in two ways depending on whether the activity was in the foreground.
+
+```kotlin
+import com.paypal.android.cardpayments.CardFinishApproveOrderResult
+
+// In your composable screen
+OnLifecycleOwnerResumeEffect {
```
sshropshire: Both `OnLifecycleOwnerResumeEffect` and `OnNewIntentEffect` are only defined in our demo app. They are not standard Android library calls. This would result in a compilation error for the merchant most likely. We could provide those helper methods in this guide. Both are less than 10 lines of code.

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+}
+
+// In your ViewModel
+fun completeAuthChallenge(intent: Intent) {
+    cardClient.finishApproveOrder(intent)?.let { result ->
+        // finishApproveOrder returns null if the intent is not for this auth session
+        when (result) {
+            is CardFinishApproveOrderResult.Success -> {
+                // Authentication succeeded; capture/authorize the order on your server
+                val orderStatus = result.status  // nullable — check before use
```
sshropshire:
```suggestion
```
We can remove this property reference as well. Technically the enum `.Success` type itself is the status.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,769 @@
...
+
+The vault flow has three steps: your server creates a setup token, the SDK attaches the card to the setup token, and your server creates a payment token from the setup token.
+
+### What you'll do
+
+Collect card details, call `cardClient.vault()`, handle any 3D Secure challenge, then tell your server to create a payment token.
+
+### Server call
+
+Your server creates a setup token and returns the setup token ID. See the [PayPal vaulting documentation](https://developer.paypal.com/docs/) for the server-side implementation.
```
sshropshire, line 610: We may need to correct this PayPal vaulting documentation link. It just points to developer.paypal.com/docs and also it's attempting to link the reader to PayPal docs, but we're in the Card section.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,769 @@
...
+            is CardFinishVaultResult.Success -> {
+                val setupTokenStatus = result.status  // nullable — check before use
+                createPaymentTokenOnServer(result.setupTokenId)
+            }
+
+            is CardFinishVaultResult.Failure ->
+                showError(result.error.errorDescription)
+
+            CardFinishVaultResult.Canceled ->
+                resetToIdleState()
```
sshropshire, line 680: This is going to be highly dependent on merchant UI. If we're talking Demo app though, we're basically removing the loading spinner from the "Submit" button, which allows the user to launch the flow again in the case of Cancelation (or if there was no result).

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,769 @@
...
+import com.paypal.android.paypalwebpayments.PayPalWebCheckoutFundingSource
+import com.paypal.android.paypalwebpayments.PayPalPresentAuthChallengeResult
+
+val checkoutRequest = PayPalWebCheckoutRequest(
+    orderId = orderId,                                           // from your server
+    fundingSource = PayPalWebCheckoutFundingSource.PAYPAL,       // PAYPAL, PAYPAL_CREDIT, or PAY_LATER
+    appSwitchWhenEligible = true,                                // switch to PayPal app if installed
+    appLinkUrl = "https://your-app-domain.example.com",          // HTTPS return URL (or null)
+    fallbackUrlScheme = "com.example.myapp"                      // custom scheme fallback (or null)
+)
```
sshropshire, line 768:
```suggestion
// for deep linking with app links
val appLinkStrategy = ReturnToAppStrategy.AppLink("https://your-app-domain.example.com")

// for deep linking with custom url scheme
val customUrlSchemeStrategy = ReturnToAppStrategy.CustomUrlScheme("com.example.myapp")

val checkoutRequest = PayPalWebCheckoutRequest(
    orderId = orderId,                                           // from your server
    fundingSource = PayPalWebCheckoutFundingSource.PAYPAL,       // PAYPAL, PAYPAL_CREDIT, or PAY_LATER
    appSwitchWhenEligible = true,                                // switch to PayPal app if installed
    returnToAppStrategy = appLinkStrategy ?: customUrlSchemeStrategy // choose the strategy that works best for your app
)
```
We changed this recently to make it more clear and have merchants explicitly declare the deep link strategy they want the SDK to use.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,769 @@
...
+```kotlin
+import com.paypal.android.paypalwebpayments.PayPalWebVaultRequest
+import com.paypal.android.paypalwebpayments.PayPalPresentAuthChallengeResult
+
+val vaultRequest = PayPalWebVaultRequest(
+    setupTokenId = setupTokenId,                                 // from your server
+    appSwitchWhenEligible = true,                                // switch to PayPal app if installed
+    appLinkUrl = "https://your-app-domain.example.com",          // HTTPS return URL (or null)
+    fallbackUrlScheme = "com.example.myapp"                      // custom scheme fallback (or null)
+)
```
sshropshire, line 874: We can update this snippet with the new `ReturnToAppStrategy` type as well.

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+- Card payment triggers 3D Secure challenge — customer completes authentication
+- Card payment triggers 3D Secure challenge — customer cancels
+- PayPal checkout completes
+- PayPal checkout — customer cancels in Chrome Custom Tab
+- PayPal checkout — app switch to PayPal app (if testing `appSwitchWhenEligible = true`)
+- Process death recovery — stop the app while authentication is in progress, return to it, and verify the flow completes correctly
+
+**Test process death recovery**
+
+1. Start a checkout flow that requires browser-switch (PayPal web checkout or card 3D Secure).
```
sshropshire:
```suggestion
1. Start a checkout flow that requires authentication via Chrome Custom Tab (PayPal web checkout or card 3D Secure).
```

Docs/IntegrationGuide/draft.md
```diff
@@ -0,0 +1,769 @@
...
+4. Verify that your `restore()` call rebuilds the SDK client state and that `finishStart()` or `finishApproveOrder()` returns the correct result.
+
+---
+
+## Go live checklist
+
+Before going live, complete the following:
+
+- [ ] Replace `Environment.SANDBOX` with `Environment.LIVE` in your `CoreConfig`
+- [ ] Replace your sandbox client ID with your live client ID in `paypal.properties`
```
sshropshire:
```suggestion
- [ ] Replace your sandbox client ID with your live client ID
```

naolson-paypal:
@sshropshire and @kgangineni - Guide was updated based on your feedback - Please review this iteration.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,688 @@
+# PayPal Android SDK Integration Guide
+
+## Overview
+
+This guide shows you how to accept card payments and PayPal payments in your Android app using the PayPal Android SDK. You learn how to set up the SDK, process one-time payments, and vault payment methods for future use. Before you begin, review the Prerequisites section to make sure your environment is ready.
```
sshropshire: It would be nice to hyperlink to the Prerequisites section.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,688 @@
...
+
+<PlaceholderImage description="End-to-end flow diagram showing the three actors: your Android app (SDK), your server, and PayPal APIs. Arrows show: app collects payment details → app calls SDK → SDK contacts PayPal → PayPal returns result to SDK → SDK returns result to app → app calls your server to capture/authorize." />
+
+---
+
+## How the PayPal Android SDK works
+
+The SDK handles the client-side portion of your payment integration. For card payments, the SDK submits card data directly to PayPal using encrypted channels and returns a result your server uses to capture or authorize the order. For PayPal payments, the SDK opens a Chrome Custom Tab (or the PayPal app if installed) so the customer can log in and approve the payment. When the customer finishes, PayPal redirects them back to your app using a deep link. Your server then captures or authorizes the approved order using the Orders v2 API.
```
sshropshire: In the places where we mention Chrome Custom Tab or PayPal app launch, I wonder if it would make more sense to always mention app switch first to give it priority in the onboarding process.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,688 @@
...
+    // Optional: PayPal payment button UI components
+    implementation 'com.paypal.android:payment-buttons:2.3.0'
+}
+```
+
+All modules are published to Maven Central under the `com.paypal.android` group.
+
+**Snapshot builds**
+
+To test upcoming features before they ship, use snapshot builds. First, add the snapshot repository to your top-level `build.gradle`:
```
sshropshire, line 146:
```suggestion
If you'd like to test upcoming features before they ship, use snapshot builds. First, add the snapshot repository to your top-level `build.gradle`:
```
This feels more natural tonally. The number of our merchants who use snapshot builds should be an extreme minority.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,688 @@
...
+
+Then reference the latest snapshot version:
+
+```groovy
+implementation 'com.paypal.android:card-payments:2.3.0-SNAPSHOT'
+```
+
+### Configure deep links
+
+The SDK opens a Chrome Custom Tab or the PayPal app for certain flows (PayPal web checkout, PayPal vaulting, 3D Secure card authentication). When the customer finishes, PayPal redirects them back to your app using a deep link. Configure your app to receive this redirect.
```
sshropshire:
```suggestion
The SDK opens a Chrome Custom Tab or the PayPal app for certain flows (PayPal web checkout, PayPal vaulting, 3D Secure card authentication). When the customer finishes, PayPal redirects them back to your app using a deep link. You'll need to configure your app to receive this redirect.
```
Another tonal change. Mostly to highlight the importance and criticality of configuring the host application to receive deep links.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,688 @@
...
+
+The SDK supports two return URL strategies. Choose one or configure both as a fallback:
```
sshropshire: After our last refactor I'm not sure if merchants can still configure a "fallback" scenario when app links don't work. We haven't discussed the need for this feature, so we took it out. It didn't seem necessary and was somewhat out of scope for the completion of the app switch feature.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,688 @@
...
+    <intent-filter>
+        <action android:name="android.intent.action.VIEW"/>
+        <data android:scheme="com.example.myapp"/>
+        <category android:name="android.intent.category.DEFAULT"/>
+        <category android:name="android.intent.category.BROWSABLE"/>
+    </intent-filter>
+</activity>
+```
+
+Use this scheme as your `fallbackUrlScheme` in SDK requests: `"com.example.myapp"`.
```
sshropshire: Again the property `fallbackUrlScheme` no longer exists.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,688 @@
...
+    <!-- Standard launcher intent-filter -->
+    <intent-filter>
+        <action android:name="android.intent.action.MAIN" />
+        <category android:name="android.intent.category.LAUNCHER" />
+    </intent-filter>
+
+    <!-- Custom URL scheme for returning to app -->
+    <intent-filter>
+        <action android:name="android.intent.action.VIEW"/>
+        <data android:scheme="com.example.myapp"/>
```
sshropshire, line 223:
```suggestion
        <data android:scheme="com.example.your-app-domain"/>
```
If we want to be consistent with the app links example, we should use reverse DNS convention for the custom url scheme.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,688 @@
...
+implementation 'com.paypal.android:card-payments:2.3.0-SNAPSHOT'
+```
+
+### Configure deep links
+
+The SDK opens a Chrome Custom Tab or the PayPal app for certain flows (PayPal web checkout, PayPal vaulting, 3D Secure card authentication). When the customer finishes, PayPal redirects them back to your app using a deep link. Configure your app to receive this redirect.
+
+The SDK supports two return URL strategies. Choose one or configure both as a fallback:
+
+**Option 1: Custom URL Scheme**
```
sshropshire: Similar to how we should more heavily promote app switch as a primary option, App Links would be a better Option 1 to give it priority during the onboarding process.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,688 @@
...
+    <category android:name="android.intent.category.DEFAULT" />
+    <category android:name="android.intent.category.BROWSABLE" />
+    <data
+        android:host="your-app-domain.example.com"
+        android:pathPrefix="/"
+        android:scheme="https" />
+</intent-filter>
+```
+
+Use your HTTPS URL as the `appLinkUrl` in SDK requests: `"https://your-app-domain.example.com"`.
```
sshropshire: These examples will need to be updated to use the new `ReturnToAppStrategy` parameter.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,688 @@
...
+
+```kotlin
+// Create configuration (use Environment.LIVE for production)
+val coreConfig = CoreConfig(
+    clientId = "your-client-id",
+    environment = Environment.SANDBOX
+)
+
+// Create the card payments client
+val cardClient = CardClient(context, coreConfig)
```
sshropshire, line 254:
```suggestion
val cardClient = CardClient(applicationContext, coreConfig)
```
Using `applicationContext` as an example is less restrictive, but this is an opinion and not a requirement for the docs to be correct (they already are).

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,688 @@
...
+`CoreConfig` parameters:
+
+| Parameter | Type | Required | Description |
+|---|---|---|---|
+| `clientId` | String | Yes | Your PayPal client ID from the Developer Dashboard |
+| `environment` | Environment | No | `Environment.SANDBOX` (default) or `Environment.LIVE` |
+
+---
+
+## Process death recovery
```
sshropshire:
```suggestion
## Process kill recovery
```
Google does tend to use the phrase "process kill" more frequently than "process death" ([source](https://developer.android.com/guide/components/activities/process-lifecycle)).

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,688 @@
...
+}
+```
+
+The same `instanceState` / `restore()` pattern applies to `PayPalWebCheckoutClient`.
+
+---
+
+## Demo app architecture
+
+The [Demo app](Demo/) included in this repository uses Jetpack Compose Navigation with a single entry-point activity. All SDK flows live within `MainActivity`, which hosts a `NavHost`. Individual payment screens are composable destinations. Deep link returns arrive at `MainActivity` and are dispatched to the active screen.
```
sshropshire:
```suggestion
The [Demo app](Demo/) included in this repository uses Jetpack Compose Navigation with a single entry-point activity. All SDK flows live within `MainActivity`, which contains a `NavHost`. Individual payment screens are composable destinations. Deep link returns arrive at `MainActivity` and are dispatched to the active screen.
```
Verbage nit. Best to avoid overloaded terms to lessen the cognitive load on the reader.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,688 @@
...
+
+Override `onResume()` and `onNewIntent()` in your activity to handle the return. Both are needed because Android can deliver the intent in two ways depending on whether the activity was in the foreground.
+
+```kotlin
+class MainActivity : ComponentActivity() {
+
+    override fun onResume() {
+        super.onResume()
+        // Called when the activity returns to the foreground
+        handleReturnIntent(intent)
```
sshropshire, line 506: This is better since it doesn't require those utility methods from our Demo app to work, but we should have a Jetpack compose example somewhere in the docs. For new apps, this integration pattern will be considered obsolete and can lead to some bad architectural decisions that explode into technical debt as merchant apps mature.

naolson-paypal:
@sshropshire and @kgangineni - Let me know what you think about this pass at the docs

Docs/IntegrationGuide.md
kgangineni: `ReturnToAppStrategy` is not available in 2.3.0

kgangineni (commented):
rest LGTM
except on `ReturnToAppStrategy`

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+    participant PayPal as PayPal APIs
+
+    App->>Server: Create order
+    Server->>PayPal: Orders v2 API (create)
+    PayPal-->>Server: orderId
+    Server-->>App: orderId
+    App->>SDK: approveOrder(cardRequest)
+    SDK->>PayPal: Confirm payment source (REST API)
+    PayPal-->>SDK: Approval result
+    SDK-->>App: Success or AuthorizationRequired
```
sshropshire, line 78:
```suggestion
    SDK-->>App: Success or Authorization Required
```

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+
+## How the PayPal Android SDK works
+
+The SDK handles the client-side portion of your payment integration. For card payments, the SDK submits card data directly to PayPal using encrypted channels and returns a result your server uses to capture or authorize the order. For PayPal payments, the SDK launches a checkout experience where the customer logs in and approves the payment. When the customer finishes, PayPal redirects them back to your app using a deep link. Your server then captures or authorizes the approved order using the Orders v2 API.
+
+### App switch vs. Chrome Custom Tab
+
+The SDK supports two ways to present PayPal checkout:
+
+- **App switch** — If the customer has the PayPal app installed and the merchant opts in with `appSwitchWhenEligible = true`, the SDK checks eligibility with PayPal's backend. If eligible, the SDK launches the PayPal app directly. This provides a native experience where the customer may already be logged in, reducing friction.
```
sshropshire, line 95:
```suggestion
- **App switch** — If the customer has the PayPal app installed and you set `appSwitchWhenEligible = true` in your app, the SDK checks eligibility with PayPal's backend. If eligible, the SDK launches the PayPal app directly. This provides a native experience where the customer may already be logged in, reducing friction.
```
nit: There is a tone inconsistency I noticed. In the diagram, we say "Your Android App," and here we say "the merchant." It feels like we should pick one or the other, otherwise it looks like this doc was written for two different audiences.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+- **Chrome Custom Tab** — The default experience. The SDK opens a Chrome Custom Tab where the customer logs in to PayPal in a browser window.
+
+When app switch is enabled, the SDK handles eligibility automatically. If the PayPal app is not installed, the eligibility check fails, or the backend determines the customer is not eligible, the SDK silently falls back to Chrome Custom Tab. No additional error handling is needed — the fallback is invisible to both the merchant and the customer.
```
sshropshire, line 98:
```suggestion
When app switch is enabled, the SDK checks eligibility automatically. If the PayPal app is not installed, the eligibility check fails, or the backend determines the customer is not eligible, the SDK silently falls back to Chrome Custom Tab. No additional error handling is needed — the fallback is invisible to both the merchant and the customer.
```
🧶

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+
+```kotlin
+val returnToAppStrategy = ReturnToAppStrategy.AppLink(
+    appLinkUrl = "https://your-app-domain.example.com"
+)
+```
+
+**Option 2: Custom URL Scheme**
+
+Register a custom URL scheme in your `AndroidManifest.xml`. Your activity needs to use `android:launchMode="singleTop"` so that the OS delivers the return intent to the existing activity instance rather than creating a new one.
```
sshropshire, line 206: The `singleTop` callout is relevant to both integration patterns. We should probably make the same callout in the app links section.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+)
+```
+
+> **Architecture warning:** Only the activity that declares these intent-filters can receive deep links back into the host application. If your app has multiple activities, route all SDK flows through the activity that declares the deep link intent-filters.
+
+---
+
+## Initialize the SDK
+
+Create a `CoreConfig` object with your client ID and environment. Then create a client for each payment method you need. The Demo app creates clients inside ViewModel constructors so they are scoped to the screen lifecycle.
```
sshropshire, line 244:
```suggestion
Create a `CoreConfig` object with your client ID and environment. Then create a client for each payment method you need. The Demo app creates clients inside Jetpack ViewModels so they are scoped to the screen lifecycle.
```
🧶 : technically it being in a ViewModel constructor does not directly tie it to the lifecycle of the host View. The fact that the payment client is owned by the ViewModel is the reason why it is scoped to the screen lifecycle.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+| Parameter | Type | Required | Description |
+|---|---|---|---|
+| `clientId` | String | Yes | Your PayPal client ID from the Developer Dashboard |
+| `environment` | Environment | No | `Environment.SANDBOX` (default) or `Environment.LIVE` |
+
+---
+
+## Process kill recovery
+
+Android may stop your app process while the customer is in the external browser completing authentication. When the customer returns, Android launches a fresh process and delivers the deep link intent. If your SDK client objects were destroyed during the process stop, they have no auth state and cannot process the return.
```
sshropshire, line 274:
```suggestion
Android may stop your app process while the customer is in the external browser completing authentication. When the customer returns, Android cold starts your app and delivers the deep link intent. If your SDK client objects were destroyed during the process stop, they have no auth state and cannot process the return.
```
🧶 : the technical term here is cold start.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+
+    App->>Server: Create order
+    Server->>PayPal: Orders v2 API (create)
+    PayPal-->>Server: orderId
+    Server-->>App: orderId
+    App->>SDK: approveOrder(cardRequest)
+    SDK->>PayPal: Confirm payment source (REST API)
+    PayPal-->>SDK: Success
+    SDK-->>App: CardApproveOrderResult.Success
+    App->>Server: Capture/authorize order
```
sshropshire, line 377:
```suggestion
    App->>Server: Capture/Authorize order
```
🧶

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+| `SCA.SCA_ALWAYS` | Triggers 3D Secure authentication even when not required. |
+
+### Handle the result
+
+After `approveOrder` returns `Success`, tell your server to capture or authorize the order. Optionally collect device data to help reduce fraud risk before the server call.
+
+```kotlin
+// Optional: collect device data to help reduce fraud risk before capturing
+val dataCollectorRequest = PayPalDataCollectorRequest(
+    hasUserLocationConsent = false  // set to true only if user consented to location data
```
sshropshire, line 485: They may already have a `clientMetadataId` from the initial server-side create order call. I can't remember if it's best for them to pass that in, or let the data collector generate a new one that they then pass to the `capture/authorize` call but we should verify.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+
+                is CardFinishApproveOrderResult.Failure ->
+                    showError(result.error.errorDescription)
+
+                CardFinishApproveOrderResult.Canceled ->
+                    // Customer canceled 3D Secure; re-enable the pay button
+                    resetToIdleState()
+
+                CardFinishApproveOrderResult.NoResult -> {
+                    // This intent was not for this auth session
```
sshropshire, line 532:
```suggestion
                    // The intent is not associated with this CardClient instance
```
nit: Since we don't mention auth sessions anywhere else in this guide (unless I'm mistaken). Auth session is more of an internal detail now.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+fun CardPaymentScreen(cardClient: CardClient) {
+    val context = LocalContext.current
+    val lifecycleOwner = LocalLifecycleOwner.current
+
+    // Handle return when the activity resumes (e.g., returning from Chrome Custom Tab)
+    val lifecycleState by lifecycleOwner.lifecycle.currentStateFlow.collectAsState()
+    LaunchedEffect(lifecycleState) {
+        if (lifecycleState == Lifecycle.State.RESUMED) {
+            val activity = context as? ComponentActivity ?: return@LaunchedEffect
+            activity.intent?.let { intent -> handleFinishApproveOrder(cardClient, intent) }
```
sshropshire, line 557:
```suggestion
            val activity = context as? ComponentActivity
            activity?.intent?.let { intent -> handleFinishApproveOrder(cardClient, intent) }
```
🧶 : this feels simpler.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+        }
+    }
+
+    // Handle return when a new intent is delivered to a singleTop activity
+    DisposableEffect(Unit) {
+        val activity = context as? ComponentActivity ?: return@DisposableEffect onDispose {}
```
sshropshire, line 563: I'm not entirely sure what's needed here to simplify this line of code, but the `return@` seems unnecessary and makes the code seem more complex than it is.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+            CardFinishApproveOrderResult.Canceled ->
+                resetToIdleState()
+            CardFinishApproveOrderResult.NoResult ->
+                resetToIdleState()
+        }
+    }
+}
+```
+
+> **Understanding `NoResult`:** `NoResult` means the SDK does not have enough information to determine the outcome — for example, if the customer returned to the app without completing authentication, or if the deep link intent was not associated with the current auth session. The payment may or may not have succeeded server-side. When you receive `NoResult`, query the [Orders v2 API](https://developer.paypal.com/docs/api/orders/v2/) from your server to check the current order status before showing an error or retrying.
```
sshropshire, line 590: I'm not opposed to using the term `auth session` externally, but it would probably help if we give it it's own definition in the docs somewhere.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+
+**Test app switch scenarios**
+
+App switch requires a physical device with the PayPal app installed. Test the following:
+
+- PayPal checkout with `appSwitchWhenEligible = true` and PayPal app installed — verify the PayPal app opens and the customer can approve the payment
+- PayPal checkout with `appSwitchWhenEligible = true` and PayPal app **not** installed — verify the SDK falls back to Chrome Custom Tab without errors
+- PayPal vault with `appSwitchWhenEligible = true` — verify the vault flow works through the PayPal app
+- App switch return — verify that `finishStart()` / `finishVault()` returns the correct result when the customer returns from the PayPal app
+- Eligibility fallback — if the backend determines the customer is not eligible for app switch, verify the SDK silently falls back to Chrome Custom Tab
```
sshropshire, line 974: This one may be hard for merchants to test.

Docs/IntegrationGuide.md
```diff
@@ -0,0 +1,1031 @@
...
+
+Before going live, complete the following:
+
+- [ ] Replace `Environment.SANDBOX` with `Environment.LIVE` in your `CoreConfig`
+- [ ] Replace your sandbox client ID with your live client ID
+- [ ] Verify deep link configuration works in a release build (not just debug)
+- [ ] If using App Links, confirm your `assetlinks.json` file is hosted and verified on your live domain
+- [ ] Test process death recovery on a physical device running a live-like build
+- [ ] Confirm that all result variants (`Success`, `Failure`, `Canceled`, `NoResult`) are handled in every callback
+- [ ] Confirm that nullable return values (`status`, `orderId`, `payerId`) are null-checked before use
```
sshropshire, line 997: nit: This should be compiler enforced on the Kotlin-side, though Java null checks may still be required. Ideally our API should remove the possibility of nullability as well, but we'll have to wait until we can make breaking changes to address those issues with our API.

sshropshire (approved):

PR #431, opened 2026-03-17
Card vault automation tests tests

kgangineni:
### Summary of changes
 - automates tests with 3DS check and without 3DS check
 - adds util functions to check if Chrome is visible or not

 ### Checklist

 - [ ] Added a changelog entry

-


Demo/src/androidTest/java/com/paypal/android/robots/DemoRobot.kt
```diff
@@ -239,4 +262,118 @@ class DemoRobot(
...
+        composeTestRule.waitUntilExactlyOneExists(
+            hasText("Create Setup Token")
+        )
+    }
+
+    fun createSetupToken(
+        returnToAppStrategy: ReturnToAppStrategyOption,
+        sca: com.paypal.android.cardpayments.threedsecure.SCA
+    ) = apply {
+        setSCA(sca)
```
sshropshire, line 281:
```suggestion
        selectSCAOption(sca)
```
nit: This reads a bit better from a testing perspective. `setSCA()` feels more like it's happening through code instead of through user interaction.

sshropshire (approved):

PR #432, opened 2026-03-19
append path to return url in confirm-payment source

kgangineni:
### Summary of changes
 - Append path to returnUrl in confirmPaymentSource, when returnUrl is `CustomUrlScheme://` call fails

 ### Checklist

 - [ ] Added a changelog entry

-


Demo/src/main/java/com/paypal/android/ui/approveorder/ApproveOrderViewModel.kt
```diff
@@ -84,7 +84,10 @@ class ApproveOrderViewModel @Inject constructor(
                 securityCode = cardSecurityCode
             )
             val returnUrl =
-                ReturnUrlFactory.createGenericReturnUrl(returnToAppStrategyOption.toReturnToAppStrategy())
+                ReturnUrlFactory.createGenericReturnUrl(
+                    returnToAppStrategyOption.toReturnToAppStrategy(),
+                    "returnPath"
```
sshropshire:
```suggestion
                    "return-path"
```
nit: Camel case feels odd for a URL.

sshropshire (approved):

PR #438, opened 2026-04-30
Add Google Pay Implementation

sshropshire:
### Summary of changes

 - Add Google Pay End-to-end implementation
 - Unit tests and documentation will follow in a separate PR

 ### Checklist

- @sshropshire

### DIFF
--- a/GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayAPI.kt
+++ b/GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayAPI.kt
@@ -26,10 +26,12 @@ internal class GooglePayAPI(
     )
 
     @OptIn(InternalSerializationApi::class)
-    suspend fun getGooglePayConfig(): SDKResult<GooglePayConfig> {
+    suspend fun getGooglePayConfig(merchantId: String): SDKResult<GooglePayConfig> {
         @RawRes val resId = R.raw.graphql_google_pay_config_sandbox
         return when (val result = resourceLoader.loadRawResource(applicationContext, resId)) {
-            is LoadRawResourceResult.Success -> sendGraphQLGooglePayConfigRequest(result.value)
+            is LoadRawResourceResult.Success ->
+                sendGraphQLGooglePayConfigRequest(result.value, merchantId)
+
             is LoadRawResourceResult.Failure -> TODO("signal error")
         }
     }
@@ -37,15 +39,13 @@ internal class GooglePayAPI(
     @OptIn(InternalSerializationApi::class)
     private suspend fun sendGraphQLGooglePayConfigRequest(
         query: String,
+        merchantId: String
     ): SDKResult<GooglePayConfig> {
         val variables = GetGooglePayConfigVariables(
             clientId = coreConfig.clientId,
-            // TODO: allow this to be provided as a param
-            merchantId = listOf("V9YP27HFNG2LW"),
+            merchantId = listOf(merchantId),
             // TODO: see if we need this
             merchantOrigin = "com.paypal.android.sdk",
-//            // TODO: allow this to be provided as a param
-//            buyerCountry = "US"
         )
 
         val graphQLRequest = GraphQLRequest(
--- a/GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayAPI.kt
+++ b/GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayAPI.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import androidx.annotation.RawRes
 import com.paypal.android.corepayments.CoreConfig
 import com.paypal.android.corepayments.LoadRawResourceResult
+import com.paypal.android.corepayments.PayPalSDKError
 import com.paypal.android.corepayments.ResourceLoader
 import com.paypal.android.corepayments.graphql.GraphQLClient
 import com.paypal.android.corepayments.graphql.GraphQLRequest
@@ -32,7 +33,11 @@ internal class GooglePayAPI(
             is LoadRawResourceResult.Success ->
                 sendGraphQLGooglePayConfigRequest(result.value, merchantId)
 
-            is LoadRawResourceResult.Failure -> TODO("signal error")
+            is LoadRawResourceResult.Failure -> {
+                val error =
+                    PayPalSDKError(123, "Unable to launch google pay: config query missing.")
+                SDKResult.Failure(error)
+            }
         }
     }
 
@@ -59,22 +64,19 @@ internal class GooglePayAPI(
             )
         return when (graphQLResponse) {
             is GraphQLResult.Success -> {
-                val responseData = graphQLResponse.response.data
-                if (responseData == null) {
-                    TODO("handle null response error")
+                val googlePayConfig = graphQLResponse.response.data?.googlePayConfig
+                if (googlePayConfig == null) {
+                    val error = PayPalSDKError(123, "Unable to launch google pay: config missing.")
+                    SDKResult.Failure(error)
                 } else {
-                    val googlePayConfig = responseData.googlePayConfig
-                    if (googlePayConfig == null) {
-                        TODO("handle error")
-                    } else {
-                        SDKResult.Success(googlePayConfig)
-                    }
+                    SDKResult.Success(googlePayConfig)
                 }
             }
 
             is GraphQLResult.Failure -> {
-                print(graphQLResponse)
-                TODO("handle graphql failure error")
+                val error =
+                    PayPalSDKError(123, "Unable to launch google pay: unable to fetch config.")
+                SDKResult.Failure(error)
             }
         }
     }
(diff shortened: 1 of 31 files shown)


GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayClient.kt
```diff
@@ -1,11 +1,126 @@
...
+        } else {
+            val error = PayPalSDKError(123, "GooglePay finish start failed.")
+            GooglePayFinishStartResult.Failure(error)
+        }
+    }
+
+    companion object {
+        fun createPaymentsClient(context: Context): PaymentsClient {
+            val walletOptions = Wallet.WalletOptions.Builder()
+                .setEnvironment(WalletConstants.ENVIRONMENT_TEST)
```
kgangineni: extend `CoreConfig` to contain environment for Google Pay, use it here dynamically

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayClient.kt
```diff
@@ -1,11 +1,126 @@
...
-        TODO("Implement Google Pay")
+    constructor(context: Context, config: CoreConfig) : this(
+        googlePayAPI = GooglePayAPI(context, config),
+        paymentsClient = createPaymentsClient(context)
+    )
+
+
+    @OptIn(InternalSerializationApi::class)
+    suspend fun start(request: GooglePayCheckoutRequest): GooglePayStartResult {
+        val merchantId = request.merchantId
```
kgangineni, line 30: should merchantId be coming from `GooglePayCheckoutRequest`, merchant passing everytime
this doesn't look like `start`, it's eligibility check for google pay

GooglePay/src/main/java/com/paypal/android/googlepay/LaunchGooglePay.kt
```diff
@@ -0,0 +1,50 @@
...
+    ): Intent {
+        return googlePayContract.createIntent(context, input.task)
+    }
+
+    override fun parseResult(
+        resultCode: Int,
+        intent: Intent?
+    ): GooglePayLaunchResult? {
+        val googlePayResult = googlePayContract.parseResult(resultCode, intent)
+        val success = googlePayResult.status.statusCode == CommonStatusCodes.SUCCESS
```
kgangineni: `val success = googlePayResult.status.statusCode in  listOf(CommonStatusCodes.SUCCESS, CommonStatusCodes.SUCCESS_CACHE)`
?

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayAPI.kt
```diff
@@ -0,0 +1,144 @@
...
+                    val error = PayPalSDKError(123, "Unable to launch google pay: config missing.")
+                    SDKResult.Failure(error)
+                } else {
+                    SDKResult.Success(googlePayConfig)
+                }
+            }
+
+            is GraphQLResult.Failure -> {
+                val error =
+                    PayPalSDKError(123, "Unable to launch google pay: unable to fetch config.")
```
kgangineni, line 79: need to define error codes

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayAPI.kt
```diff
@@ -0,0 +1,144 @@
...
+    @OptIn(InternalSerializationApi::class)
+    private suspend fun sendGraphQLGooglePayConfigRequest(
+        query: String,
+        merchantId: String
+    ): SDKResult<GooglePayConfig> {
+        val variables = GetGooglePayConfigVariables(
+            clientId = coreConfig.clientId,
+            merchantId = listOf(merchantId),
+            // TODO: see if we need this
+            merchantOrigin = "com.paypal.android.sdk",
```
kgangineni, line 54: this may be looking for host origin?

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayClient.kt
sshropshire: Can you explain this more?

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayClient.kt
sshropshire: Where should we put it?

GooglePay/src/main/java/com/paypal/android/googlepay/LaunchGooglePay.kt
sshropshire: What's `SUCCESS_CACHE`?

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayAPI.kt
sshropshire: Yeah definitely. We also probably don't need error codes. That's a holdover from iOS. That's a separate item though.

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayAPI.kt
sshropshire: You mean like the way JS does it?

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayClient.kt
kgangineni: we can change CoreConfig to include GooglePayEnvironment like below
```kotlin
data CoreConfig(
val clientId: String,
val environment: Environment = Environment.SANDBOX,
val googlePayEnvironment = (WalletConstants.ENVIRONMENT_TEST
)
```

GooglePay/src/main/java/com/paypal/android/googlepay/LaunchGooglePay.kt
kgangineni: `SUCCESS_CACHE` is when it already checked before and didn't performed realtime check and got values from cache, in our situation I think this is valid success case 
https://developers.google.com/android/reference/com/google/android/gms/common/api/CommonStatusCodes#SUCCESS_CACHE()

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayAPI.kt
kgangineni: guessing by the name, not sure if it's expecting appId or host of merchant

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayClient.kt
kgangineni: may be `CoreConfig` all details which will not change in every transaction stays in 1 place

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayClient.kt
sshropshire: But then core would have a dependency on Google Pay. So if someone is using CardPayments, they will need to pull in Google Pay into their project as well even though they aren't using that feature.

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayClient.kt
sshropshire: Google Pay is the only feature that requires a `merchantId`. I wouldn't feel comfortable doing that unless merchant ID was required for all payment methods.

GooglePay/src/main/java/com/paypal/android/googlepay/LaunchGooglePay.kt
sshropshire: Looking at the docs it doesn't seem like `SUCCESS_CACHE` is a possibility. This is a good callout in general though, I'll look into the other error codes mentioned [here](https://developers.google.com/android/reference/com/google/android/gms/common/api/CommonStatusCodes#SUCCESS_CACHE()).

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayAPI.kt
sshropshire: On Android the merchant isn't guaranteed to have a host. On JS I think they get this from [Window.location](https://developer.mozilla.org/en-US/docs/Web/API/Window/location)–I'm not aware of an equivalent on Android.

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayCheckoutRequest.kt
```diff
@@ -1,3 +1,3 @@
 package com.paypal.android.googlepay
 
-data class GooglePayCheckoutRequest(val merchantId: String?)
+data class GooglePayCheckoutRequest(val merchantId: String? = null)
```
kgangineni: we can get rid of merchantId in Request since we already have this in `CoreConfig` now

kgangineni (approved):

GooglePay/src/main/java/com/paypal/android/googlepay/GooglePayCheckoutRequest.kt
sshropshire: Good call yeah. I may have to remove `GooglePayCheckoutRequest` temporarily and we can define the API in another PR (the compiler won't allow a data class with no member variables).

PR #443, opened 2026-06-03
Improve Analytics

anibalb2500:
> Before submitting this PR, note that we cannot accept language translation PRs.
> We have a dedicated localization team to provide the translations. If there is
> an error in a specific translation, you may open an issue and we will escalate
> it to the localization team.

### Summary of changes

**PaymentButtons — align with established analytics pattern and fix fake credentials**

`PaymentButton` was creating an `AnalyticsService` with `clientId = "N/A"` and
`environment = Environment.LIVE`, causing `payment-button:initialized` and
`payment-button:tapped` to fire against live FPTI with invalid credentials on
every button instantiation — even in sandbox environments.

- Add `PaymentButtonEvent` enum with typed `INITIALIZED` and `TAPPED` constants,
  following the pattern established by `CheckoutEvent`/`VaultEvent`
  (PayPalWebPayments) and `ApproveOrderEvent`/`VaultEvent` (CardPayments)
- Add `PaymentButtonAnalytics` wrapper class delegating to `AnalyticsService`,
  consistent with `CardAnalytics` and `PayPalWebAnalytics` in other modules
- Replace the fake `AnalyticsService` with an optional `CoreConfig?` constructor
  parameter. If provided, a real `AnalyticsService` is created and events fire.
  If not provided (e.g. XML inflation), `analytics` is null and no events are
  sent — no fake credentials, no polluted live FPTI data
- Payment buttons declared in XML now support analytics via two new attributes,
  `app:paypal_client_id` and `app:paypal_environment`, removing the need to set
  `coreConfig` programmatically after inflation

**CorePayments — generic analytics params map**

`AnalyticsService.sendAnalyticsEvent` previously leaked checkout and button
concepts (`orderId`, `buttonType`) into the core layer. These are now replaced
with a generic `params: Map<String, String>` so `AnalyticsService` and
`AnalyticsEventData` have no knowledge of what the params mean.

- `sendAnalyticsEvent(name, orderId?, buttonType?)` →
  `sendAnalyticsEvent(name, params: Map<String, String> = emptyMap())`
- `AnalyticsEventData` now holds a `params` map instead of typed fields
- `TrackingEventsAPI` extracts known FPTI keys (`order_id`, `button_type`) from
  the map when building the serialized request — the wire format is unchanged
- Add `AnalyticsParams` object with `ORDER_ID` and `BUTTON_TYPE` constants,
  accessible across all SDK modules via `@RestrictTo(LIBRARY_GROUP)`
- New `setup_token_id` analytics param. 

Each analytics wrapper now fully owns its own parameter definitions:
- PayPalWebAnalytics     → mapOf(AnalyticsParams.ORDER_ID   to orderId)
- CardAnalytics          → mapOf(AnalyticsParams.ORDER_ID   to orderId)
- PaymentButtonAnalytics → mapOf(AnalyticsParams.BUTTON_TYPE to buttonType)

Adding a new FPTI param in the future requires only updating the relevant
wrapper and adding the key extraction in `TrackingEventsAPI` — the core
`AnalyticsService` and `AnalyticsEventData` classes need no changes.

**CorePayments — minor logging and documentation improvements**
- Extract `"[PayPal SDK]"` log tag into a named constant to eliminate duplication
- Correct log severity on analytics failure paths (`Log.d` → `Log.w` / `Log.e`)
- Add KDoc to `CoreConfig` documenting that FPTI analytics always route to the
  live endpoint regardless of the configured `Environment`

### Evidence
- Full PayPal web checkout flow validated end-to-end using token `2K552147JL722662P `
- CAL trace confirms `UpdateClientConfig.POST` fired correctly (76ms) and checkout
  completed successfully (`pay` call present, all 22 calls status 0)

https://github.com/user-attachments/assets/a3650af4-38b3-48cf-8240-55f4666eb356

Corresponding FPTI Showing all relevant FPTI events
image

### Checklist
- [x] Added a changelog entry

### Authors
- @anbojorquez


Demo/src/main/res/values/strings.xml
```diff
@@ -1,6 +1,9 @@
 <resources>
     <string name="app_name">AndroidSDK</string>
 
+    <!-- PayPal SDK credentials for Demo app -->
+    <string name="paypal_demo_client_id">AQTfw2irFfemo-eWG4H5UY-b9auKihUpXQ2Engl4G1EsHJe2mkpfUv_SN3Mba0v3CfrL6Fk_ecwv9EOo</string>
```
santugowda: Do we need to expose this PayPal client ID on a public repository ?

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
```diff
@@ -225,17 +248,26 @@ abstract class PaymentButton<C : PaymentButtonColor> @JvmOverloads constructor(
...
+            updateCoreConfigFrom(typedArray)
+        }
+    }
+
+    private fun updateCoreConfigFrom(typedArray: TypedArray) {
+        val clientId = typedArray.getString(R.styleable.PaymentButton_paypal_client_id)
+        if (!clientId.isNullOrBlank()) {
+            val environmentValue = typedArray.getInt(
+                R.styleable.PaymentButton_paypal_environment,
+                0 // default: sandbox
```
santugowda: Nit : Extract to a named constant like p`rivate const val ENVIRONMENT_ATTR_SANDBOX = 0 `
Right now this 0 is tied to `<enum name="sandbox" value="0"/> in attrs.xml.` 
If someone changes that enum value, this silently maps to the wrong environment.

CardPayments/src/main/java/com/paypal/android/cardpayments/analytics/CardAnalytics.kt
```diff
@@ -1,14 +1,27 @@
...
+            name = event.value,
+            params = buildParams(orderId)
+        )
     }
 
     fun notify(event: VaultEvent, setupTokenId: String?) {
-        analyticsService.sendAnalyticsEvent(event.value, setupTokenId)
+        analyticsService.sendAnalyticsEvent(
+            name = event.value,
+            params = buildParams(setupTokenId)
```
santugowda: Right now setupTokenId is mapped to AnalyticsParams.ORDER_ID. 

Add AnalyticsParams.SETUP_TOKEN_ID to avoid misleading vault event data.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/analytics/PayPalWebAnalytics.kt
```diff
@@ -1,14 +1,27 @@
...
+            name = event.value,
+            params = buildParams(orderId)
+        )
     }
 
     fun notify(event: VaultEvent, setupTokenId: String?) {
-        analyticsService.sendAnalyticsEvent(event.value, setupTokenId)
+        analyticsService.sendAnalyticsEvent(
+            name = event.value,
+            params = buildParams(setupTokenId)
```
santugowda: same as above add `AnalyticsParams.SETUP_TOKEN_ID  ` instead of sending setupTokenId as ORDER_ID.

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
anibalb2500: Removed

Demo/src/main/res/values/strings.xml
anibalb2500: Removed

CardPayments/src/main/java/com/paypal/android/cardpayments/analytics/CardAnalytics.kt
anibalb2500: Great catch. Added.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/analytics/PayPalWebAnalytics.kt
anibalb2500: Done.

PR #444, opened 2026-06-05
Improve Analytics

anibalb2500:
> Before submitting this PR, note that we cannot accept language translation PRs.
> We have a dedicated localization team to provide the translations. If there is
> an error in a specific translation, you may open an issue and we will escalate
> it to the localization team.

### Summary of changes

**PaymentButtons — introduce typed analytics wrapper**

`PaymentButton` previously fired analytics events inline using a bare `AnalyticsService`.
This PR introduces a thin typed wrapper layer consistent with how other modules handle analytics:

- Add `PaymentButtonEvent` enum with typed `INITIALIZED` and `TAPPED` constants, following
  the pattern established by `CheckoutEvent`/`VaultEvent` (PayPalWebPayments) and
  `ApproveOrderEvent`/`VaultEvent` (CardPayments)
- Add `PaymentButtonAnalytics` wrapper class delegating to `AnalyticsService`, consistent
  with `CardAnalytics` and `PayPalWebAnalytics` in other modules
- Refactor `setOnClickListener` override to dispatch `TAPPED` through the new wrapper

**CorePayments — generic analytics params and vault token fix**

`AnalyticsService.sendAnalyticsEvent` previously leaked checkout and button concepts
(`orderId`, `buttonType`) into the core layer. These are now replaced with a generic
`AnalyticsEventParams` data class so `AnalyticsService` and `AnalyticsEventData` have no
knowledge of what the params mean.

- `sendAnalyticsEvent(name, orderId?, buttonType?)` → `sendAnalyticsEvent(name, params: AnalyticsEventParams)`
- `AnalyticsEventParams` holds typed fields: `orderId`, `vaultSetupToken`, `buttonType`, `appSwitchEnabled`
- `TrackingEventsAPI` maps known FPTI keys (`order_id`, `vault_setup_token`) from the params
  struct when building the serialized request — the wire format is unchanged
- Fix: vault notify in `PayPalWebAnalytics` and `CardAnalytics` was passing setup token as
  `orderId`; now uses dedicated `vaultSetupToken` field
- New `vault_setup_token` FPTI param wired through `TrackingEventData`

**CorePayments — minor logging and documentation improvements**

- Extract `"[PayPal SDK]"` log tag into a named `TAG` constant to eliminate duplication
- Correct log severity on analytics failure paths (`Log.d` → `Log.w` / `Log.e`)
- Add KDoc to `CoreConfig` documenting that FPTI analytics always route to the live endpoint
  regardless of the configured `Environment`

#### Web checkout flow validated end-to-end using Order-ID `2DH291271Y007604B`

https://github.com/user-attachments/assets/02644785-c01c-4453-b836-ff896e08b92e

image

#### Web vault flow validated end-to-end using vault_setup_token `5C810075PJ396333L`

https://github.com/user-attachments/assets/299139e2-d9e5-48f3-a3f8-ce2fbfc7392e

image

#### Buttons Regression Testing

https://github.com/user-attachments/assets/c1c62b76-80eb-484a-84e0-a5e428ba83d8

image

**Note** I removed the following improvements from PaymentButton that were previously added, to stay in scope of our work
- Click Listener not working when added via XML
- Sending dynamic 'partner-client-id' analytics param as other modules. 

### Checklist

- [x] Added a changelog entry

### Authors

- @anbojorquez


CardPayments/src/main/java/com/paypal/android/cardpayments/analytics/CardAnalytics.kt
```diff
@@ -1,14 +1,27 @@
...
+            name = event.value,
+            params = buildParams(orderId)
+        )
     }
 
     fun notify(event: VaultEvent, setupTokenId: String?) {
-        analyticsService.sendAnalyticsEvent(event.value, setupTokenId)
+        analyticsService.sendAnalyticsEvent(
+            name = event.value,
+            params = setupTokenId?.let { mapOf(AnalyticsParams.SETUP_TOKEN_ID to it) } ?: emptyMap()
```
sshropshire: q: Would it make sense to make another `buildParams(setupTokenId: String?)` method here for consistency? Or maybe an `AnalyticsParams` data class that is convertible to `Map<String, String>` to avoid a method naming collision.

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
```diff
@@ -5,6 +5,5 @@ internal data class AnalyticsEventData(
     val environment: String,
     val eventName: String,
     val timestamp: Long,
-    val orderId: String?,
-    val buttonType: String? = null
+    val params: Map<String, String> = emptyMap()
```
sshropshire: We could make this a Nullable type, then we wouldn't need a default `emptyMap()` and external callers can ignore the `params` arg after we make it optional.

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsService.kt
```diff
@@ -40,29 +43,25 @@ class AnalyticsService internal constructor(
                 CoroutineScope(dispatcher)
             )
 
-    fun sendAnalyticsEvent(name: String, orderId: String? = null, buttonType: String? = null) {
+    fun sendAnalyticsEvent(name: String, params: Map<String, String> = emptyMap()) {
```
sshropshire: nit: same comment here, params could technically be optional.

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsServiceWrapper.kt
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Marker interface for analytics wrapper classes. Enforces that every wrapper
+ * holds an [AnalyticsService] instance, making the pattern consistent across
+ * [com.paypal.android.cardpayments.analytics.CardAnalytics],
+ * [com.paypal.android.paypalwebpayments.analytics.PayPalWebAnalytics], and
+ * [com.paypal.android.paymentbuttons.analytics.PaymentButtonAnalytics].
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+interface AnalyticsServiceWrapper {
```
sshropshire: q: What's the purpose of this interface?

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
```diff
@@ -18,24 +18,31 @@ import com.google.android.material.shape.MaterialShapeDrawable
...
 
-    internal val analyticsService: AnalyticsService =
-        AnalyticsService(context, CoreConfig(clientId = "N/A", environment = Environment.LIVE))
+    /**
+     * Analytics for button events. Populated if a [CoreConfig] was provided at construction.
+     * If null, no button events are sent.
+     * Internal so subclasses can fire INITIALIZED in their own init blocks.
+     */
+    internal val analytics: PaymentButtonAnalytics? =
+        coreConfig?.let { PaymentButtonAnalytics(AnalyticsService(context, it)) }
```
sshropshire: qq: `CoreConfig` may be `null` here and we won't get analytics. I feel like merchant's probably won't be instantiating these buttons programmatically (they'll probably be inflated via XML). Would merchants have to set `CoreConfig` programatically after the view is inflated?

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
```diff
@@ -196,6 +203,7 @@ abstract class PaymentButton<C : PaymentButtonColor> @JvmOverloads constructor(
 
         orientation = HORIZONTAL
         gravity = Gravity.CENTER
+        isClickable = true
```
sshropshire: Random question: I haven't reviewed this button code in a while (I believe it came from Native XO originally). Is this to resolve a bug we currently have? How did this change come about.

PaymentButtons/src/test/java/com/paypal/android/paymentbuttons/analytics/PaymentButtonAnalyticsTest.kt
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Test
+    fun `notify sends TAPPED event with correct button type param`() {
+        sut.notify(PaymentButtonEvent.TAPPED, "Pay Later")
+
+        verify(exactly = 1) {
+            analyticsService.sendAnalyticsEvent(
+                "payment-button:tapped",
+                params = mapOf(AnalyticsParams.BUTTON_TYPE to "Pay Later")
```
sshropshire:
```suggestion
                params = mapOf(AnalyticsParams.BUTTON_TYPE to "pay_later")
```
random q: Would `snake_case` or `hyphen-case` work better here for analytics? I'm not sure if there are any rules around the formatting about the values we send.

PaymentButtons/src/test/java/com/paypal/android/paymentbuttons/PaymentButtonTest.kt
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @After
+    fun tearDown() {
+        activityController.destroy()
+    }
+
+    // Injects a mock PaymentButtonAnalytics into the internal 'analytics' field
+    private fun injectMockAnalytics(button: PaymentButton<*>) {
+        val field = PaymentButton::class.java.getDeclaredField("analytics")
```
sshropshire: Can we do this with a constructor instead of using reflection? It feels like a potential test code smell that we should at least discuss.

sshropshire:
@anibalb2500 we may need to figure out a branching strategy when @kgangineni gets back. This may belong in a feature branch (or at least go into `develop`).

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsServiceWrapper.kt
anibalb2500: It enforces a consistent pattern across all analytics wrappers in the SDK. Every wrapper (CardAnalytics, PayPalWebAnalytics, PaymentButtonAnalytics) must hold an AnalyticsService instance. 

Without the interface, that's just a convention with nothing enforcing it. It also makes the pattern self-documenting for anyone adding a new analytics wrapper in the future.

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
anibalb2500: Excellent point.

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsService.kt
anibalb2500: Done.

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
anibalb2500: I intentionally removed XML attribute support for CoreConfig after discussion in the Slack thread. 

Since Android Views are moving to maintenance mode, won't merchants be using compose?

I can re-add support for passing in clientId and env in XML attributes, if that's what's preferred Happy to do it.

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
anibalb2500: Good question. While testing the button analytics I found that the tapped event wasn't being logged. That's because this view inherits from LinearLayout, not Button so it's not clickable by default.

PaymentButtons/src/test/java/com/paypal/android/paymentbuttons/PaymentButtonTest.kt
anibalb2500: Done.

PaymentButtons/src/test/java/com/paypal/android/paymentbuttons/analytics/PaymentButtonAnalyticsTest.kt
anibalb2500: I'm not sure, but that's something we can discuss separately and address as a follow up. Android and iOS both need to have consistent naming here.

CardPayments/src/main/java/com/paypal/android/cardpayments/analytics/CardAnalytics.kt
anibalb2500: Good catch. I've gone ahead and updated the buildParams to accept both the parameter name and value. This way we only have one method.

349446840464 (commented):
Can I ask about the sandbox new resperatory analizer , as new SQL that you create? With data files set?

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsServiceWrapper.kt
sshropshire: I'm still confused about this "wrapper" concept. The genericness of the `Wrapper` suffix feels like a code smell–if it had a legitimate purpose it would arguably have a better name explaining the role that it takes. We also have no declarations in this PR of a variable of type `AnalyticsServiceWrapper` meaning this interface isn't being used.

CardPayments/src/main/java/com/paypal/android/cardpayments/analytics/CardAnalytics.kt
```diff
@@ -1,14 +1,27 @@
...
     fun notify(event: VaultEvent, setupTokenId: String?) {
-        analyticsService.sendAnalyticsEvent(event.value, setupTokenId)
+        analyticsService.sendAnalyticsEvent(
+            name = event.value,
+            params = buildParams(AnalyticsParams.SETUP_TOKEN_ID, setupTokenId)
+        )
     }
+
+    private fun buildParams(key: String, value: String?): Map<String, String>? =
+        value?.let { mapOf(key to it) }
```
sshropshire: I'm also starting to become critical of this method. Is there a reason we can't just create the maps inline? I'll add some suggestions to do so to see if it cleans things up.

CardPayments/src/main/java/com/paypal/android/cardpayments/analytics/CardAnalytics.kt
```diff
@@ -1,14 +1,27 @@
...
-internal class CardAnalytics(private val analyticsService: AnalyticsService) {
+internal class CardAnalytics(
+    override val analyticsService: AnalyticsService
+) : AnalyticsServiceWrapper {
 
     fun notify(event: ApproveOrderEvent, orderId: String?) {
-        analyticsService.sendAnalyticsEvent(event.value, orderId)
+        analyticsService.sendAnalyticsEvent(
+            name = event.value,
+            params = buildParams(AnalyticsParams.ORDER_ID, orderId)
```
sshropshire:
```suggestion
            params = mapOf(AnalyticsParams.ORDER_ID to orderId)
```
This feels a lot simpler and more straightforward. The `buildParams` method only creates maps with a single key/value pair, which is very limiting. Using `mapOf` here gives us a lot more flexibility.

CardPayments/src/main/java/com/paypal/android/cardpayments/analytics/CardAnalytics.kt
```diff
@@ -1,14 +1,27 @@
...
+        )
     }
 
     fun notify(event: VaultEvent, setupTokenId: String?) {
-        analyticsService.sendAnalyticsEvent(event.value, setupTokenId)
+        analyticsService.sendAnalyticsEvent(
+            name = event.value,
+            params = buildParams(AnalyticsParams.SETUP_TOKEN_ID, setupTokenId)
```
sshropshire:
```suggestion
            params = mapOf(AnalyticsParams.SETUP_TOKEN_ID to setupTokenId)
```
+1

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
sshropshire: We'll still need support for XML since technically merchants will be able to use these views in XML layouts by virtual of them being Android Views. If we want to avoid this, we will have to author pure Composables (which isn't a bad idea, the buttons have a considerable amount of technical debt and there are new button designs in the wild that we have yet to adopt).

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
sshropshire: Yeah there may be some accessibility issues with our current button implementation I'm realizing now.

PaymentButtons/src/test/java/com/paypal/android/paymentbuttons/analytics/PaymentButtonAnalyticsTest.kt
sshropshire: Yeah we'll want these values to align across both iOS and Android (and possibly JS where possible).

PaymentButtons/src/test/java/com/paypal/android/paymentbuttons/PaymentButtonTest.kt
```diff
@@ -0,0 +1,101 @@
...
+            mockAnalyticsService.sendAnalyticsEvent(
+                PaymentButtonEvent.INITIALIZED.value,
+                params = any()
+            )
+        }
+    }
+
+    @Test
+    fun `no analytics event fires when no CoreConfig is provided`() {
+        val button = PayPalButton(context) // analytics is null — no CoreConfig
```
sshropshire: Just curious: is this a deal breaker for the analytics team? What happens if the initial button press cannot be captured?

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
...
+    across all SDK modules.
+* PaymentButtons
+  * Fix `payment-button:initialized` and `payment-button:tapped` events being sent to live
+    FPTI with `clientId = "N/A"`. `PaymentButton` now accepts an optional `CoreConfig`
+    constructor parameter — analytics fires with real credentials if provided, and is
+    silently suppressed if not (e.g. XML inflation).
+  * Add `PaymentButtonEvent` enum with typed `INITIALIZED` and `TAPPED` constants,
+    consistent with `CheckoutEvent`/`VaultEvent` in other modules.
+  * Add `PaymentButtonAnalytics` wrapper class, consistent with `CardAnalytics` and
+    `PayPalWebAnalytics` in other modules.
```
sshropshire:
```suggestion
```
Nothing in this PR should be merchant facing.

CardPayments/src/main/java/com/paypal/android/cardpayments/analytics/CardAnalytics.kt
anibalb2500: Latest changes remove the need for map.

CardPayments/src/main/java/com/paypal/android/cardpayments/analytics/CardAnalytics.kt
anibalb2500: No longer applicable.

CardPayments/src/main/java/com/paypal/android/cardpayments/analytics/CardAnalytics.kt
anibalb2500: No longer applicable.

PaymentButtons/src/test/java/com/paypal/android/paymentbuttons/PaymentButtonTest.kt
anibalb2500: We won't have a complete picture of the time line between button press and api calls.

anibalb2500:
@sshropshire I've rebased the PR and refactored the event params. Please take a look when you get a chance.

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsServiceWrapper.kt
anibalb2500: Removing the wrapper.

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
anibalb2500: Added analytics support for PaymentButtons added via XML

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
```diff
@@ -5,7 +5,5 @@ internal data class AnalyticsEventData(
     val environment: String,
     val eventName: String,
     val timestamp: Long,
-    val orderId: String?,
-    val buttonType: String? = null,
-    val appSwitchEnabled: Boolean
+    val params: AnalyticsEventParams = AnalyticsEventParams()
```
sshropshire: q: why not make this nullable? It feels more semantically correct if the value is absent when there are no additional analytics params to track. Do you agree?

Demo/src/main/res/layout/pay_later_button_test_layout.xml
```diff
@@ -23,30 +23,35 @@
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:layout_margin="8dp"
+            app:paypal_client_id="@string/paypal_demo_client_id"
```
sshropshire: qq: What happens if this isn't set?

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
```diff
@@ -225,20 +234,29 @@ abstract class PaymentButton<C : PaymentButtonColor> @JvmOverloads constructor(
...
-            analyticsService.sendAnalyticsEvent(
-                "payment-button:tapped",
-                orderId = null,
-                buttonType = fundingType.buttonType
+    private fun initAnalyticsFromXML(typedArray: TypedArray) {
+        val clientId = typedArray.getString(R.styleable.PaymentButton_paypal_client_id)
+        if (!clientId.isNullOrBlank()) {
+            val environmentValue = typedArray.getInt(
+                R.styleable.PaymentButton_paypal_environment,
+                0 // default to sandbox
```
sshropshire: q: What does zero mean in this context?

sshropshire: I see the comment, but why can't we use raw enum values or nullables in this scenario. It's confusing to use a magic number here.

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
anibalb2500: Good suggestion. I agree.

Demo/src/main/res/layout/pay_later_button_test_layout.xml
anibalb2500: Analytics in the PaymentButton simply doesn't get set. Click listener still works, so functionality that merchant passes still gets executed. 

This video shows me clicking the white PayPal button without the `paypal_client_id` set, and then clicking the black PayPal button. On the FPTI view, you'll see once we the search completes we only have one event. (_see time stamp on event update_).
 

https://github.com/user-attachments/assets/bf93d590-d2f5-441e-a65a-b76138c88464

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
anibalb2500: Great call out. Addressing.

Demo/src/main/res/layout/pay_later_button_test_layout.xml
sshropshire: Ok yeah. Basically it's just a no-op that makes sense

PaymentButtons/src/main/java/com/paypal/android/paymentbuttons/PaymentButton.kt
anibalb2500: The best solution I found for this was defining a local constant above with the name 'sandboxEnvironmentAttr'. I couldn't reference the raw enum values directly, as attrs.xml doesn't allow for that, and workarounds would require additional work.

sshropshire: Actually there is a [hasValue()](https://developer.android.com/reference/android/content/res/TypedArray#getInt(int,%20int)) method that could be useful here. Honestly the root cause is Android's terrible API. But at least with `hasValue()` we know that the fallback value will always be discarded in the successful branching case (we could even put that in a comment to make it clear and explicit).

anibalb2500: Sorry @sshropshire. Not sure if we're on the same page. We need a way to map the value from `PaymentButtons/src/main/res/values/attrs.xml`, using `hasValue` wouldn't help us out here unless we plan on passing in `Environment.Sandox.ordinal` and make sure that the the enums are ordered identically. But that would introduce some coupling between `PaymentButton` and `CorePayments`. 

Please let me know your thoughts, or if I'm missing something.

anibalb2500: No longer applicable.

PR #453, opened 2026-07-02
Custom environment support

anibalb2500:

### Summary of changes

- Adds `Environment.CUSTOM` as a debug-only enum entry (excluded from release builds via source sets) that allows the SDK to target a custom REST and GraphQL base URL at runtime
- Adds a Settings screen to the Demo app (debug builds only) where developers can configure the SDK REST URL, SDK GraphQL URL, and an optional Client ID override; settings are persisted via SharedPreferences
- Fixes `PayPalWebCheckoutClient.baseUrl` to derive from `environment.graphQLEndpoint` so custom environments correctly route web checkout instead of falling through to the live URL

### Notes
- Merchants won't see any changes. 

### Checklist
- [ ] Added a changelog entry


PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -532,10 +531,7 @@ class PayPalWebCheckoutClient internal constructor(
     }
 
     private val baseUrl: String
-        get() = when (coreConfig.environment) {
```
kgangineni, line 535: why this change is needed?

FraudProtection/src/main/java/com/paypal/android/fraudprotection/CoreConfigMagnesExt.kt
```diff
@@ -6,5 +6,5 @@ import lib.android.paypal.com.magnessdk.Environment
 internal val CoreConfig.magnesEnvironment: Environment
     get() = when (environment) {
         com.paypal.android.corepayments.Environment.LIVE -> Environment.LIVE
-        com.paypal.android.corepayments.Environment.SANDBOX -> Environment.SANDBOX
```
kgangineni, line 9: this is functionality change
consider debug/release source sets or any better way

Demo/src/main/java/com/paypal/android/customenvironment/CustomEnvironmentConfig.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import com.paypal.android.corepayments.CoreConfig
+import com.paypal.android.corepayments.Environment
+
+/**
+ * Holds the SDK URLs for a custom environment.
+ *
+ * All values are blank by default; [isConfigured] is true only when both URL fields are filled in.
+ */
+data class CustomEnvironmentConfig(
```
kgangineni: not sure if we need this complexity in demo app
consider making Custom button only visible in debug variant
if possible avoid debug/release source sets in demo app
at the moment I am not able to run demo app in release variant, it shows a compiler error

FraudProtection/src/main/java/com/paypal/android/fraudprotection/CoreConfigMagnesExt.kt
anibalb2500: Done.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
anibalb2500: The original when expression hardcoded the URLs and had an implicit `else -> "https://www.sandbox.paypal.com/"` fallback. Meaning Environment.CUSTOM would silently be routed to the sandbox PayPal checkout URL regardless of what environment the user configuredd. 

This change gets baseUrl from graphQLEndpoint directly, so whatever environment the user configured (including CUSTOM) is used to build the correct checkout URL. 

Without this fix, a developer pointing the SDK at a custom environment would still land on sandbox's checkout page.

Does this make sense?

Demo/src/main/java/com/paypal/android/customenvironment/CustomEnvironmentConfig.kt
anibalb2500: Reduced the number of debug/release source sets in demo app, and fixed issue with release variant not compiling.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -531,11 +530,7 @@ class PayPalWebCheckoutClient internal constructor(
             .build()
     }
 
-    private val baseUrl: String
-        get() = when (coreConfig.environment) {
-            Environment.LIVE -> "https://paypal.com/"
-            Environment.SANDBOX -> "https://sandbox.paypal.com/"
-    }
+    private val baseUrl: String = coreConfig.environment.graphQLEndpoint
```
kgangineni: given it's simplified now, do we need extra private val for this it can be accessed from `coreConfig.environment.graphQLEndpoint`

kgangineni:
@anibalb2500 thanks for all the updates,
one more item, there's no way to enter url for merchant server now, which is always pointing to sandbox, this url resides in demo app, used for creating orders, setup token etc, provide a way to enter this value in UI by default in custom

Demo/src/main/java/com/paypal/android/customenvironment/DemoEnvironmentSettings.kt
```diff
@@ -0,0 +1,34 @@
...
+     * Optional merchant server base URL override. When blank, falls back to the default
+     * merchant server URL.
+     */
+    val customMerchantBaseUrl: String = "",
+) {
+    /**
+     * True when the config is ready to use. Only false when [selectedEnvironment] is CUSTOM
+     * and either URL field is blank.
+     */
+    val isValidEnvironment: Boolean
```
kgangineni, line 30: include check for merchant server base url too

Demo/src/main/java/com/paypal/android/customenvironment/SettingsViewModel.kt
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+
+    /**
+     * Validates a URL field. Both http:// and https:// are accepted.
+     *
+     * When [optional] is true, a blank value is considered valid (field is not required).
+     * When [optional] is false, the field is required.
+     */
+    private fun validateUrl(url: String, optional: Boolean = false): String? {
```
kgangineni: can be simplified by using something like this
`private fun validateUrl(url: String, optional: Boolean = false): String? {
    if (url.isBlank()) return if (optional) null else "URL is required"
    if (url != url.trim() || url.contains(' ')) return "URL must not contain spaces"
    return try {
        val uri = URI(url)
        when {
            uri.scheme !in listOf("http", "https") -> "URL must start with http:// or https://"
            uri.host.isNullOrBlank() -> "Enter a valid URL (e.g. https://api.example.com)"
            else -> null
        }
    } catch (e: URISyntaxException) {
        "Enter a valid URL (e.g. https://api.example.com)"
    }
}`

or use `okhttp.HttpUrl`

kgangineni (approved):
few comments
other than that LGTM

kgangineni:
@anibalb2500 plz check with local server by entering fields in custom environment, before merging

PR #459, opened 2026-07-16
Add shopper session analytics events

anibalb2500:

### Summary of changes
 - Added new shopper-session-aware analytics events for `CheckoutEvent`/`VaultEvent`: `SESSION_NOT_STARTED`, `APP_SWITCH_STARTED/SUCCEEDED/FAILED/CANCELED`, `BROWSER_PRESENTATION_STARTED/SUCCEEDED/FAILED/CANCELED`, and `HANDLE_RETURN_STARTED/SUCCEEDED/FAILED`, replacing the old `AUTH_CHALLENGE_PRESENTATION_*` events.
 - Added a new `CreatePayPalSessionEvent` (`STARTED`/`SUCCEEDED`/`FAILED`) fired from `createPayPalSession()` for both checkout and vault flows.
 - Threaded `shopper_session_id`, `app_switch_url`, `error_description`, `start_time`, `is_cached_session`, and `is_vault_request` through `PayPalWebAnalytics` → `AnalyticsService` → `AnalyticsEventData` → `TrackingEventsAPI` so they're included in the FPTI payload.
 - Updated `PayPalWebCheckoutClient`'s `start()`/`vault()`/`finishStart()`/`finishVault()` flows to fire the new events with the resolved `shopperSessionId` at each stage.
 - `is_cached_session` is now derived from whether the caller passed an `existingPayPalSessionId` into `createPayPalSession()`, matching how iOS derives it from `PayPalUserIdentity`.
 - Moved `paypalNativeAppInstalled`, `buyerEmailAddressMerchantPassed`, and `existingPayPalSessionId` off of `CreateShopperSessionWithAppSwitchEligibilityAPI.invoke()`'s parameter list and onto `CreateShopperSessionWithAppSwitchEligibilityParams`, updating all call sites and tests. Also removed the unused `merchantCountry` field.
 - Removed `analytics.notify()` calls from the deprecated/legacy `startAsync`, `vaultAsync`, and deprecated `start()`/`vault()` paths (these methods are being removed in a follow-up PR), and updated/removed the corresponding unit tests.
 - General lint cleanup (detekt `LongMethod`/`MaxLineLength` fixes via pure extract-method refactors, no behavior change).

### Checklist
 - [ ] Added a changelog entry (not needed — no merchant-facing changes)

- anibalb2500

### DIFF
--- a/CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
+++ b/CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
@@ -7,5 +7,15 @@ internal data class AnalyticsEventData(
     val timestamp: Long,
     val orderId: String?,
     val buttonType: String? = null,
-    val appSwitchEnabled: Boolean
+    val appSwitchEnabled: Boolean,
+    // fields below support the mobile SDK analytics events documented at
+    // https://paypal.atlassian.net/wiki/spaces/~7120203361479131b645799d3eacdd2de5b990/pages/2982842759
+    val shopperSessionId: String? = null,
+    val appSwitchUrl: String? = null,
+    val errorDescription: String? = null,
+    // captured right before firing a `:started` event and passed forward to the matching
+    // `:succeeded`/`:failed` event so FPTI can derive latency as (t - start_time)
+    val startTime: Long? = null,
+    val isCachedSession: Boolean? = null,
+    val isVaultRequest: Boolean? = null
 )
--- a/CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
+++ b/CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
@@ -8,13 +8,9 @@ internal data class AnalyticsEventData(
     val orderId: String?,
     val buttonType: String? = null,
     val appSwitchEnabled: Boolean,
-    // fields below support the mobile SDK analytics events documented at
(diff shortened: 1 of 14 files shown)


CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
```diff
@@ -7,5 +7,11 @@ internal data class AnalyticsEventData(
     val timestamp: Long,
     val orderId: String?,
     val buttonType: String? = null,
-    val appSwitchEnabled: Boolean
+    val appSwitchEnabled: Boolean,
```
sshropshire: Still concerned that this isn't `appSwitchEnabled: Boolean?`. The more I look at these analytics properties, the more they begin to resemble a `Map<AnalyticsKey, Any>` data structure, except in this case if app switch enabled defaults to `false` a key/value pair for `appSwitchEnabled=false` will always be added to the analytics parameters dictionary whenever `sendEvent()` is called without specifying if app switch is enabled.

sshropshire: e.g. for `CardClient.vault()`, app switch enabled will be `false`, even though this feature has nothing to do with app switch. To me it would make more sense semantically for the `appSwitchEnabled` analytics param to be absent in that case.

sshropshire (approved):

CorePayments/src/main/java/com/paypal/android/corepayments/analytics/AnalyticsEventData.kt
anibalb2500: Addressed.

PR #463, opened 2026-07-20
adding should vault option

anibalb2500:

### Summary of changes

Added a "Store in Vault" option to the PayPal Checkout demo flow. A new reusable StoreInVaultOptionForm component wraps the existing EnumOptionList for the StoreInVaultOption enum, and is rendered in the Create Order step of PayPalCheckoutView as well as reused in CreateOrderWithVaultOptionForm. The selected state is held in PayPalUiState.shouldVaultOption (StoreInVaultOption enum) and bridged through PayPalCheckoutViewModel.shouldVault, which feeds the vault flag into createOrder().
Vaulting is now expressed on the PayPal payment source (matching paypal-ios). CreateOrderUseCase builds payment_source.paypal.attributes.vault instead of attaching the vault to the card payment source, so the vault applies to the PayPal Checkout as intended.
Extended the Vault model to send the full attribute set — storeInVault = "ON_SUCCESS", usageType = "MERCHANT", customerType = "CONSUMER" — mirroring the iOS demo's vault attributes.
Included the required experience_context on the PayPal payment source. PayPal's Orders v2 API returns 422 Unprocessable Entity for a PayPal payment source without return_url/cancel_url, so PayPalPaymentSource now carries a PayPalOrderExperienceContext (with nativeApp made optional for the non–app-switch vault case).
Corrected the token query-param logic in PayPalWebCheckoutClient.getLaunchUri() so the token is appended to the launch/fallback URL's encoded query directly (updated corresponding unit tests).

 ### Checklist

 - [ ] Added a changelog entry

-


PayPalWebPayments/src/test/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClientUnitTest.kt
```diff
@@ -1698,6 +1698,16 @@ class PayPalWebCheckoutClientUnitTest {
         shopperSessionConfig = ShopperSessionConfig("fake-session-id", "")
     )
 
+    private fun placeholderTokenUrl(
+        baseUrl: String,
+        tokenType: String,
+        tokenKey: String,
+        extraQuery: String? = null,
+    ): String {
+        val extra = extraQuery?.let { "$it&" }.orEmpty()
```
sshropshire: This line made it back in somehow.

sshropshire (approved):

sshropshire (approved):

PR #464, opened 2026-07-20
Fix Token Query Param

anibalb2500:

### Summary of changes

 - Fixing a bug where we append the order id or setup token to the universal link incorrectly. 

 ### Checklist

 - [ ] Added a changelog entry

-


PayPalWebPayments/src/test/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClientUnitTest.kt
```diff
@@ -1698,6 +1698,16 @@ class PayPalWebCheckoutClientUnitTest {
         shopperSessionConfig = ShopperSessionConfig("fake-session-id", "")
     )
 
+    private fun placeholderTokenUrl(
+        baseUrl: String,
+        tokenType: String,
+        tokenKey: String,
+        extraQuery: String? = null,
+    ): String {
+        val extra = extraQuery?.let { "$it&" }.orEmpty()
```
sshropshire: Can we remove this line since it never gets used?

sshropshire (approved):

sshropshire (approved):

PR #465, opened 2026-07-21
fix: detect app-switch cancel without opType

ayer-ribeiro:

### Summary of changes

`parseWebCheckoutSuccessResult` only detected a checkout cancel via `opType == "cancel"`. The app-switch flow never sets opType, so a real cancel fell into the malformed-result branch and surfaced _"Result did not contain the expected data. Payer ID or Order ID is null."_ instead of a cancel state.

Detect the outcome from PayerID presence (the approval signal) instead:

<table>
  <tr>
    <th>Case</th>
    <th>Result</th>
  </tr>
  <tr>
    <td>opType == "cancel"</td>
    <td>Canceled (unchanged; covers cancels that carry a PayerID)</td>
  </tr>
  <tr>
    <td>orderId + PayerID present</td>
    <td>Success</td>
  </tr>
  <tr>
    <td>no order context</td>
    <td>malformed Failure</td>
  </tr>
  <tr>
    <td>order present, no PayerID</td>
    <td>Canceled (buyer cancelled before approval)</td>
  </tr>
</table> 

This is URL-independent, so it works whether the merchant's returnAppUrl and cancelAppUrl are distinct or identical. 

### Recording

https://github.com/user-attachments/assets/7dc80f5f-a039-45f9-8045-aad198fab13a

 ### Checklist

- @ayer-ribeiro


kgangineni (approved):
LGTM

PR #466, opened 2026-07-21
Add GetReturnLinkTypeUseCase to select return-link strategy at launch

ayer-ribeiro:

### Summary of changes

The v3 shopper-session checkout/vault flow always launched with a hardcoded App Link return strategy and never fell back to the custom URL scheme, even though a fallback scheme is captured from the merchant. App Links can silently fail to route back to the merchant app (unverified Digital Asset Links, the shopper disabling "Open supported links", or a non-App-Links-compatible default browser), leaving shoppers stranded.

Introduce GetReturnLinkTypeUseCase (backed by GetDefaultAppUseCase and GetAppLinksCompatibleBrowserUseCase) to decide at launch time whether an App Link return will actually route on the device, and wire it into PayPalWebCheckoutClient so the SDK picks between AppLink(returnAppUrl) and CustomUrlScheme(fallbackSchemeUrl). Add a link_type analytics param (applink/deeplink) reporting which strategy was used.

- New use cases + LinkType enum in CorePayments (manual DI, LIBRARY_GROUP)
- resolveReturnToAppStrategy() in PayPalWebCheckoutClient for checkout + vault
- link_type threaded through the analytics pipeline to the FPTI event
- Unit tests for the use cases, analytics forwarding/serialization, and both strategy branches at the client level

### Tests

image

### Recordings

https://github.com/user-attachments/assets/86013f09-199a-4911-b08f-b5eda95092a2

https://github.com/user-attachments/assets/b844d850-2ec2-41f9-ac12-0586d70d3ef8

 ### Checklist

- @ayer-ribeiro


CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
```diff
@@ -0,0 +1,89 @@
...
+     */
+    operator fun invoke(
+        appLinkReturnUrl: String?,
+        fallbackSchemeUrl: String?,
+        checkoutUri: Uri = DEFAULT_CHECKOUT_URI,
+    ): ReturnToAppStrategy {
+        // Without a custom URL scheme there is nothing to deep-link to, so App Link is the only
+        // possible return type. Short-circuit before doing any package-manager probing.
+        if (fallbackSchemeUrl.isNullOrBlank()) {
+            return ReturnToAppStrategy.AppLink(appLinkReturnUrl.orEmpty())
```
sshropshire: question: What happens when both `fallbackSchemeUrl` and `appLinkReturnUrl` are `null`?

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * True when the merchant app is the default handler for its own App Link return URI. Returns
+     * false when the shopper has unchecked "Open supported links" for the merchant app in Android
+     * settings, or when App Link (Digital Asset Links) verification isn't configured.
+     */
+    private fun isMerchantDefaultHandlerForReturnUrl(appLinkReturnUrl: String?): Boolean {
+        val appLinkReturnUri = appLinkReturnUrl?.takeIf { it.isNotBlank() }?.toUri()
+        return appLinkReturnUri != null && applicationContext.packageName == getDefaultApp(appLinkReturnUri)
+    }
```
sshropshire: question: why is this necessary? The Android OS is responsible for routing deep links it feels like we're doing extra work here.

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    private companion object {
+        // A neutral https URL used only to probe the device's default browser: its default handler
+        // reveals whether an App Link return would route. This must not be an app-link-verified path
+        // for any PayPal-owned app, or it would resolve to that app instead of a browser and defeat
+        // the probe.
+        private val DEFAULT_CHECKOUT_URI = "https://www.paypal.com/checkout".toUri()
```
sshropshire: question: Is there risk of this ever changing?

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/HasAppLinksCompatibleBrowserUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+        APP_LINK_COMPATIBLE_BROWSERS.any { getDefaultAppUseCase(browserUri)?.contains(it) == true }
+
+    private companion object {
+        private val APP_LINK_COMPATIBLE_BROWSERS = listOf(
+            "com.android.chrome",
+            "com.brave.browser",
+            "com.sec.android.app.sbrowser",
+            "org.mozilla.firefox",
+            "com.microsoft.emmx",
+        )
```
sshropshire: question: Is there an explicit reason why we're excluding other browsers?

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/HasAppLinksCompatibleBrowserUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @suppress
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class HasAppLinksCompatibleBrowserUseCase(
+    private val getDefaultAppUseCase: GetDefaultAppUseCase,
+) {
+
+    operator fun invoke(browserUri: Uri?): Boolean =
+        APP_LINK_COMPATIBLE_BROWSERS.any { getDefaultAppUseCase(browserUri)?.contains(it) == true }
```
sshropshire: NOTE: `getDefaultAppUseCase()` is called 5 times here.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -1145,6 +1177,22 @@ class PayPalWebCheckoutClient internal constructor(
     }
 
     // endregion
+
+    private companion object {
+        /**
+         * Builds a [GetReturnToAppStrategyUseCase] wired with its collaborators from an application
+         * [Context], following the SDK's manual dependency-injection convention.
+         */
+        private fun buildReturnToAppStrategyUseCase(applicationContext: Context): GetReturnToAppStrategyUseCase {
```
sshropshire: Would it be better to move this DI logic into the constructor of `GetReturnToAppStrategyUseCase`?

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
anibalb2500: This checks whether the app is currently Android OS's resolved handler for the apps own App Link return URL. If it isn't (i.e. "Open supported links" is disabled) the return URL will open as a webpage instead of routing back to the app. 

We need to know that before launch, since there's no retry after the fact, so we can fall back to the custom URL scheme instead.

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
ayer-ribeiro: In the scenario above:

The SDK currently uses the AppLink type and passes an empty string: https://github.com/paypal/paypal-android/pull/466/changes/BASE..5a72a3dad472bb1eb3099b18c432b08b8ee32cf6#diff-f882f26c7b2c9957c47f177e3c548650493f86937bbe0e9c7d9f988dafb48232

With the changes in this PR, the flow would return early at the check that also returns AppLink with an empty string.

This PR does not change the existing behavior, but we may want to throw an error to make the merchant aware that the integration is not correct, or consider other error handling. We can discuss this internally.

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
ayer-ribeiro: This URL is not an actual checkout URL. It is only used to verify whether there is a browser on the device that the OS can use to open the app link.

I used a similar approach to the existing Braintree integration: https://github.com/braintree/braintree_android/blob/main/BraintreeCore/src/main/java/com/braintreepayments/api/core/usecase/GetReturnLinkTypeUseCase.kt#L29C52-L29C87

Do you think we should try a different approach?

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
ayer-ribeiro: yeah, I agree, I'll work on it

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
sshropshire: Yeah definitely if we missed something and we aren't throwing an error.

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
sshropshire: Won't Chrome Custom Tabs fallback to a webpage anyways? I'm trying to determine why we need to perform the check, the only use case I can think of is maybe for analytics if we wanted to log which path was taken.

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
sshropshire: Looking at https://www.paypal.com/.well-known/assetlinks.json, I originally thought the `/checkout` path might be qualified separately, but it seems like anything hitting the PayPal domain. Does that make it an `app-link-verified` path? Looking at the comment above it says we need a neutral URL.

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/HasAppLinksCompatibleBrowserUseCase.kt
sshropshire: This isn't a deal breaker, but we could technically call it once.

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/HasAppLinksCompatibleBrowserUseCase.kt
sshropshire: This seems overly restrictive to me. I'm not sure what makes a browser app link compatible.

CorePayments/src/main/java/com/paypal/android/corepayments/LinkType.kt
```diff
@@ -0,0 +1,16 @@
...
+ * The return-link type reported to analytics via the `link_type` param.
+ *
+ * Note: exposed for internal PayPal SDK use only. Not covered by Semantic Versioning.
+ *
+ * @suppress
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+enum class LinkType(val stringValue: String) {
+    APP_LINK("applink"),
+    DEEP_LINK("deeplink"),
```
sshropshire, line 15:
```suggestion
    APP_LINK("app_link"),
    DEEP_LINK("deep_link"),
```
To better match the name?

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
anibalb2500: The use case is actually determining which urls to pass in as return urls, whether in app link form or deep link form. If we pass in app link as the return url, and then app switch into PayPal, once PayPal tries to redirect to merchant and merchant app doesn't support app links, the intent launched by PayPal will open browser.

sshropshire: Oh got it. I understand now this also covers the scenario where merchant supports app links but user has disabled them on their personal device.

CorePayments/src/main/java/com/paypal/android/corepayments/LinkType.kt
anibalb2500: Our analytics require this to be applink and deeplink

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/HasAppLinksCompatibleBrowserUseCase.kt
anibalb2500: It's intentionally an opt-in list built from actual testing that was done for the braintree_android. These browsers and pre-tested and confirmed to be compatible with android app links. 

Some browsers intercept and keep the redirect inside their own in-app browser session instead of letting the OS route it back to the merchant app, which breaks the return-to-app flow.

sshropshire: Got it so basically we want to force it to fallback to custom url scheme in those cases to be sure.

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
anibalb2500: Added new error `RETURN_TO_APP_URL_CONFIG_MISSING`. Created iOS story to check if we need to add it there.

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
anibalb2500: Good catch. 

We reused this from braintree_android, which can reuse that URL because its equivalent check ORs "resolves to PayPal/Venmo app" directly into the same function. 

Ours splits that into a separate check (canResolvePayPalAppSwitch, which also enforces a minimum app version Braintree's doesn't), so this probe's only job is reading the actual fallback browser. 

Reusing paypal.com here would let the PayPal app's own domain claim mask that answer. 

Updated DEFAULT_CHECKOUT_URI to https://example.com/checkout, a domain no app can claim, and renamed to DEFAULT_BROWSER_PROBE_URI

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
```diff
@@ -0,0 +1,114 @@
...
+     *   or blank, since there would be no usable return route back to the merchant app.
+     */
+    operator fun invoke(
+        appLinkReturnUrl: String?,
+        fallbackSchemeUrl: String?,
+        checkoutUri: Uri = DEFAULT_BROWSER_PROBE_URI,
+    ): ReturnToAppStrategy {
+        require(!appLinkReturnUrl.isNullOrBlank() || !fallbackSchemeUrl.isNullOrBlank()) {
+            "Either appLinkReturnUrl or fallbackSchemeUrl must be provided to return to the merchant app."
+        }
```
sshropshire: q: Do these throw?

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
sshropshire: Does that error get thrown somewhere?

CorePayments/src/main/java/com/paypal/android/corepayments/usecase/GetReturnToAppStrategyUseCase.kt
anibalb2500: Yes, they throw `IllegalArgumentException` but it's unreachable in production as we fail before calling this if we find that appLinkReturnUrl or fallbackSchemeUrl are not provided. 

Kept it in, in case we ever reuse this use case.

sshropshire (approved):

CorePayments/src/main/AndroidManifest.xml
```diff
@@ -6,6 +6,12 @@
 
     <queries>
         <package android:name="com.paypal.android.p2pmobile" />
+
+        <intent>
```
sshropshire, line 12: nit: I'd add a comment here explaining why it's necessary.

sshropshire (approved):

PR #468, opened 2026-07-21
Fallback to PatchCCO If Create CreateShopperSessionWithAppSwitchEligibilityAPI fails

anibalb2500:

### Summary of changes

 - Adding logic to call the PatchCCOWithAppSwitchEligibility if CreateShopperSessionWithAppSwitchEligibilityAPI fails, in a second attempt to app switch user. 
 - Updated demo app to add return and cancel urls in order creation 
 - Removed unit tests that test methods we will be deprecating, as they'd fail due to changes to the `getLaunchUri` in `PayPalWebCheckoutClient`

 ### Checklist

 - [ ] Added a changelog entry

-


anibalb2500:
Closing. We should simply throw `SESSION_NOT_CREATED` error instead of falling back

PR #473, opened 2026-07-23
Appending platform query param.

anibalb2500:

### Summary of changes

 - Adding `platform` query param to the browser switch url, to be aligned with iOS and future proof the experience.
 - Organizing methods 

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
+++ b/PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
@@ -701,6 +701,17 @@ class PayPalWebCheckoutClient internal constructor(
     private fun canAttemptPayPalAppSwitch(): Boolean =
         deviceInspector.isPayPalInstalled && deviceInspector.canResolvePayPalAppSwitch()
 
+    private fun CreateShopperSessionWithAppSwitchEligibilityResponse.getLaunchUri(token: String): Uri {
+        return if (appSwitchEnabled) {
+            redirectUrl.toUri()
+        } else {
+            checkoutFallbackUrl.toUri()
+        }.appendTokenQueryParam(token)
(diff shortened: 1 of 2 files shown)


PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -759,6 +768,18 @@ class PayPalWebCheckoutClient internal constructor(
             .build()
     }
 
+    /**
+     * Appends platform as a query param to browser switch urls, to be aligned with the iOS SDK.
+     * The iOS sdk appends this to fix an issue with redirecting back to merchant app.
+     */
+    private fun Uri.appendPlatformQueryParam(): Uri {
```
kgangineni: if intention is to have cascading functions to append Query params, use `apply{}` block rather than returning this in all these functions

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -701,62 +701,65 @@ class PayPalWebCheckoutClient internal constructor(
     private fun canAttemptPayPalAppSwitch(): Boolean =
         deviceInspector.isPayPalInstalled && deviceInspector.canResolvePayPalAppSwitch()
 
-    /**
-     * Drops a trailing '&' (so appendQueryParameter doesn't produce a double separator) and
-     * appends the given token as a query param.
-     */
-    private fun Uri.appendTokenQueryParam(token: String): Uri {
+    private fun CreateShopperSessionWithAppSwitchEligibilityResponse.getLaunchUri(token: String): Uri {
```
sshropshire, line 704: question: would it be better if we move this method to  `CreateShopperSessionWithAppSwitchEligibilityResponse` ?

PR #476, opened 2026-07-28
Set payment method selected on create order

anibalb2500:

### Summary of changes

 - Setting payment_method_selected in createOrder call of Demo App. 
 - Updating input in PayPal Checkout to set price of order, to correctly test Pay Later and PayPal Credit
 - Removing `funding_source` query param to correctly handle PayLater and PayPal Credit flows. 

 ### Checklist

 - [ ] Added a changelog entry

-


Demo/src/main/java/com/paypal/android/models/OrderRequest.kt
```diff
@@ -7,4 +7,7 @@ import kotlinx.serialization.Serializable
 data class OrderRequest(
     val intent: OrderIntent,
     val shouldVaultOnSuccess: Boolean,
+    val amount: String = "10.99",
+    val paymentMethodSelected: String = "PAYPAL",
+    val userAction: String = "PAY_NOW",
```
sshropshire, line 12: Should these be enums?

anibalb2500: Done.

kgangineni (approved):

PR #477, opened 2026-07-28
App Switch Analytics Corrections

anibalb2500:

### Summary of changes

 - Removing `vault-wo-purchase` events and using checkout events for all flows. 
 - Removing app switch and browser switch cancel events.

 ### Checklist

 - [ ] Added a changelog entry

-


PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/analytics/AppSwitchAnalyticsEventParams.kt
```diff
@@ -9,16 +9,11 @@ import com.paypal.android.corepayments.model.CreateShopperSessionWithAppSwitchEl
...
+     * [PayPalWebCheckoutClient.vault]). Which one it is can be determined from [isVault].
      */
-    val checkoutOrderId: String? = null,
-
-    /**
-     * The setup token id being approved. Set by [PayPalWebCheckoutClient.vault] for a vault flow.
-     * Mutually exclusive with [checkoutOrderId].
-     */
-    val vaultSetupTokenId: String? = null,
+    val orderId: String? = null,
```
sshropshire:
```suggestion
    val orderIdOrSetupTokenId: String? = null,
```
Would this name be more appropriate based on the docstring?

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
kgangineni: not part of this PR but related, make this inline or `initAnalyticsParams()`
`appSwitchAnalyticsEventParams` to `analyticsEventParams`
I see resetting as well for analytics why is that function needed?

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -157,15 +156,15 @@ class PayPalWebCheckoutClient internal constructor(
             return
         }
         appSwitchAnalyticsEventParams = appSwitchAnalyticsEventParams.copy(
-            checkoutOrderId = orderId,
+            orderId = orderId,
```
kgangineni: these are not just appSwitchAnalytics, `analyticsEventParams`? to be more generic

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/analytics/PayPalWebAnalytics.kt
```diff
@@ -5,21 +5,11 @@ import com.paypal.android.corepayments.analytics.AnalyticsService
 
 internal class PayPalWebAnalytics(private val analyticsService: AnalyticsService) {
 
-    fun notify(event: CheckoutEvent, params: AppSwitchAnalyticsEventParams, errorDescription: String? = null) {
+    fun notify(event: PayPalEvent, params: AppSwitchAnalyticsEventParams, errorDescription: String? = null) {
```
sshropshire: 🔥

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
anibalb2500: Good catch. Renamed.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/analytics/AppSwitchAnalyticsEventParams.kt
anibalb2500: Good idea. Renamed.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
anibalb2500: Good idea, implemented. 

Also, I removed the `resetAppSwitchAnalyticsEventParams` method. I added this to reset the `analyticsEventParams` in scenarios where old `analyticsEventParams` data could be used. 

For instance, if `start()`/`vault()` was called without ever calling createPayPalSession() first - no merchantId/bnCode/clientId would be included in analytics or  a second start()/vault() call after a previous session was already consumed, without a fresh createPayPalSession()​ where we'd see analytics param from the first transaction. 

I've added unit tests to cover these scenarios.

anibalb2500: Also, you'll see this in the beginning of `start()` and `vault()` 
```
        if (deferred == null) {
            initAnalyticsEventParams()
        }
        analyticsEventParams = analyticsEventParams.copy(
            orderIdOrSetupTokenId = orderId,
            isVault = false,
        )
        if (deferred == null) {
            notifyUserPerceivedLatencyError(LatencyFlow.CHECKOUT, startTime)
            notifyCheckoutSessionNotStarted(callback)
            return
        }
```
This is so we can keep the `orderIdOrSetupTokenId` and `isVault` analytics params in the events logged by `notifyUserPerceivedLatencyError` and `notifyCheckoutSessionNotStarted`

kgangineni (approved):

PR #450, opened 2026-06-24
(MAJOR): remove all deprecated APIs

kgangineni:
### Summary
Removes all remaining deprecated public APIs across `FraudProtection`, `CardPayments`, and `PayPalWebPayments` modules.

## New

None

## Modified

None

## Breaking Changes

- [ ] No breaking changes
- [x] Breaking changes (describe migration steps below)

## Migration:

- `PayPalDataCollector.collectDeviceData(context, clientMetadataId, additionalData)` → use `collectDeviceData(context, request)` with a `PayPalDataCollectorRequest`
- `CardClient.finishApproveOrder(intent, authState)` → use `finishApproveOrder(intent)`
- `CardClient.finishVault(intent, authState)` → use `finishVault(intent)`
- `PayPalWebCheckoutClient.finishStart(intent, authState)` → use `finishStart(intent)`
- `PayPalWebCheckoutClient.finishVault(intent, authState)` → use `finishVault(intent)`
- `CardPresentAuthChallengeResult.Success.authState` and `PayPalPresentAuthChallengeResult.Success.authState` → removed; auth state is now managed internally by the SDK
- `PayPalWebCheckoutClient.start(activity, request)` → use `start(activity, request, callback)`
- `PayPalWebCheckoutClient.vault(activity, request)` → use `vault(activity, request, callback)`
- `PayPalWebCheckoutClient(context, configuration, urlScheme)` → use `PayPalWebCheckoutClient(context, configuration)` and set `returnToAppStrategy` in `PayPalWebCheckoutRequest`/`PayPalWebVaultRequest` (e.g. `ReturnToAppStrategy.CustomUrlScheme(urlScheme)`)
- `PayPalWebVaultRequest.approveVaultHref` → removed; property was previously ignored

### Checklist

- [x] Added a changelog entry

PR #481, opened 2026-07-30
Clean up and removing Remove dead code

anibalb2500:

### Summary of changes
- Removes `PayPalWebCheckoutClient`'s private 4-arg `getLaunchUri(context, token, tokenType, fallbackUri)` helper — it was only reachable from the deprecated `startAsync`/`vaultAsync` methods removed in #450, so it's now dead code.
- Removes the `patchCCOWithAppSwitchEligibility` dependency from `PayPalWebCheckoutClient`, since its only caller was the dead `getLaunchUri` above.
- Deletes `PatchCCOWithAppSwitchEligibility` entirely (class + unit test) — with the constructor dependency gone, nothing else in the codebase calls it.
- Deletes `UpdateClientConfigAPI` (class + unit test) — its only remaining caller was `PatchCCOWithAppSwitchEligibility` (for the `Defaults.INTEGRATION_ARTIFACT`/`USER_EXPERIENCE_FLOW` constants), so it's dead now too.
- Deletes the now-unused Patch CCO / Update Client Config GraphQL model classes and their `.graphql` query resource files.
- Removes the corresponding dead mocks and four `@Ignore`d "patchCCO fallback" tests from `PayPalWebCheckoutClientUnitTest` that referenced this dependency; tightened two remaining test names/comments that referenced the removed fallback behavior.
- Regenerates `CorePayments.api` (BCV dump) to drop the removed internal classes' generated `kotlinx.serialization` ABI entries.
- Organized methods in `PayPalWebCheckoutClient` into detailed regions. 
- Condensed `launchCheckoutWithSession` and `launchVaultWithSession` methods into one `launchMethod`

### Checklist
- [] Added a changelog entry 
   - N/A (internal-only, `@RestrictTo(LIBRARY_GROUP)` classes — no public API surface changed)

- @anibalb2500


CorePayments/src/main/java/com/paypal/android/corepayments/UpdateClientConfigAPI.kt
```diff
@@ -1,132 +0,0 @@
...
-import androidx.annotation.RestrictTo
-import com.paypal.android.corepayments.graphql.GraphQLClient
-import com.paypal.android.corepayments.graphql.GraphQLRequest
-import com.paypal.android.corepayments.graphql.GraphQLResult
-import kotlinx.serialization.InternalSerializationApi
-import kotlinx.serialization.Serializable
-
-@OptIn(InternalSerializationApi::class)
-@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-class UpdateClientConfigAPI(
```
kgangineni, line 13: my bad on this, we're still using this in venmo feature, revert all related code for `UpdateClientConfigAPI`

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -258,10 +256,11 @@ class PayPalWebCheckoutClient internal constructor(
                 analytics.notify(PayPalEvent.STARTED, params = analyticsEventParams)
 
                 if (shopperSession != null) {
-                    val result = launchVaultWithSession(
+                    val result = launch(
```
sshropshire: `launch()` is a lot less descriptive, it's hard to determine at a glance what this method invocation does rather than `launchVault()` or `launchCheckout()` would.

sshropshire: Basically `tokenType == TokenType.ORDER_ID` means `launchCheckout()` here. We could make this more straightforward to improve scannability.

Image

sshropshire (approved):

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
sshropshire: This is take or leave, I don't think it's a show stopper.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -589,6 +462,115 @@ class PayPalWebCheckoutClient internal constructor(
...
+
+    /**
+     * Appends the given shopperSessionId as a query param, if it's not blank
+     */
+    private fun Uri.Builder.appendShopperSessionIdQueryParam(shopperSessionId: String) {
+        if (shopperSessionId.isNotBlank()) {
+            appendQueryParameter("shopperSessionId", shopperSessionId)
+        }
+
+        appendObservabilityQueryParams()
```
kgangineni: `appendObservabilityQueryParams()` is called separately and also here

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -414,113 +366,34 @@ class PayPalWebCheckoutClient internal constructor(
...
-                }
-                analytics.notify(
-                    event,
-                    params = analyticsEventParams,
-                    errorDescription = result.error.errorDescription
-                )
-            }
+        // If failed to launch, log latency with error.
+        val flowType = if (isVault) LatencyFlow.VAULT else LatencyFlow.CHECKOUT
+        if (result is PayPalPresentAuthChallengeResult.Failure) {
```
kgangineni: this may be correct, we're logging endTime only in success scenarios but not error

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
anibalb2500: Good idea.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
anibalb2500: How embarrassing. Good catch.

PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
anibalb2500: Updated. Good catch.

PR #484, opened 2026-07-30
Pass AnalyticsEventParams into latency analytics events

anibalb2500:

### Summary of changes

 - Adding shopperSessionId and orderId as analytics params for latency events, so they'll be included in analytics searches when searching by either of those fields.  

### Evidence
Screenshot 2026-07-30 at 1 27 58 PM

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
+++ b/PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
@@ -163,13 +163,11 @@ class PayPalWebCheckoutClient internal constructor(
             isVault = false,
         )
         if (urlConfig != null && !urlConfig.isValid()) {
-            logUserPerceivedLatencyError(LatencyFlow.CHECKOUT, startTime)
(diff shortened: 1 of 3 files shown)


PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt
```diff
@@ -392,7 +390,13 @@ class PayPalWebCheckoutClient internal constructor(
             returnToAppStrategy = returnToAppStrategy,
         )
         logPresentAuthChallengeResult(result)
-        logUserPerceivedLatency(flowType, result, startTime, endTime)
+
+        // If failed to launch, log latency with error.
+        if (result is PayPalPresentAuthChallengeResult.Failure) {
+            logUserPerceivedLatencyError(flowType, startTime)
```
kgangineni: Q: for errors why only logging it as just error without error code or any info?

anibalb2500: Good catch. I added error description to latency errors.

PR #485, opened 2026-07-30
setting default merchantId

anibalb2500:

### Summary of changes

 - Setting the default merchant id for demo app, in MerchantIntegration

 ### Checklist

 - [ ] Added a changelog entry

-


sshropshire (approved):

Demo/src/main/java/com/paypal/android/api/services/MerchantIntegration.kt
```diff
@@ -4,6 +4,6 @@ enum class MerchantIntegration(val baseUrl: String, val clientId: String, val me
     DEFAULT(
         baseUrl = "https://ppcp-mobile-demo-sandbox-87bbd7f0a27f.herokuapp.com/",
         clientId = "AQTfw2irFfemo-eWG4H5UY-b9auKihUpXQ2Engl4G1EsHJe2mkpfUv_SN3Mba0v3CfrL6Fk_ecwv9EOo",
-        merchantId = "" // TODO: Set with sandbox merchant account ID - currently only set by custom env settings
+        merchantId = "4823724566368455678"
```
kgangineni: this is correct merchantId V9YP27HFNG2LW
that format also doesn't look like merchantId

PR #489, opened 2026-08-07
fix: app switch user action

ayer-ribeiro:

### Summary of changes

 - Propagate the selected PayPal user action into the Demo create-order request
 - Always include `paymentSource.paypal.experienceContext` to ensure return and cancel url are passed.

 ### Checklist

- @ayer-ribeiro


Demo/src/main/java/com/paypal/android/usecase/CreateOrderUseCase.kt
```diff
@@ -22,44 +22,45 @@ class CreateOrderUseCase @Inject constructor(
...
+            sdkSampleServerAPI.createOrder(request.toOrderRequestBody())
+        }
+    }
+}
 
-            val purchaseUnit = PurchaseUnit(
-                amount = amount
+private fun OrderRequest.toOrderRequestBody(): OrderRequestBody {
+    val vaultAttributes = shouldVaultOnSuccess
+        .takeIf { it }
```
sshropshire, line 33: quick question: why's `takeIf` needed here?

sshropshire: It just seems like clever code. We're basically casting a `Boolean` to `Boolean?` so we can call `let` on it to run some conditional logic. An `if` statement would be much more expressive here.

ayer-ribeiro:
Closing this PR since the underlying fix was implemented in backend side instead of sdk (paypal-android).
The change needed to persist the `userAction` sent by PPCP Native SDK through `createPayPalSession` was applied in the backend patch for BWP flows, and we verified it works in both app switch and fallback. The continue flow still shows Link and review as expected.

PR #495, opened 2026-08-19
Add V3.0.0 Android integration guides

naolson-paypal:
## Summary
- Ports the new-structure V3.0.0 Android docs (Install & Setup, PayPal Checkout, Payment Buttons, Card/ACDC, Migrate V2→V3, Troubleshooting, Concepts & Responsibilities, Which Guide Do I Use) out of Confluence into `Docs/`, with internal links rewritten to work on GitHub.
- Organized into a folder hierarchy (`Docs/README.md` top-level entry, `getting-started/`, `integration-guides/`) so the intended reading order is explicit rather than relying on a flat file list.
- Every technical claim was fanned out and validated against the actual App Switch source (now merged into `main` via #493) — CoreConfig field names, result-type shapes, nullability, artifact coordinates, etc. — and corrected where the guides didn't match reality.
- Venmo is intentionally excluded from this PR — it hasn't merged yet. Those docs live on `docs/v3-venmo-guides`, ready for the follow-up release that includes Venmo.

## Test plan
- [ ] Spot-check that relative links resolve when browsing `Docs/` on GitHub
- [ ] Confirm code samples (CoreConfig construction, createPayPalSession/start/finishStart signatures) match the current main API surface


Docs/android-migrate-v2-to-v3.md
```diff
@@ -0,0 +1,111 @@
...
++             is PayPalFinishStartResult.Failure -> showError(result.error)
++             PayPalFinishStartResult.NoResult -> { /* intent was not a checkout return */ }
++         }
++     }
+  }
+```
+
+## Server-side change
+
+> **Note:** `TokenType` is currently marked internal-only in the SDK (`@RestrictTo(LIBRARY_GROUP)`). This guide shows the call shape `createPayPalSession()` expects; confirm with the SDK team that `TokenType` is public before shipping code that imports it directly.
```
kgangineni: This needs to be kept as a public API and not `@RestrictTo(LIBRARY_GROUP)`, we can fix in next release will not have any impact for integration to merchants

kgangineni (approved):
LGTM

Docs/android-migrate-v2-to-v3.md
naolson-paypal: Makes sense — removed the doc callout since the SDK team will handle this in a follow-up release with no impact to the call shape shown here. See 16cde8ad.

PR #501, opened 2026-08-28
(PATCH): migrate PayPal web fallback from Chrome Custom Tabs to Auth Tab

ayer-ribeiro:
Hi team! 👋

Chrome Custom Tabs are being superseded by [AndroidX Browser's Auth Tab](https://developer.android.com/reference/androidx/browser/auth/AuthTabIntent) for auth-style in-app browser flows, so I thought it'd be worth updating the PayPal web fallback (the path used when the PayPal app isn't installed or the buyer isn't app-switch eligible) to launch through Auth Tab instead of a Custom Tab.
I'm happy to contribute this migration — hopefully it's useful! Happy to adjust anything based on feedback.

### What changed

- The PayPal web fallback now launches via `androidx.browser.auth.AuthTabIntent` instead of a Chrome Custom Tab.
- Closing the Auth Tab before completing the flow now maps to a `Canceled` result, same as the existing Custom Tab behavior.
- App switch and Card 3DS flows are untouched — they keep using the existing Custom Tab implementation, so this is scoped to the web fallback only.
- `androidx.browser` bumped to `1.9.0` to get `AuthTabIntent`. That version's own AAR metadata requires `compileSdk 36`, AGP `8.9.1`, and Gradle `8.11.1`, so those got bumped alongside it to keep the build green.
- Docs and changelog updated to reflect Auth Tab instead of Custom Tab in the fallback path.

## New

- None — no new public API surface added.

## Modified

- Internal `browserswitch` package (`AuthTabActivity`, `AuthTabClient`, `BrowserSwitchClient`, `BrowserSwitchOptions`, `BrowserSwitchPendingState`) to support launching and handling Auth Tab results for the PayPal web fallback. These are internal implementation details, not public API.

## Recording

https://github.com/user-attachments/assets/d5e3db0e-6867-49d5-965e-003986f10972

## Breaking Changes

- [x] No breaking changes
- [ ] Breaking changes (describe migration steps below)

##Migration:

N/A — no public API changes, so no migration steps are needed for integrators. The behavior change is purely which in-app browser surface is used for the web fallback.

### Checklist

- [x] Added a changelog entry


anibalb2500 (approved):
LGTM

CardPayments/src/main/java/com/paypal/android/cardpayments/CardAuthLauncher.kt
```diff
@@ -75,6 +75,7 @@ internal class CardAuthLauncher(private val browserSwitchClient: BrowserSwitchCl
         return when (val result = captureDeepLink(requestCode, intent, authState)) {
             is CaptureDeepLinkResult.Success -> parseApproveOrderSuccessResult(result.deepLink)
             is CaptureDeepLinkResult.Failure -> CardFinishApproveOrderResult.Failure(result.reason)
+            is CaptureDeepLinkResult.Canceled -> CardFinishApproveOrderResult.NoResult
```
kgangineni: this should be `Canceled`

CardPayments/src/main/java/com/paypal/android/cardpayments/CardAuthLauncher.kt
```diff
@@ -84,6 +85,7 @@ internal class CardAuthLauncher(private val browserSwitchClient: BrowserSwitchCl
         return when (val result = captureDeepLink(requestCode, intent, authState)) {
             is CaptureDeepLinkResult.Success -> parseVaultSuccessResult(result.deepLink)
             is CaptureDeepLinkResult.Failure -> CardFinishVaultResult.Failure(result.reason)
+            is CaptureDeepLinkResult.Canceled -> CardFinishVaultResult.NoResult
```
kgangineni: same here

PayPalPayments/src/main/java/com/paypal/android/paypalpayments/PayPalClient.kt
```diff
@@ -170,7 +171,6 @@ class PayPalClient internal constructor(
         applicationScope.launch {
             try {
                 val shopperSession = deferred.await()
-                shopperSessionDeferred = null
```
anibalb2500, line 173: This actually seems out of scope. This would allow `start()` and `vault()` to be called with old `shopperSession` - we'd at least want to check `shopperSession.shopperSessionConfig.expiresAt` before reusing in `start` and `vault()` value.

CorePayments/src/main/java/com/paypal/android/corepayments/browserswitch/AuthTabClient.kt
```diff
@@ -0,0 +1,101 @@
...
+class AuthTabClient internal constructor(
+    private val authTabRegistry: AuthTabRegistry = AuthTabRegistry.shared,
+) {
+
+    @Suppress("ReturnCount") // Keeps each catch local to the registry registration call.
+    fun launch(
+        activity: ComponentActivity,
+        options: BrowserSwitchOptions,
+    ): LaunchAuthTabResult {
+        authTabRegistry.initialize(activity.application)
```
kgangineni: `authTabRegistry.register()`?, I am guessing this has to be 2 step process
registration should happen before activity is created

CorePayments/src/main/java/com/paypal/android/corepayments/browserswitch/AuthTabClient.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    private fun launchRegistered(
+        activity: ComponentActivity,
+        options: BrowserSwitchOptions,
+        launcher: ActivityResultLauncher<Intent>,
+    ): LaunchAuthTabResult {
+        val returnUrlScheme = options.returnUrlScheme
+        val appLinkUri = if (returnUrlScheme == null) options.appLinkUrl?.toUri() else null
+        val appLinkHost = appLinkUri?.host
+        val validationError = when {
```
kgangineni: this looks unnecessarily complex
```
when{
appLinkUrl !=null && validAppLinkUrl -> proceed 
returnUrlScheme!=null -> proceed
else -> {
   // error cases, 
}
```
can we make something like this

